PR #33, opened 2026-01-14
fix(sources): detect and fix misindexed YouTube sources

teng-lin:
## Summary

- Add detection for YouTube sources that were incorrectly indexed as web pages, resulting in only footer text (~100-200 chars) instead of the full transcript (~6000+ chars)
- Add `SourcesAPI.reindex()` method to delete and re-add a source, forcing NotebookLM to re-process it
- Add `source reindex` CLI command with confirmation prompt and `--force` flag
- Add `SourceContentType` enum and `YOUTUBE_MIN_EXPECTED_CHARS` constant for detection
- Update `source fulltext` command to show warnings when misindexed sources are detected
- Fix critical data loss risk by using add-then-delete order in reindex operation

### Detection Heuristics

A YouTube source is considered misindexed when:
1. URL is a YouTube URL but `source_type == 5` (WEB_PAGE) instead of `9` (YOUTUBE), OR
2. Character count is below 500 (threshold for likely footer-only content)

## Test plan

- [x] Unit tests for `SourceFulltext.is_likely_misindexed()` covering all edge cases
- [x] Unit tests for `SourceContentType` enum values
- [x] Unit tests for `YOUTUBE_MIN_EXPECTED_CHARS` constant threshold
- [x] Manual E2E testing: Added YouTube URL and verified transcript download works
- [x] Pre-commit checks pass (ruff format, ruff check, mypy, pytest)

### DIFF
--- a/src/notebooklm/_sources.py
+++ b/src/notebooklm/_sources.py
@@ -309,6 +309,17 @@ class SourcesAPI:
         if video_id:
             result = await self._add_youtube_source(notebook_id, url)
         else:
+            # Warn if URL looks like YouTube but we couldn't extract a video ID.
+            # This could indicate: (1) non-video YouTube page (channel, playlist),
+            # (2) new YouTube URL format we don't handle yet, or (3) malformed URL.
+            # In any case, adding as web page which may result in poor indexing.
+            if is_youtube_url(url):
+                logger.warning(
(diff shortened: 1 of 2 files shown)


teng-lin:
Should not make it so complicated.

PR #79, opened 2026-01-22
Fix Windows CLI hanging due to asyncio ProactorEventLoop

### LINKED ISSUE #75: notebooklm list hangs indefinitely on Windows (asyncio ProactorEventLoop issue)
old title: notebooklm list have no response. already log-ed in.

Image

hi, on wi10 i use sandboxie plus, which blocks stdio but allow TCP pass thru.
i ran vscode and all sorts of LLM things inside. including claude code, notebooklm-py.

notebooklm login works.

but notebooklm list and other "remote" commands failed.
those "local" commands worked e.g. notebooklm checking version.

thanks

wreuon:
Fixes #75

## Issue

CLI commands hang indefinitely on Windows after successful authentication. Affects all network-based commands: `list`, `create`, `source add`, `generate`, etc.

## Symptoms

```bash
$ notebooklm list
# Hangs forever with no output or error
```

Debug output shows successful authentication followed by hang:
```
DEBUG [notebooklm.auth] Authentication tokens obtained successfully
# Then nothing - process hangs
```

## Root Cause

Python's default `ProactorEventLoop` on Windows blocks indefinitely at `_overlapped.GetQueuedCompletionStatus()` during IOCP (I/O Completion Port) operations. This is a known Windows asyncio issue affecting network requests.

## Solution

Switch to `WindowsSelectorEventLoopPolicy` on Windows platforms. This event loop implementation is more reliable for network I/O on Windows.

## Testing

Tested on Windows 11 with Chinese locale:
- ✅ `notebooklm list` completes in <5 seconds (was hanging indefinitely)
- ✅ All CLI commands work correctly
- ✅ No impact on Linux/macOS (fix is Windows-specific)
- ✅ No breaking changes

## Impact

This is a critical bug fix that makes the CLI unusable on Windows systems. Without this fix, users must create wrapper scripts to work around the issue.

### DIFF
--- a/src/notebooklm/notebooklm_cli.py
+++ b/src/notebooklm/notebooklm_cli.py
@@ -23,6 +23,7 @@ LLM-friendly design:
   notebooklm ask "what are the key themes?"
 """
 
+import asyncio
 import logging
 import os
 import sys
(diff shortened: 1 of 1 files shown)


ccchan234:
https://github.com/teng-lin/notebooklm-py/issues/75

PR #113, opened 2026-02-10
Update helpers.py to check for title name as wel

Tikkel77:
Updated helper function resolve_partial_id to do a match on title as well, not just on ID. Also, increased length to for checking: no check was done if length was >20, updated to >50.

Potential bugs that may arise due to this change: item.title may not be available for each entity?


teng-lin:
Closing this PR due to CI failures that have remained unresolved for over two months.

If you'd like to continue this work, please feel free to reopen the PR after ensuring all CI checks pass. You can verify locally with:

```bash
```

Thanks for your contribution!

PR #160, opened 2026-03-07
test: add regression coverage for notebook summary/description RPC parsing

maxfraieho:
## Observed behavior (v0.3.3)

`get_summary()` returns a string representation of a Python list instead of the actual summary:
```
"[['Цей документ описує...', ...]]"
```

`get_description()` always returns `summary = ""` (empty string).

## Root cause

The SUMMARIZE RPC response has a 3-level nested structure:
```python
result = [
    [                               # result[0]: outer block
        ["actual summary text"],    # result[0][0]: summary block (list)
        [[question, prompt], ...],  # result[0][1][0]: topics
    ],
    "session-id-string"             # result[1]
]
```

The actual summary string is at `result[0][0][0]`.

In v0.3.3:
- `get_summary()` returned `str(result[0])` → string repr of the outer list
- `get_description()` checked `isinstance(result[0][0], str)` → always `False` (it's a list) → `summary = ""`

Both issues are already fixed in `main`. This PR adds regression tests to prevent future regressions.

## Tests added

`tests/unit/test_notebook_parsing.py` — no live network required (uses `AsyncMock`):
- `get_summary()` returns correct text from `result[0][0][0]`
- `get_description()` extracts summary and topics correctly
- Edge cases: empty result, malformed response, fallback behavior

## Note

Mutation APIs (create/rename/delete notebooks, sources, artifacts) were **not** affected by these bugs.

### DIFF
--- a/tests/unit/test_notebook_parsing.py
+++ b/tests/unit/test_notebook_parsing.py
@@ -1,0 +1,133 @@
+import pytest
+from unittest.mock import AsyncMock
+
+
+class RPCMethod:
+    SUMMARIZE = "SUMMARIZE"
+
+
+class SuggestedTopic:
+    def __init__(self, question: str, prompt: str):
+        self.question = question
+        self.prompt = prompt
+
+
+class NotebookDescription:
+    def __init__(self, summary: str, suggested_topics: list[SuggestedTopic]):
+        self.summary = summary
+        self.suggested_topics = suggested_topics
+
+
+class NotebookClient:
+    def __init__(self, core):
+        self._core = core
+
+    async def get_summary(self, notebook_id: str) -> str:
+        params = [notebook_id, [2]]
+        result = await self._core.rpc_call(
+            RPCMethod.SUMMARIZE,
+            params,
+            source_path=f"/notebook/{notebook_id}",
+        )
+        if result and isinstance(result, list) and len(result) > 0:
+            try:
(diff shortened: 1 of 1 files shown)


teng-lin:
Thanks for taking the time to look into this, @maxfraieho! The intent to add regression coverage is appreciated.

However, this bug was already fixed in PR #150, which also added comprehensive regression tests covering all the scenarios in this PR — including happy path, empty responses, malformed topics, and various edge cases. You can see the full test coverage in `tests/integration/test_notebooks.py` (classes `TestSummary`, `TestDescribeEdgeCases`, etc.) and `tests/e2e/test_notebooks.py`.

There are also a couple of issues with the current approach:

- The test file reimplements `NotebookClient`, `RPCMethod`, `SuggestedTopic`, and `NotebookDescription` inline rather than importing from the actual production code. This means if the real code regresses, these tests would still pass — defeating the purpose of regression tests.
- The `[[None]]` edge case expects `"None"` but the actual behavior would produce `"[None]"`.
- CI is failing due to a ruff import ordering issue.

Since we already have thorough coverage for this fix, I'm going to close this one. If you'd like to contribute tests in the future, `tests/integration/test_notebooks.py` is a great reference for the pattern we use — mocking RPC responses via `build_rpc_response()` and testing against the real `NotebookLMClient`.

Thanks again for the contribution!

Issue #180, opened 2026-03-10
Deep research report disappears after `import_sources()` — `report_id` returned but never used


mudsg24:
Hi,

Thank you so much for the prompt resolution of #116 — the Template Append Mode has been incredibly useful in my workflow.

I'd like to raise a new, unrelated issue regarding the Research Automation workflow. I've investigated the problem with the help of Claude, which assisted me in tracing through the codebase and drafting the analysis below. I hope this level of detail makes it easier to pinpoint the root cause.

---

### Problem

When using the Research Automation workflow to complete deep research programmatically, the **deep research report visible in the NotebookLM web UI disappears** after calling `import_sources()`. The same deep research completed manually through the web UI preserves the report normally.

### Steps to reproduce

1. Start deep research via `client.research.start(notebook_id, query, mode="deep")`
2. Poll until completion via `client.research.poll(notebook_id)`
3. Import sources via `client.research.import_sources(notebook_id, task_id, sources)`
4. Open the notebook in the NotebookLM web UI → the deep research report is gone

**Expected:** The deep research report remains accessible in the web UI after programmatic import, just as it does when completed manually.

**Actual:** The report disappears. Only the imported sources remain.

### Current workaround

In my automation pipeline, I changed the workflow to **fire-and-forget**: I call `client.research.start()` but never call `poll()` or `import_sources()`. Instead, I complete the deep research manually in the web UI to preserve the report. This works but defeats the purpose of automation.

### Analysis

I traced through the code and found what I believe is the root cause:

1. **`start()` returns `report_id` but it is never used anywhere** (`_research.py:100`):

   ```python
   report_id = result[1] if len(result) > 1 else None
   return {
       "task_id": task_id,
       "report_id": report_id,  # ← extracted but never consumed
       ...
   }
   ```

2. **`import_sources()` only sends `task_id`, not `report_id`** (`_research.py:247`):

   ```python
   params = [None, [1], task_id, notebook_id, source_array]
   ```

3. The server likely treats the `IMPORT_RESEARCH` RPC call as "consuming" the research task. Without `report_id` (or an additional RPC call), the server has no instruction to preserve the report.

4. The web browser likely sends either:

   - An additional parameter (possibly `report_id`) in the import call, or
   - A separate RPC call to accept/save the report

5. **Additionally**, `import_sources()` filters out sources without URLs (`_research.py:223`):

   ```python
   valid_sources = [s for s in sources if s.get("url")]
   ```

[...]

PR #181, opened 2026-03-10
fix: allow importing title-only sources in deep research

Jah-yee:
## Summary

This PR adds an `allow_title_only` parameter to `import_sources()` to support deep research sources that often only have titles without URLs.

## Problem

When using deep research, sources frequently only have titles without URLs. The current implementation filters out these sources, causing them to be silently dropped during import (issue #180).

## Solution

1. Added `allow_title_only` parameter (default: `False`) to `import_sources()`
2. When enabled, allows importing sources that only have titles (no URLs)
3. Updated the RPC payload format to handle title-only sources correctly

## Usage

```python
# Import deep research sources (including title-only)
imported = await client.research.import_sources(
    notebook_id,
    task_id,
    sources,
    allow_title_only=True  # Allow title-only sources
)
```

This is a backward-compatible change - existing code continues to work as before.

### DIFF
--- a/src/notebooklm/_research.py
+++ b/src/notebooklm/_research.py
@@ -197,6 +197,7 @@ class ResearchAPI:
         notebook_id: str,
         task_id: str,
         sources: list[dict[str, str]],
+        allow_title_only: bool = False,
     ) -> list[dict[str, str]]:
         """Import selected research sources into the notebook.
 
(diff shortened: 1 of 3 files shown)


teng-lin:
Closing this PR due to CI failures that have remained unresolved for over two weeks.

If you'd like to continue this work, please feel free to reopen the PR after ensuring all CI checks pass (linting, type checks, and tests). You can run the full check locally with:

```bash
```

Thanks for your contribution!

PR #196, opened 2026-03-11
feat(cli): add timeout flag to source add-research

### LINKED ISSUE #194: Add timeout option to source add-research import retries
## Summary

## Why this matters

## Proposed change

## Notes

stablegenius49:
## Summary
- add a `--timeout` option to `notebooklm source add-research`
- pass that value through to `import_with_retry()` when `--import-all` is used
- document the new CLI flag and cover default/custom timeout behavior in unit tests

## Testing
- `python -m pytest tests/unit/cli/test_source.py tests/unit/cli/test_notebook.py tests/unit/cli/test_research.py -q`
- `python -m ruff check src/notebooklm/cli/source.py tests/unit/cli/test_source.py`

Closes #194.


teng-lin:

PR #202, opened 2026-03-12
feat: add --timeout option to source add-research command

### LINKED ISSUE #194: Add timeout option to source add-research import retries

hobostay:
## Summary
Add configurable `--timeout` option to `source add-research` command for consistency with `research wait --import-all` command.

## Changes
- Add `--timeout` option (default: 300 seconds) to `source add-research`
- Replace hardcoded 60-iteration loop with dynamic calculation based on timeout value
- Update docstring with example showing custom timeout usage
- Timeout is ignored when using `--no-wait` flag

## Why this matters
Previously, `source add-research` had a hardcoded 300-second timeout (60 iterations × 5 seconds), while `research wait` had a configurable timeout. This inconsistency meant:
- The same import operation had different retry budgets depending on which command invoked it
- Users couldn't customize the timeout for long-running deep research sessions started via `source add-research`

## Testing
- All existing unit tests pass (46 tests in `tests/unit/cli/test_source.py`)
- Command loads successfully with new `timeout` parameter

Fixes #194

### DIFF
--- a/src/notebooklm/_chat.py
+++ b/src/notebooklm/_chat.py
@@ -21,7 +21,7 @@ from .types import AskResult, ChatReference, ConversationTurn
 
 logger = logging.getLogger(__name__)
 
-_DEFAULT_BL = "boq_labs-tailwind-frontend_20260301.03_p0"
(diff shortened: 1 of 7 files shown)


teng-lin:

PR #205, opened 2026-03-12
fix(auth): use domain-preserving cookie jar for token fetch

### LINKED ISSUE #114: RPC GET_NOTEBOOK failed
[github_issue_notebooklm.md](https://github.com/user-attachments/files/25276655/github_issue_notebooklm.md)

wavebyrd:
## Summary

Fixes #114. The token-fetch path in `fetch_tokens()` flattened all browser cookies into a single `Cookie` header string. When `httpx` followed redirects across Google domains (e.g., `notebooklm.google.com` -> `accounts.google.com`), every cookie was sent to every domain regardless of its original scope. This caused Google to reject the request or redirect to the account chooser/login page, resulting in `RPC rLM1Ne returned null result data` errors for `GET_NOTEBOOK`.

### What changed

- `fetch_tokens()` now accepts an optional `cookie_jar` parameter (`httpx.Cookies`). When provided, cookies are passed via `httpx.AsyncClient(cookies=...)` instead of a flat `Cookie` header. This preserves per-domain cookie scoping across redirects.
- All call sites now load a domain-preserving cookie jar via `load_httpx_cookies()` (which already existed for download auth) and pass it to `fetch_tokens()`.
- The flat-header fallback is retained for backward compatibility when `cookie_jar` is not provided.

### Files changed

- `src/notebooklm/auth.py` - Updated `fetch_tokens()` signature and `AuthTokens.from_storage()`
- `src/notebooklm/cli/helpers.py` - Updated `get_client()` to pass cookie jar
- `src/notebooklm/cli/session.py` - Updated `auth check --test` to pass cookie jar
- `src/notebooklm/cli/download.py` - Updated both download entry points to pass cookie jar
- `tests/unit/test_auth.py` - Added `TestFetchTokensWithCookieJar` test class

### Root cause (from issue diagnosis by @mouseroser)

Cookies from Playwright storage were flattened into a plain `Cookie` header for the token-fetch request. When redirects crossed into `accounts.google.com`, cookie domain scope was lost. A domain-preserving `httpx.Cookies` jar keeps the right cookies attached to each domain in the redirect chain.

## Test plan

- [ ] Existing `TestFetchTokens` tests continue to pass (legacy flat-header path)
- [ ] New `TestFetchTokensWithCookieJar` tests verify cookie jar path works correctly
- [ ] New test verifies cookie jar path does not inject flat cookie header values
- [ ] New test verifies redirect-to-login detection works with cookie jar
- [ ] CI passes

### DIFF
--- a/src/notebooklm/auth.py
+++ b/src/notebooklm/auth.py
@@ -175,6 +175,10 @@ class AuthTokens:
         This is the recommended way to create AuthTokens for programmatic use.
         It loads cookies from storage and fetches CSRF/session tokens automatically.
 
+        Uses a domain-preserving httpx.Cookies jar for token fetching so that
+        redirects across Google domains (e.g., notebooklm.google.com to
+        accounts.google.com) retain proper cookie scoping.
+
         Args:
             path: Path to storage_state.json. If None, uses default location
                   (~/.notebooklm/storage_state.json).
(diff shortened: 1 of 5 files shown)


teng-lin:
Closing this PR due to CI failures that have remained unresolved for over two weeks.

If you'd like to continue this work, please feel free to reopen the PR after ensuring all CI checks pass (linting, type checks, and tests). You can run the full check locally with:

```bash
```

Thanks for your contribution!

PR #208, opened 2026-03-15
cli: add --timeout to source add-research

### LINKED ISSUE #194: Add timeout option to source add-research import retries

Bortlesboat:
Fixes #194.

`source add-research --import-all` had no way to control the import retry budget — it always used `import_with_retry()`'s hardcoded 1800s default.

This adds `--timeout` (default 1800) and passes it through as `max_elapsed`, which is exactly how `research wait --import-all` already works.

**Changes:**
- `source add-research`: add `--timeout INT` option
- Pass `max_elapsed=timeout` to `import_with_retry()` when `--import-all` is set
- Update existing test assertion to include `max_elapsed=1800`
- Add test verifying a custom `--timeout 600` is forwarded correctly

## Summary by CodeRabbit

* **New Features**
  * Added a --timeout option for source import operations to control how long import retry cycles may run (default: 1800 seconds).

* **Chores**
  * Centralized the default retry timeout into a named constant.

* **Tests**
  * Added/updated tests to verify the timeout value is propagated to the import retry logic.

### DIFF
--- a/src/notebooklm/cli/helpers.py
+++ b/src/notebooklm/cli/helpers.py
@@ -42,6 +42,9 @@ logger = logging.getLogger(__name__)
 CONTEXT_FILE = get_context_path()
 BROWSER_PROFILE_DIR = get_browser_profile_dir()
 
+# Default timeout for import_with_retry (seconds)
+DEFAULT_IMPORT_RETRY_TIMEOUT_SECONDS = 1800
+
 # CLI artifact type name aliases
 _CLI_ARTIFACT_ALIASES = {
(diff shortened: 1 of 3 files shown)


teng-lin:
Useful feature and the implementation has tests. Currently conflicting with `main` — please rebase. After that this should be ready to merge.

teng-lin:
Closing — `notebooklm source add-research --timeout` was added in main (see `src/notebooklm/cli/source.py:567`) and issue #194 has been closed. Thanks for the PR.

PR #215, opened 2026-03-22
fix(auth): handle Navigation interrupted on already-authenticated sessions

### LINKED ISSUE #214: [Bug] Login crashes with "Navigation interrupted" when Google session already exists
## Environment

- **OS**: Windows 11 Pro
- **Python**: 3.x
- **notebooklm-py version**: 0.3.4
- **Playwright**: latest (pip install)

## Description

`notebooklm login` crashes with `Page.goto: Navigation interrupted` when the user already has an active Google session in the browser.

This is a **very common scenario** — most users keep their Google account signed in on Chrome. The crash makes first-time setup fail for the majority of users.

## Steps to Reproduce

1. Ensure you are already signed into a Google account in your default browser (Chrome)
2. Run `notebooklm login`
3. Playwright opens Chromium and navigates to Google Accounts URL
4. Because the session already exists, Google immediately redirects to NotebookLM
5. **Crash**: `Page.goto: Navigation interrupted`

## Root Cause

In `session.py:230`, `page.goto(GOOGLE_ACCOUNTS_URL)` is called without handling the case where the browser is already authenticated. When Google detects an existing session, it redirects immediately to NotebookLM, causing a navigation race condition in Playwright.

## Error Message

```
Page.goto: Navigation interrupted by another one
```

## Current Workaround

Manually extract `storage_state.json` using a custom Playwright script that handles the redirect, or use `NOTEBOOKLM_AUTH_JSON` environment variable with manually obtained cookies.

## Suggested Fix

Options:
1. Wrap `page.goto()` with `waitUntil: 'networkidle'` or `waitUntil: 'domcontentloaded'` and catch navigation errors
2. Detect already-authenticated state (e.g., check if redirected to notebooklm.google.com) and skip the login flow
3. Use `page.goto()` with `wait_until='commit'` and handle the redirect chain gracefully

## Related

- #213 (--browser chrome option for system Chrome)
- #205 (cookie jar domain scope fix)
- #133 (short auth validity)

---

*Discovered during a comparative evaluation of notebooklm-py vs notebooklm-mcp-cli.*

ahnbu:
## Summary

- Wrap cookie-forcing `page.goto()` calls in try-except to handle redirect race condition
- When a user already has an active Google session, Playwright throws `Navigation interrupted` because Google immediately redirects — this is safe to ignore since cookies are already set

## Problem

`notebooklm login` crashes with `Page.goto: Navigation interrupted` when the user already has an active Google session in the browser. This is the most common scenario — most users keep Google signed in on Chrome.

**Reproduction steps:**
1. Be signed into Google in your default browser
2. Run `notebooklm login`
3. Complete login, press ENTER
4. Crash at the cookie-forcing navigation (line 258-259)

## Fix

The two `page.goto()` calls for regional cookie forcing (`accounts.google.com` → `notebooklm.google.com`) are wrapped in try-except. When the redirect race occurs, the exception is silently caught — the cookies are already set in an authenticated session, so skipping is safe.

## Test plan

- [x] Tested on Windows 11 with Google session already active — login completes successfully
- [ ] Verify regional users (UK, CN) still get correct cookies

Fixes #214

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced login resilience to handle navigation interruptions and race conditions that previously caused authentication to fail.

### DIFF
--- a/src/notebooklm/cli/session.py
+++ b/src/notebooklm/cli/session.py
@@ -255,8 +255,16 @@ def register_session_commands(cli):
 
             # Force .google.com cookies for regional users (e.g. UK lands on
             # .google.co.uk). Use "load" not "networkidle" to avoid analytics hangs.
-            page.goto(GOOGLE_ACCOUNTS_URL, wait_until="load")
-            page.goto(NOTEBOOKLM_URL, wait_until="load")
+            # Wrap in try-except: already-authenticated sessions may redirect
+            # and cause "Navigation interrupted" (see GitHub issue #214).
+            try:
+                page.goto(GOOGLE_ACCOUNTS_URL, wait_until="load")
+            except Exception:
+                pass  # redirect race is harmless — cookies already set
+            try:
+                page.goto(NOTEBOOKLM_URL, wait_until="load")
+            except Exception:
+                pass  # same: redirect means we're already on NotebookLM
 
             current_url = page.url
             if NOTEBOOKLM_HOST not in current_url:


teng-lin:
Closing this as stale/superseded. #214 is already completed on main with narrower Playwright navigation handling and tests. This PR catches broad exceptions around login navigation, which could mask unrelated Playwright failures, so we should not merge it as-is. Thanks for the original fix attempt.

PR #216, opened 2026-03-22
fix(sources): resolve YouTube source add returning 'API returned no data'

ahnbu:
## Summary

- Remove `allow_null=True` from `_add_youtube_source()` so null API responses trigger the retry/auth-recovery path in `rpc_call`, matching `_add_url_source()` behaviour
- Update `Source.from_api_response()` to try metadata index 5 (YouTube) in addition to index 7 (web pages), reflecting a Google-side response structure change

## Root Cause

When adding a YouTube source, `_add_youtube_source()` called `rpc_call()` with `allow_null=True`. If the API returned null (e.g. due to stale session cookies), this bypassed the automatic retry/auth-recovery mechanism. The caller then saw `result is None` and raised `SourceAddError("API returned no data")`.

In contrast, `_add_url_source()` uses the default `allow_null=False`, which means null responses raise `RPCError` and trigger token refresh + retry — and URL sources work fine.

Additionally, Google moved YouTube source URLs from metadata index 7 to index 5 in the API response structure. The URL extraction in `Source.from_api_response()` now tries both positions.

## Test plan

- [ ] Verify YouTube URL source add succeeds (e.g. `notebooklm source add --url "https://www.youtube.com/watch?v=VIDEO_ID" NOTEBOOK_ID`)
- [ ] Verify regular URL source add still works
- [ ] Verify `notebooklm list --json` correctly extracts YouTube source URLs

Related: #214

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for YouTube sources: null API responses now trigger retry and authentication recovery flows instead of being silently processed
  * Enhanced source URL extraction to check multiple metadata locations, improving compatibility with API response structure variations

* **Documentation**
  * Updated changelog to document YouTube error handling refinements and URL extraction improvements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **YouTube source add "API returned no data"** — `_add_youtube_source` used `allow_null=True` which silently swallowed null API responses and bypassed retry/auth-recovery. Now uses `allow_null=False` (matching `_add_url_source`) so stale sessions trigger automatic token refresh.
+- **YouTube URL extraction from metadata** — `Source.from_api_response` now tries metadata index 5 (YouTube) in addition to index 7 (web pages), reflecting a Google-side response structure change.
+
 ## [0.3.4] - 2026-03-12
 
 ### Added
--- a/src/notebooklm/_sources.py
+++ b/src/notebooklm/_sources.py
@@ -866,7 +866,13 @@ class SourcesAPI:
         return bool(video_id and re.match(r"^[a-zA-Z0-9_-]+$", video_id))
 
     async def _add_youtube_source(self, notebook_id: str, url: str) -> Any:
-        """Add a YouTube video as a source."""
+        """Add a YouTube video as a source.
+
+        Note: allow_null is intentionally False (default) so that null API
+        responses trigger the retry/auth-recovery path in rpc_call, matching
+        the behaviour of _add_url_source.  Previously allow_null=True silently
+        swallowed null responses, preventing automatic retry on stale tokens.
+        """
         params = [
             [[None, None, None, None, None, None, None, [url], None, None, 1]],
             notebook_id,
@@ -877,7 +883,6 @@ class SourcesAPI:
             RPCMethod.ADD_SOURCE,
             params,
             source_path=f"/notebook/{notebook_id}",
-            allow_null=True,
         )
 
     async def _add_url_source(self, notebook_id: str, url: str) -> Any:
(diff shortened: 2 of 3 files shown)


teng-lin (approved):
## PR Review — Multi-Model Analysis (Claude + Gemini)

### Overall: Approve ✅

Clean, well-scoped fix with clear root cause analysis. Both changes are correct and low-risk.

---

### Consensus (Both Reviewers)

- **`allow_null=True` removal is correct** — Aligns YouTube with `_add_url_source()` behavior; enables retry/auth-recovery on null responses. The same RPC method (`ADD_SOURCE`) already works fine with `allow_null=False` for URL sources.
- **Index `(7, 5)` fallback is pragmatic** — Reasonable adaptation to Google's response structure change. Trying `7` first (web pages, common case) before `5` (YouTube) avoids false positives.
- **Both `types.py` code paths updated consistently** — Medium-nested and deeply-nested branches use the same `for idx in (7, 5)` loop with `break`.
- **Docstring addition is valuable** — Prevents future devs from re-adding `allow_null=True`.

---

### Recommendations

**1. (Important) Add unit tests for `Source.from_api_response` with YouTube metadata**

The fix adapts to a Google-side response structure change. A test with a YouTube response fixture where the URL lives at index 5 (not 7) would:
- Confirm the new fallback works
- Prevent regression if the loop logic is refactored later
- Document the expected response structure

Suggested test cases:
- YouTube source with URL at metadata index 5 → URL extracted correctly
- Web source with URL at metadata index 7 → URL still extracted (regression check)
- Source where both index 5 and 7 have data → index 7 wins (priority check)

**2. (Suggestion) Consider lightweight URL validation on index-5 extraction**

If a non-YouTube source type ever stores non-URL data at index 5, the current code would extract it as a URL. A simple `if url and url.startswith("http")` guard after the loop would add safety without complexity. Low priority since the `url` field is display-only and doesn't drive logic.

---

### Deep Analysis Notes

- Removing `allow_null=True` is safe even for non-auth null responses — the resulting `RPCError("No result found for RPC ID...")` is more informative than the previous `SourceAddError("API returned no data")`.
- The extensive `allow_null=True` usage elsewhere (`_notes.py`, `_sharing.py`, `_artifacts.py`) is correct for those operations; this fix is specific to `ADD_SOURCE` where null signals auth failure.
- The `if url: break` in both loops is critical — without it, index 5 would overwrite index 7's result. Both paths include it correctly.

🤖 Reviewed with Claude + Gemini

teng-lin:
Likely a real bug worth fixing. Two asks: (1) rebase onto current `main` (currently conflicting); (2) add a regression test that asserts the YouTube URL is extracted from metadata index 5 when the response shape that triggered the bug is returned — a fixture-driven decoder test in `tests/unit/` is ideal.

teng-lin:
Closing — the underlying bug (#214) has been resolved. The metadata[5] / metadata[7] URL-extraction fix landed in `src/notebooklm/types.py` (see `_extract_source_url`). Thanks for the diagnosis.

PR #225, opened 2026-03-27
fix: skip Unix-only file permissions on Windows

### LINKED ISSUE #212: Permission denied on storage_state.json on Windows with Python 3.14
Title: Permission denied on storage_state.json on Windows with Python 3.14

  Environment:
  - OS: Windows 11
  - Python: 3.14
  - notebooklm-py: 0.3.3 / 0.3.4

  Problem:

  After a successful notebooklm login, the storage_state.json file is saved but subsequent      
  commands (list, auth check, etc.) fail with:

  PermissionError: [Errno 13] Permission denied:
  'C:\Users\<user>\.notebooklm\storage_state.json'

  The file is created with Unix-style permissions (mode=0o700 / mode=0o600) via Python's        
  pathlib.mkdir() and open(). On Windows, this translates to an ACL that blocks read access from
   any new process — including other notebooklm commands running in a different shell session.  

  The same issue affects the browser_profile directory, which also becomes inaccessible
  ([WinError 183] or Access Denied) after creation.

  Setting NOTEBOOKLM_HOME to a different path (e.g. AppData\Local\notebooklm) does not resolve  
  the issue — the permission problem follows the file regardless of location.

  Root cause (suspected):

  Python's os.mkdir(path, mode=0o700) on Windows does not honor the mode argument the same way  
  as on Unix. The resulting directory/file ends up with an ACL that denies access to other      
  processes owned by the same user.

  Suggested fix:

  After creating storage_state.json and browser_profile, explicitly reset Windows ACLs using    
  icacls or os.chmod() with a permissive mode, or avoid passing mode=0o700 on Windows entirely: 

  import sys
  if sys.platform != "win32":
      path.mkdir(mode=0o700, parents=True, exist_ok=True)
  else:
      path.mkdir(parents=True, exist_ok=True)

voidborne-d:
## Problem

On Windows, `notebooklm login` saves `storage_state.json` successfully, but all subsequent commands fail with:

```
PermissionError: [Errno 13] Permission denied: C:\Users\<user>\.notebooklm\storage_state.json
```

Reported in #212.

## Root Cause

Python's `Path.mkdir(mode=0o700)` and `Path.chmod(0o600)` on Windows do not work like Unix. The `mode` argument to `mkdir()` is silently ignored, but the resulting directory gets a restrictive Windows ACL that blocks other processes (including other `notebooklm` commands in a new shell) from reading files inside it.

Same issue affects `browser_profile/` directory.

## Fix

Skip Unix-specific permission calls on Windows:
- `mkdir()` without `mode=` → inherits permissive ACLs from parent (standard Windows behavior)
- `chmod()` skipped → no-op on Windows anyway, avoids ACL confusion

Unix behavior (0o700 directory, 0o600 file) is fully preserved on Linux and macOS.

## Changes

- `paths.py`: `get_home_dir(create=True)` skips `mode=0o700` and `chmod(0o700)` on Windows
- `session.py`: login command skips `mode=0o700` on `mkdir()` and skips `chmod(0o600)` on Windows

Closes #212

## Summary by CodeRabbit

* **Bug Fixes**
  * Better Windows compatibility during account login and home-directory setup by avoiding Unix-style permission operations on Windows.
  * Maintains secure permission tightening on non-Windows systems.
  * Reduces setup and login failures and prevents inappropriate file-permission changes across platforms.

### DIFF
--- a/src/notebooklm/cli/session.py
+++ b/src/notebooklm/cli/session.py
@@ -189,20 +189,6 @@ def register_session_commands(cli):
             )
             raise SystemExit(1)
 
-        try:
-            from playwright.sync_api import sync_playwright
-        except ImportError:
-            if browser == "msedge":
-                install_hint = "  pip install notebooklm[browser]"
-            else:
-                install_hint = "  pip install notebooklm[browser]\n  playwright install chromium"
-            console.print(f"[red]Playwright not installed. Run:[/red]\n{install_hint}")
-            raise SystemExit(1) from None
-
-        # Pre-flight check: verify Chromium browser is installed (skip for Edge)
-        if browser == "chromium":
-            _ensure_chromium_installed()
-
         storage_path = Path(storage) if storage else get_storage_path()
         browser_profile = get_browser_profile_dir()
         if sys.platform == "win32":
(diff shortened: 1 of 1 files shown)


voidborne-d:
@teng-lin Added the regression tests you requested in `3f49a67`.

New test class `TestLoginWindowsPermissions` in `tests/unit/cli/test_session.py`:

1. **`test_windows_login_skips_mode_and_chmod`** — monkeypatches `sys.platform="win32"`, traces `Path.mkdir` and `Path.chmod`, asserts `mode=` kwarg is absent and `chmod` is never called
2. **`test_unix_login_sets_mode_and_chmod`** — same approach with `sys.platform="linux"`, asserts ≥2 `mkdir(mode=0o700)` calls and ≥2 `chmod(0o700)` calls
3. **`test_windows_storage_chmod_skipped`** — verifies the `sys.platform != "win32"` guard exists for `storage_state.json` `chmod(0o600)`

This covers both the `paths.py` changes (already tested in `test_paths.py`) and the `session.py` login command branch.

Issue #229, opened 2026-03-29
download -a (--artifact) flag downloads same binary regardless of artifact ID


simossene:
## Bug Description

The `notebooklm download infographic -a <artifact_id>` command matches the correct artifact metadata but downloads the same binary content for every artifact ID. The CLI reports the correct artifact title in the output, but the actual downloaded file is always the same (appears to be the latest artifact).

## Steps to Reproduce

1. Generate 3 infographics in a notebook with different prompts and sources:
   ```bash
   notebooklm generate infographic "prompt A" -s <source> -n <notebook> --json
   # Returns task_id: aaa
   notebooklm generate infographic "prompt B" -s <source> -n <notebook> --json
   # Returns task_id: bbb
   notebooklm generate infographic "prompt C" -s <source> -n <notebook> --json
   # Returns task_id: ccc
   ```

2. Wait for all to complete, then download by artifact ID:
   ```bash
   notebooklm download infographic "file-1.png" -a aaa -n <notebook> --force
   notebooklm download infographic "file-2.png" -a bbb -n <notebook> --force
   notebooklm download infographic "file-3.png" -a ccc -n <notebook> --force
   ```

3. CLI output shows correct artifact titles:
   ```
   Artifact: Title A (matched by ID: aaa)
   Artifact: Title B (matched by ID: bbb)
   Artifact: Title C (matched by ID: ccc)
   ```

4. **But all 3 files are visually identical** — the same binary content (verified by visual inspection; MD5 differs slightly likely due to download metadata).

## Expected Behavior

Each file should contain the unique content of the specified artifact.

## Workaround

Using `--all` to download all artifacts to a directory works correctly — produces distinct files with correct content:
```bash
notebooklm download infographic --all ./output/ -n <notebook>
```

## Environment

- notebooklm-py version: 0.3.4
- OS: Windows 11
- Python: 3.12

teng-lin:
Fixed in #230.

**Root cause**: `download_infographic` used `reversed()` iteration to find the infographic URL, which picked up the last matching URL-like field in the artifact structure. Since all artifacts shared the same structural shape, the reverse scan always landed on the same field — causing identical downloads regardless of artifact ID.

**Fix**: Replaced the iteration-based heuristic with fixed-index extraction at `art[14][2][0][1][0]`, matching how audio, video, and slide deck URLs are already extracted at known indices. Both `download_infographic()` and `_is_media_ready()` now share the same helper.

[...]

PR #230, opened 2026-03-29
fix: use fixed-index extraction for infographic URLs (#229)

teng-lin:
download_infographic used reversed() to scan the entire artifact data
structure for a URL-like pattern. Unlike other download methods (audio,
video, slide-deck) that use fixed indices, this reverse scan picks up
the last matching field. If Google added new URL-containing fields at
higher indices, the reverse scan would match those instead of the
actual content URL, causing all downloads to return the same binary
regardless of artifact ID.

Switch to forward iteration so the original content URL (at a lower
index) is found first.

https://claude.ai/code/session_013cR9fXhaAM6z53HBXQJfUp

## Summary by CodeRabbit

* **Bug Fixes**
  * Infographic downloads now consistently select the first valid media URL in forward order, avoiding inconsistent selection when multiple candidates exist.

* **Tests**
  * Reworked assertion formatting across tests for readability.
  * Added a unit test confirming infographic download uses the first matching URL and performs a single download attempt.

### DIFF
--- a/src/notebooklm/_artifacts.py
+++ b/src/notebooklm/_artifacts.py
@@ -1250,10 +1250,12 @@ class ArtifactsAPI:
         if not info_art:
             raise ArtifactNotReadyError("infographic")
 
-        # Extract URL from metadata
+        # Extract URL from metadata.
+        # Iterate forward (low→high index) so we match the original content
+        # URL field rather than any later URL-containing fields Google may add.
         try:
             metadata = None
-            for item in reversed(info_art):
+            for item in info_art:
                 if isinstance(item, list) and len(item) > 0 and isinstance(item[0], list):
                     if len(item) > 2 and isinstance(item[2], list) and len(item[2]) > 0:
                         content_list = item[2]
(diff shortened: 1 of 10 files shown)


teng-lin:
## Addressing review feedback

All three review findings have been addressed in commit `e4245aa`:

### 1. CodeRabbit Major — Align readiness/download URL selection
> `_is_media_ready()` delegates to `_find_infographic_url()` which scans in reverse, while `download_infographic()` iterates forward.

**Fixed**: Both `download_infographic()` and `_is_media_ready()` now delegate to `_find_infographic_url()`, which uses fixed-index extraction at `art[14][2][0][1][0]` — matching how audio (`art[6][5]`), video (`art[8]`), and slide deck (`art[16][3]`) URLs are extracted. Verified against 19 recorded cassettes and 3 live infographic downloads with distinct MD5 checksums.

### 2. CodeRabbit Nitpick — Regression test for multiple URL fields
> Add a test with two URL-bearing entries asserting the first is selected.

**Fixed**: `test_download_infographic_uses_fixed_index` tests with a URL at index 14 (expected) and a decoy URL at index 15 (ignored), asserting only the index-14 URL is downloaded.

### 3. CodeRabbit Nitpick — Unused `mock_core` variable
> Rename to `_mock_core` to suppress static analysis warning.

**Fixed**: Already addressed in commit `2ff7d0e`.

PR #236, opened 2026-04-01
fix: deduplicate research sources before import

wanderernupur-arch:
## Summary

- Adds `deduplicate_research_sources()` helper that checks existing notebook sources by URL before importing research results
- Prevents duplicate source creation when `--import-all` retries on RPC timeouts (the primary cause of 4-5x duplicates after deep research)
- Applied at three levels: `source add-research`, `research wait`, and between retries in `import_with_retry()`
- Report entries (result_type=5) are always preserved since they have no URL to deduplicate against

## Problem

Deep research with `--import-all` had no deduplication at any level. When an RPC call timed out and retried, the full source list was re-sent — and Google's server created new source entries each time. With default exponential backoff, this produced 4-5 copies of every source.

Additionally, running both `source add-research --import-all` and later `research wait --import-all` would double-import everything since neither checked what was already in the notebook.

## Changes

| File | Change |
|------|--------|
| `cli/helpers.py` | New `deduplicate_research_sources()` + dedup between retries in `import_with_retry()` |
| `cli/source.py` | Dedup before import in `source add-research --import-all` |
| `cli/research.py` | Dedup before import in `research wait --import-all` (both JSON and console paths) |
| `tests/unit/cli/test_helpers.py` | 7 new tests for dedup logic + updated existing retry tests |

## Design decisions

- **CLI-only policy**: Dedup is in the CLI layer, not the library API. `client.research.import_sources()` retains one-shot behavior for library consumers who may want explicit control.
- **URL-based matching**: Compares against `source.url` from existing sources. Sources without URLs (text, pasted) are ignored. Report entries are always kept.
- **`getattr` for URL access**: Handles source objects that may not have a `url` attribute (e.g., text sources) without raising `AttributeError`.

## Test plan

- [x] All 1431 existing unit tests pass
- [x] 7 new tests covering: URL filtering, report preservation, no-duplicate passthrough, all-duplicate detection, missing URL attribute handling, dedup between retries, skip-retry-when-all-imported
- [x] `ruff check`, `ruff format`, `mypy` all clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the source import workflow to detect and prevent duplicate research sources from being added to notebooks. The system now checks for existing sources before importing new ones, automatically skipping items that are already present.

* **Tests**
  * Added comprehensive test coverage for source deduplication functionality, validating edge cases and retry behavior scenarios.

### DIFF
--- a/src/notebooklm/cli/helpers.py
+++ b/src/notebooklm/cli/helpers.py
@@ -76,6 +76,60 @@ def run_async(coro):
     return asyncio.run(coro)
 
 
+async def deduplicate_research_sources(
+    client,
+    notebook_id: str,
+    sources: list[dict],
+    *,
+    json_output: bool = False,
+) -> list[dict]:
+    """Filter out research sources that already exist in the notebook.
+
+    Compares research source URLs against existing notebook sources to prevent
+    duplicate imports. Report entries (result_type=5) are always kept since
+    they have no URL to deduplicate against.
+
+    This is intentionally CLI-only policy. Library consumers calling
+    `client.research.import_sources()` directly still get one-shot behavior.
+
+    Args:
+        client: NotebookLM client instance.
+        notebook_id: The notebook ID.
+        sources: Research sources to filter.
+        json_output: If True, suppress console output.
+
+    Returns:
+        Filtered list of sources not already present in the notebook.
+    """
+    existing_sources = await client.sources.list(notebook_id)
+    existing_urls = {getattr(s, "url", None) for s in existing_sources}
+    existing_urls.discard(None)
+
+    new_sources = []
+    skipped = 0
+    for src in sources:
+        # Always keep report entries (no URL to deduplicate)
+        if src.get("result_type") == 5:
+            new_sources.append(src)
+            continue
+        url = src.get("url")
+        if url and url in existing_urls:
+            skipped += 1
+        else:
+            new_sources.append(src)
+
+    if skipped > 0:
+        logger.info(
+            "Deduplicated %d source(s) already present in notebook %s",
+            skipped,
+            notebook_id,
+        )
+        if not json_output:
+            console.print(f"[dim]Skipped {skipped} source(s) already in notebook[/dim]")
+
+    return new_sources
+
+
 async def import_with_retry(
     client,
     notebook_id: str,
@@ -106,18 +160,36 @@ async def import_with_retry(
             if remaining <= 0:
                 raise
 
+            # Before retrying, remove sources that were already imported
+            # (the timed-out call may have partially succeeded server-side)
+            sources = await deduplicate_research_sources(
+                client,
+                notebook_id,
+                sources,
+                json_output=json_output,
+            )
+            if not sources:
+                logger.info(
+                    "All sources already imported after timeout; skipping retry for notebook %s",
+                    notebook_id,
+                )
+                if not json_output:
+                    console.print("[green]All sources already imported[/green]")
+                return []
+
             sleep_for = min(delay, max_delay, remaining)
             logger.warning(
-                "IMPORT_RESEARCH timed out for notebook %s; retrying in %.1fs (attempt %d, %.1fs elapsed)",
+                "IMPORT_RESEARCH timed out for notebook %s; retrying in %.1fs (attempt %d, %.1fs elapsed, %d sources remaining)",
                 notebook_id,
                 sleep_for,
                 attempt + 1,
                 elapsed,
+                len(sources),
             )
             if not json_output:
                 console.print(
                     f"[yellow]Import timed out; retrying in {sleep_for:.0f}s "
-                    f"(attempt {attempt + 1})[/yellow]"
+                    f"(attempt {attempt + 1}, {len(sources)} sources remaining)[/yellow]"
                 )
             await asyncio.sleep(sleep_for)
             delay = min(delay * backoff_factor, max_delay)
--- a/src/notebooklm/cli/research.py
+++ b/src/notebooklm/cli/research.py
@@ -12,6 +12,7 @@ import click
 from ..client import NotebookLMClient
 from .helpers import (
     console,
+    deduplicate_research_sources,
     display_report,
     display_research_sources,
     import_with_retry,
@@ -188,16 +189,27 @@ def research_wait(ctx, notebook_id, timeout, interval, import_all, json_output, 
                     "report": report,
                 }
                 if import_all and sources and task_id:
-                    imported = await import_with_retry(
+                    sources = await deduplicate_research_sources(
                         client,
                         nb_id_resolved,
-                        task_id,
                         sources,
-                        max_elapsed=timeout,
                         json_output=True,
                     )
-                    result["imported"] = len(imported)
-                    result["imported_sources"] = imported
+                    if sources:
+                        imported = await import_with_retry(
+                            client,
+                            nb_id_resolved,
+                            task_id,
+                            sources,
+                            max_elapsed=timeout,
+                            json_output=True,
+                        )
+                        result["imported"] = len(imported)
+                        result["imported_sources"] = imported
+                    else:
+                        result["imported"] = 0
+                        result["imported_sources"] = []
+                        result["deduplicated"] = True
                 json_output_response(result)
             else:
                 console.print(f"[green]✓ Research completed:[/green] {query}")
@@ -206,14 +218,22 @@ def research_wait(ctx, notebook_id, timeout, interval, import_all, json_output, 
                 display_report(report)
 
                 if import_all and sources and task_id:
-                    with console.status("Importing sources..."):
-                        imported = await import_with_retry(
-                            client,
-                            nb_id_resolved,
-                            task_id,
-                            sources,
-                            max_elapsed=timeout,
-                        )
-                    console.print(f"[green]Imported {len(imported)} sources[/green]")
+                    sources = await deduplicate_research_sources(
+                        client,
+                        nb_id_resolved,
+                        sources,
+                    )
+                    if sources:
+                        with console.status("Importing sources..."):
+                            imported = await import_with_retry(
+                                client,
+                                nb_id_resolved,
+                                task_id,
+                                sources,
+                                max_elapsed=timeout,
+                            )
+                        console.print(f"[green]Imported {len(imported)} sources[/green]")
+                    else:
+                        console.print("[green]All sources already in notebook[/green]")
 
     return _run()
--- a/src/notebooklm/cli/source.py
+++ b/src/notebooklm/cli/source.py
@@ -27,6 +27,7 @@ from ..client import NotebookLMClient
 from ..types import source_status_to_str
 from .helpers import (
     console,
+    deduplicate_research_sources,
     display_report,
     display_research_sources,
     get_source_type_display,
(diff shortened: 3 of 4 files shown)


teng-lin:
# Comprehensive Review: Critical Issues & Recommendations

## Status: ⛔ NOT READY TO MERGE

Three specialized review agents identified **critical gaps in error handling and test coverage** that must be addressed before merge.

---

## 🚨 Critical Issues (Blocking)

### 1. Unhandled API Errors in deduplicate_research_sources()

**File**: `src/notebooklm/cli/helpers.py:105`

The new dedup function calls `client.sources.list()` with **zero error handling**:

```python
existing_sources = await client.sources.list(notebook_id)  # ← Can timeout, auth fail, rate limit
```

**What fails silently:**
- `RPCTimeoutError` — Network timeout (no retry, no logging)
- `AuthenticationError` — Session expired
- `RateLimitError` — API rate limit hit
- `HTTPError` — Network connectivity issue

**User Impact**: Research results lost, cryptic error messages, command crashes with zero context.

**Fix:**
```python
try:
    existing_sources = await client.sources.list(notebook_id, json_output=json_output)
except Exception as e:
    logger.warning("Could not verify existing sources (continuing): %s", e)
    return sources  # Fallback: import all, accept duplicates
```

---

### 2. Missing Critical Test Coverage for Retry Logic

**Test Gaps** (from pr-test-analyzer):

| Missing Test | Risk Level | Impact |
|---|---|---|
| Exponential backoff progression | 8/10 | Regression could cause API storm |
| max_delay ceiling enforcement | 7/10 | Users wait 10x longer |
| Multiple retries (3+ attempts) | 6/10 | Loop condition breaks undetected |
| Elapsed time budget edge cases | 7/10 | Function overshoots deadline |
| Attempt counter accuracy | 6/10 | Off-by-one user confusion |

**Current test grade**: B+ (78%) — Good fundamentals, critical gaps

**Priority 1 tests to add:**
- `test_exponential_backoff_progression()` — Verify delays increase: 2→4→8s
- `test_backoff_respects_max_delay_ceiling()` — Verify delays never exceed max

**Priority 2 tests to add:**
- `test_multiple_retries_until_success_on_third_attempt()` — Test 3+ retries
- `test_respects_elapsed_budget_for_partial_sleep()` — Edge case safety

---

## ⚠️ Important Issues (Should Fix)

### 3. Insufficient Error Context in import_with_retry()
**File**: `src/notebooklm/cli/helpers.py:163-180`

The retry loop calls `deduplicate_research_sources()` without error handling. Wrap with try-except to provide context when dedup fails.

### 4. CLI Commands Don't Catch Dedup Errors
**Files**: `src/notebooklm/cli/research.py:192-211, 221-237` and `src/notebooklm/cli/source.py:605-617`

All entry points call `deduplicate_research_sources()` without protection. If dedup fails, JSON responses become invalid and research results are lost.

### 5. No Error Scenario Tests
All 7 new tests cover only happy paths. Zero tests for timeout scenarios, API errors, or auth failures. Error handling bugs won't surface until production.

---

## ✅ What's Excellent

From code-reviewer (no issues >= 80% confidence):
- ✅ Architecture sound — correctly placed in CLI layer per CLAUDE.md
- ✅ Applied at three strategic points (source add-research, research wait, retry loop)
- ✅ Defensive coding with `getattr()` handles edge cases
- ✅ Follows all project patterns for logging, async, testing
- ✅ All CI checks pass across platforms

---

## 📋 Action Items

**Must fix before merge (1.5-2 hours):**

1. Add error handling to `deduplicate_research_sources()` (15 min)
   - Wrap `client.sources.list()` with try-except
   - Log warnings, provide fallback (return all sources)

2. Add critical test coverage (45 min)
   - `test_exponential_backoff_progression()`
   - `test_backoff_respects_max_delay_ceiling()`

3. Add important test coverage (30 min)
   - `test_multiple_retries_until_success_on_third_attempt()`
   - `test_respects_elapsed_budget_for_partial_sleep()`

4. Add error handling to CLI commands (20 min)
   - Wrap dedup calls in research.py and source.py with user-friendly error messages

**Nice to have:**
- Parametrized tests for backoff/max_delay combinations
- Improve dedup logging to include source counts
- Add tests for error scenario paths

---

## Summary

| Aspect | Grade |
|--------|-------|
| Code Quality | A |
| Architecture | A |
| Test Coverage | B+ |
| Error Handling | D |
| **Overall** | **C** (Not Ready) |

Once critical issues are addressed, this will be a solid, production-ready PR. Happy to discuss any questions!

teng-lin (changes requested):
Thanks for a thoughtful fix — the approach is right, the edge cases (report preservation, `None` URL discard) are handled, and the regression test for the retry path is the first of any PR that has addressed #241. I ran a multi-model deep review (Claude + Gemini) and have a few asks before merge.

## Must-fix before merge

### 1. Wrap `client.sources.list()` in `import_with_retry` with try/except — `cli/helpers.py:163-168`

The new dedup call is unprotected:

```python
sources = await deduplicate_research_sources(
    client, notebook_id, sources, json_output=json_output,
)
```

`client.sources.list()` calls `GET_NOTEBOOK`, which per `_core.py:65` can raise `NetworkError | RPCTimeoutError | RateLimitError | ServerError | ClientError`. Before this PR, a transient failure between retries was survivable — the code just slept and tried again. With this PR, any such failure propagates out of `import_with_retry` and aborts the whole operation. That's a regression for flaky networks — exactly the population the retry loop exists to serve.

Suggested fix:

```python
try:
    sources = await deduplicate_research_sources(
        client, notebook_id, sources, json_output=json_output,
    )
except Exception as exc:
    logger.warning("Dedup probe failed (%s); retrying with full source list", exc)
else:
    if not sources:
        logger.info("All sources already imported after timeout; skipping retry for notebook %s", notebook_id)
        if not json_output:
            console.print("[green]All sources already imported[/green]")
        return []
```

### 2. Add one test using a real `Source` dataclass — `tests/unit/cli/test_helpers.py`

All 7 new `TestDeduplicateResearchSources` tests use `existing = MagicMock(); existing.url = "..."`. Because `MagicMock` auto-creates attributes on access, these tests would also pass if the dedup used `.get("url")` instead of `getattr(s, "url", None)` — which is the exact trap that killed the prior attempt at this fix in #242 (`.get("url")` raises `AttributeError` on a dataclass, gets swallowed by `except Exception: pass`, and dedup silently no-ops).

The codebase has ~15 tests that construct real `Source` instances (see `tests/unit/test_source_status.py`, `tests/unit/cli/test_resolve.py`). Adding one in this style would catch the whole class of dataclass-API regressions going forward:

```python
from notebooklm.types import Source

@pytest.mark.asyncio
async def test_works_with_real_source_dataclass(self):
    """Regression test for #242-style bugs: ensure dedup works with real Source objects."""
    client = MagicMock()
    client.sources.list = AsyncMock(return_value=[
        Source(id="src_1", title="Existing", url="https://example.com/a"),
    ])
    sources = [
        {"url": "https://example.com/a", "title": "A"},
        {"url": "https://example.com/b", "title": "B"},
    ]
    with patch("notebooklm.cli.helpers.console"):
        result = await deduplicate_research_sources(client, "nb_123", sources)
    assert len(result) == 1
    assert result[0]["url"] == "https://example.com/b"
```

## Recommended follow-ups (this PR or separate)

### 3. Self-dedup the input list — `cli/helpers.py:101-110`

If the research RPC ever returns two entries with the same URL in one payload, both currently pass through. Cheap fix: add a `seen_urls` set inside the loop.

### 4. Consider sleeping *before* the dedup probe — `cli/helpers.py:160-175`

On the first timeout the server may still be committing in-flight imports when we probe `sources.list()`. Moving the dedup after `asyncio.sleep(sleep_for)` gives the server time to settle. The author of #291 reported 9 residual duplicates from exactly this race condition — this reordering may eliminate them.

### 5. URL normalization — `_url_utils.py`

Dedup is raw-string equality, so `foo` vs `foo/`, UTM query params, and `http` vs `https` can leak duplicates through. A small `_normalize_url()` helper in the existing `_url_utils.py` (strip trailing slash + fragment at minimum; optionally lowercase scheme/host and drop known tracking params) would close that gap. Not a regression from today; just caps effectiveness at the margin.

## Approved as-is

- **Pre-first-import dedup in `research.py` / `source.py`** — a legit UX improvement (idempotent `--import-all`). Worth a line in CLI help or the changelog since a user who deletes a source in the UI and re-runs will now see "All sources already in notebook" instead of a re-import; that's a behavior change.
- **Report-entry preservation** (`result_type == 5` keeps reports through dedup) — correct and carefully handled.
- **Discarding `None` from `existing_urls`** — prevents spurious matches against URL-less sources like text/PDF uploads.
- **Early-return when all sources already imported** — avoids a pointless `import_sources([])` retry.
- **Updates to existing `TestImportWithRetry` tests** — correct mocking of `client.sources.list` to keep them working.

## Summary

This is the cleanest attempt at #241 across the five PRs filed for it, and the only one with a regression test for the retry path. With the two must-fixes above, it's ready to merge. Happy to break out any of the follow-ups (3–5) into separate issues if you'd rather keep this PR tight.

teng-lin:
## Empirical verification of the review findings

Checked the PR out into a worktree and ran both a unit-level simulation of #241's canonical 414-vs-67 scenario and a live API test against my own NotebookLM account. Posting the results here because two of the concerns I raised in the review are no longer theoretical.

### ✅ The core #241 fix works end-to-end

**Simulation (server commits each batch, client times out 6×):**
```
Research sources in:    67
RPC calls made:         7  (1 initial + 6 retries then success)
Server-side rows:       67
Unique URLs imported:   67
Duplicates:             0
```

**Partial-progress simulation (server lands 10/attempt before timing out):**
```
Batch sizes sent per retry:  [67, 57, 47, 37, 27, 17, 7]
Final server rows:           67
Duplicates:                  0
```
Batches shrink monotonically — exactly the behavior #241 needs.

**Live API (double-import of the same research query):**

| Step | Sources found | Imported | Cumulative in notebook |
|---|---|---|---|
| Run 1: `source add-research "Python async context managers" --mode fast --import-all` | 10 | 10 | 10 + 1 report = 11 |
| Run 2: same command | 10 (different URLs) | 9 (1 overlap dedup'd) | 19 URLs + 1 report = 20, **0 duplicates** |

Then, to directly test the matching logic against real NotebookLM server URLs (some of which are long and have query params/fragments), I fed all 19 live URLs plus 1 obviously-new URL plus 1 report entry through `deduplicate_research_sources`:
```
Input:  21 (= 19 existing + 1 new + 1 report)
Output: 2 (the new URL + the report)
```
All 19 live URLs were correctly identified — including a gnarly one like `https://pytest-test-categories.readthedocs.io/en/latest/examples/async-testing.html`.

### ❌ Empirically confirmed: the two "must-fix" concerns from the review

**(1) Unprotected `sources.list()` in `import_with_retry` — regression reproduced:**

```python
# Mock client where sources.list() raises mid-retry (e.g. network blip, expired CSRF)
client.research.import_sources = AsyncMock(side_effect=[
    RPCTimeoutError("Timed out", timeout_seconds=30.0),
    [{"id": "src_a"}, {"id": "src_b"}, {"id": "src_c"}],
])
client.sources.list = AsyncMock(side_effect=RuntimeError("network blip"))

await import_with_retry(client, "nb_1", "task_1", sources)
# RuntimeError: network blip
# -> escapes import_with_retry, aborts the whole operation
```

Pre-PR, this retry would have succeeded (just sleep and re-try). Post-PR it crashes. The fix is a 3-line try/except (shown in the review).

**(2) URL normalization — leak reproduced:**

```python
client.sources.list = AsyncMock(return_value=[
    Source(id="s1", title="A", url="https://example.com/foo"),
])
research_sources = [{"url": "https://example.com/foo/", "title": "A"}]  # trailing slash

remaining = await deduplicate_research_sources(client, "nb_1", research_sources)
# remaining has 1 source — will be duplicated on retry
```

This matches the "9 residual duplicates" @lovebuilt reported in [#291's empirical table](https://github.com/teng-lin/notebooklm-py/pull/291) — so it's not a theoretical concern, it's the observed cap on effectiveness.

### ✅ Other checks — all pass

- Report entries (`result_type=5`) correctly preserved through dedup
- URL-less sources (text, uploaded PDFs) correctly discarded from the existing-URLs set (no spurious matches)
- All 83 existing `tests/unit/cli/test_helpers.py` tests pass
- Early-return when all sources already imported avoids a useless `import_sources([])` call

### Summary

The PR fixes #241's core bug cleanly — I can see this in both simulation and live runs. The two must-fixes in my review are about:
(a) not making things worse on flaky networks, and
(b) closing the last ~13% of duplicates from URL-canonicalization races.

With those two adjustments (~15 lines), this becomes a complete fix rather than a ~87% fix. Happy to iterate if any of the reproductions above need clarification.

teng-lin:
Closing this as superseded by #319, which has now merged and covers the #241 duplicate-import failure mode with a narrower implementation and tests. This PR also had unresolved requested changes, so keeping #319 as the landed fix. Thanks for the contribution.

PR #247, opened 2026-04-05
fix: migration overwrites stale profile cookies with fresh login data

LittleBitPlanet:
## Summary
- Migration from legacy flat layout to `profiles/default/` skipped copying `storage_state.json` when the destination already existed — regardless of whether the source was newer
- This caused a recurring auth failure cycle: `login` writes fresh cookies to the legacy root path, migration deletes the fresh file but keeps the stale profile copy, auth fails
- Fix: compare `st_mtime` before skipping — if the legacy root file is newer, overwrite the profile copy

## Root Cause

`_legacy_fallback()` in `paths.py` resolves `get_storage_path()` to the root `~/.notebooklm/storage_state.json` when it exists (for backwards compat). So `login` writes there. But `ensure_profiles_dir()` runs on every CLI invocation and triggers `migrate_to_profiles()` whenever legacy files exist at root. The migration copied root → profile on first run, but on subsequent runs it saw the profile copy already existed and skipped the copy — then deleted the (newer) root file anyway.

The original comment even said "skip if destination already exists **and is newer**" but the `and is newer` part was never implemented in the condition.

## Test plan
- [ ] Verify `ruff format`, `ruff check`, `mypy` pass (confirmed locally)
- [ ] Scenario: fresh `login` → next CLI command → auth still works (fresh cookies preserved in profile)
- [ ] Scenario: profile copy is already up-to-date → migration skips correctly (no unnecessary overwrites)

## Summary by CodeRabbit

* **Bug Fixes**
  * Profile migration now compares file modification times and only replaces profile data when the source is newer, avoiding unintended overwrites.
* **Tests**
  * Added regression tests to verify timestamp-based precedence during migration (overwrite when legacy is newer; keep when profile is newer).

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/247)

### DIFF
--- a/src/notebooklm/migration.py
+++ b/src/notebooklm/migration.py
@@ -79,12 +79,12 @@ def migrate_to_profiles() -> bool:
 
     logger.info("Migrating legacy layout to profiles/default/")
 
-    # Copy files (skip if destination already exists and is newer — avoids
-    # overwriting profile data that was updated after a partial migration)
+    # Copy files — overwrite if source is newer (e.g., login wrote fresh
+    # cookies to the legacy root path after a previous migration).
     for src in legacy_files:
         dst = default_dir / src.name
-        if dst.exists():
-            logger.debug("Skipping %s (already exists in profile)", src.name)
+        if dst.exists() and dst.stat().st_mtime >= src.stat().st_mtime:
+            logger.debug("Skipping %s (profile copy is same age or newer)", src.name)
         else:
             shutil.copy2(src, dst)
             if sys.platform != "win32":
--- a/tests/unit/test_migration.py
+++ b/tests/unit/test_migration.py
@@ -137,6 +137,49 @@ class TestMigrateToProfiles:
         # Original removed
         assert not (tmp_path / "storage_state.json").exists()
 
+    def test_overwrites_stale_profile_when_legacy_is_newer(self, tmp_path):
+        """Fresh login data at legacy root overwrites a stale profile copy.
+
+        Regression for the scenario where a previous migration left a profile
+        copy in place, and a subsequent `login` wrote fresh cookies to the
+        legacy root path. The migration must prefer the newer source.
+        """
+        default_dir = tmp_path / "profiles" / "default"
+        default_dir.mkdir(parents=True)
+        stale_profile = default_dir / "storage_state.json"
+        stale_profile.write_text('{"cookies":["stale"]}')
+        os.utime(stale_profile, (1000, 1000))
+
+        fresh_legacy = tmp_path / "storage_state.json"
+        fresh_legacy.write_text('{"cookies":["fresh"]}')
+        os.utime(fresh_legacy, (2000, 2000))
+
+        with patch.dict(os.environ, {"NOTEBOOKLM_HOME": str(tmp_path)}, clear=True):
+            migrate_to_profiles()
+
+        # Profile copy now holds the fresh data, legacy file is gone
+        assert json.loads(stale_profile.read_text()) == {"cookies": ["fresh"]}
+        assert not fresh_legacy.exists()
+
+    def test_keeps_newer_profile_when_legacy_is_older(self, tmp_path):
+        """An up-to-date profile copy is preserved over an older legacy file."""
+        default_dir = tmp_path / "profiles" / "default"
+        default_dir.mkdir(parents=True)
+        fresh_profile = default_dir / "storage_state.json"
+        fresh_profile.write_text('{"cookies":["fresh"]}')
+        os.utime(fresh_profile, (2000, 2000))
+
+        stale_legacy = tmp_path / "storage_state.json"
+        stale_legacy.write_text('{"cookies":["stale"]}')
+        os.utime(stale_legacy, (1000, 1000))
+
+        with patch.dict(os.environ, {"NOTEBOOKLM_HOME": str(tmp_path)}, clear=True):
+            migrate_to_profiles()
+
+        # Profile copy is preserved, legacy file is cleaned up
+        assert json.loads(fresh_profile.read_text()) == {"cookies": ["fresh"]}
+        assert not stale_legacy.exists()
+
 
 class TestEnsureProfilesDir:
     def test_creates_on_first_run(self, tmp_path):


teng-lin (commented):
Thanks for this fix, @LittleBitPlanet! The root cause analysis in the PR description is excellent — the comment/code mismatch where `"skip if destination already exists and is newer"` was never actually implemented is a great catch.

The fix itself is correct and minimal: `>=` with `shutil.copy2`'s mtime preservation gives you idempotency for free. All 17 CI checks pass and the code reads cleanly.

## Multi-model review summary

I ran this through several review passes (including error-handling, test-coverage, and type-design analysis). Here's the consensus:

### ✅ Strengths
- Fix implements what the original comment always intended — clean, minimal 4-line diff
- `>=` comparison correctly handles idempotency via `shutil.copy2`'s mtime preservation
- Excellent PR description with clear root cause analysis

### 📝 Recommendation: add mtime tests before merge

The PR changes a branch condition but doesn't add a test that would fail if the change were reverted. Two focused tests using `os.utime` would lock in the fix:

```python
def test_overwrites_when_source_is_newer(self, tmp_path):
    """Source file newer than profile copy triggers overwrite (the bug fix)."""
    default_dir = tmp_path / "profiles" / "default"
    default_dir.mkdir(parents=True)
    dst = default_dir / "storage_state.json"
    dst.write_text('{"cookies": ["old"]}')
    os.utime(dst, (1_000_000, 1_000_000))

    src = tmp_path / "storage_state.json"
    src.write_text('{"cookies": ["fresh"]}')
    os.utime(src, (2_000_000, 2_000_000))

    with patch.dict(os.environ, {"NOTEBOOKLM_HOME": str(tmp_path)}, clear=True):
        migrate_to_profiles()
    assert json.loads(dst.read_text()) == {"cookies": ["fresh"]}


def test_skips_when_destination_is_newer(self, tmp_path):
    """Profile copy newer than legacy source is preserved."""
    default_dir = tmp_path / "profiles" / "default"
    default_dir.mkdir(parents=True)
    src = tmp_path / "storage_state.json"
    src.write_text('{"cookies": ["stale"]}')
    os.utime(src, (1_000_000, 1_000_000))

    dst = default_dir / "storage_state.json"
    dst.write_text('{"cookies": ["current"]}')
    os.utime(dst, (2_000_000, 2_000_000))

    with patch.dict(os.environ, {"NOTEBOOKLM_HOME": str(tmp_path)}, clear=True):
        migrate_to_profiles()
    assert json.loads(dst.read_text()) == {"cookies": ["current"]}
```

### 💡 Minor observations (not blocking)

1. **Directory migration inconsistency** — Lines 94-101 still use the old `if dst.exists(): skip` pattern for `browser_profile/`. If browser profile data can be regenerated at the legacy root after a previous migration, the same stale-overwrite issue could apply. Might be worth a follow-up issue.

2. **FAT32 mtime granularity** — On FAT32/exFAT, mtime has 2-second precision, so a login within the same 2-second window as a previous copy could produce a false tie. Extremely unlikely in practice but a brief code comment would be nice.

Overall this is a solid, well-motivated fix. Just the two tests to add and it's good to go. 🎉

teng-lin:
## Follow-up: deeper investigation

After a more thorough investigation of the code flow, I want to flag some concerns for @teng-lin's consideration.

### The described scenario may be unreachable

The PR describes a "recurring auth failure cycle" where `login` writes fresh cookies to the legacy root path via `_legacy_fallback()`, then migration deletes the fresh file. However, tracing the actual code paths:

1. `_legacy_fallback()` (`paths.py:232`) returns the root path **only** when `not profile_path.exists() and resolved_profile == "default"` and the root path exists
2. After a successful migration, the root file is deleted and the profile file exists
3. `get_storage_path()` → `_legacy_fallback()` → profile path exists → returns profile path
4. `login` (`session.py:210`, `330`) writes to `get_storage_path()` → writes to **profile path**, not root
5. I checked all write paths (`context.storage_state()` in Playwright login, `storage_path.write_text()` in `--browser-cookies`) — all go through `get_storage_path()` → `_legacy_fallback()` → profile path when it exists

**There is no code path in this codebase that recreates `storage_state.json` at the root after a successful migration.** For the described scenario to occur, both root and profile copies would need to exist simultaneously with root being newer — but `login` always writes to wherever `get_storage_path()` points, which is the profile path once it exists.

The only ways I can see this triggering are:
- Manual/external creation of `~/.notebooklm/storage_state.json` by a user or external tool
- A crash during migration (but then both files have identical content from `shutil.copy2`)

### The fix itself is correct but the narrative may be overstated

The code change is **harmless and technically sound** — implementing the mtime check that the original comment always described. The `>=` comparison with `shutil.copy2`'s mtime preservation is correct for idempotency. It makes the migration more robust against edge cases involving external file manipulation.

However, the "recurring auth failure cycle" framing suggests a critical production bug, when the actual impact appears limited to scenarios involving external file creation outside this tool's control.

### Contributor investigation

The contributor account (created 2026-01-21) has no prior activity. The fork → branch → PR was completed in ~61 seconds. The commit is co-authored with "Claude Opus 4.6 (1M context)". This appears to be an AI-generated contribution.

### Recommendation

The fix is correct and harmless — I'd still accept it (with the mtime tests I suggested earlier), but wanted to flag these findings for transparency. @teng-lin, you're the best judge of whether there's a scenario I'm missing.

teng-lin:
Plausible bug fix and the change is small. Before merge, please add a regression test that exercises the migration path: a stale profile cookie should be overwritten by fresh login data. A unit test in `tests/unit/` that drives the migration helper directly would be sufficient.

teng-lin:
Took this over and added the regression tests. New commit `15a2c82`:

- `test_overwrites_stale_profile_when_legacy_is_newer` — newer legacy root file overwrites a stale profile copy (the regression target; verified to fail against the pre-fix `migrate_to_profiles`)
- `test_keeps_newer_profile_when_legacy_is_older` — up-to-date profile copy is preserved when the legacy root is older

Branch was 131 commits behind `main`, so I rebased onto `cff4250` first. Diff is unchanged from the original two-line fix plus the two new test cases.

Local pre-commit suite is green: `ruff format --check`, `ruff check`, `mypy`, and full `pytest` (2665 passed).

PR #248, opened 2026-04-05
feat: fix Google account switching (#246)

### LINKED ISSUE #246: Cannot switch Google accounts: login reuses cached browser profile + crashes on new account
## Bug Summary

`notebooklm login` cannot switch to a different Google account. Two issues:

1. **Cached browser profile auto-logs in:** Even after deleting `~/.notebooklm/storage_state.json`, running `notebooklm login` automatically logs into the previously authenticated account because the browser profile at `~/.notebooklm/browser_profile/` retains Google session cookies.

2. **Crash when adding new account:** When attempting to click "Use another account" on the Google sign-in page, the browser closes and throws a `TargetClosedError`.

## Steps to Reproduce

1. Authenticate with Account A: `notebooklm login`
2. Delete auth state: `rm ~/.notebooklm/storage_state.json`
3. Run `notebooklm login` again — it auto-logs into Account A (no account picker shown)
4. If you manually navigate to "Use another account" before it auto-redirects, the browser closes with:

```
playwright._impl._errors.TargetClosedError: Page.goto: Target page, context or browser has been closed
```

## Expected Behavior

- A `notebooklm logout` command (or `notebooklm auth logout`) that clears both `storage_state.json` and the browser profile session
- OR `notebooklm login --new` / `notebooklm login --account` flag that opens a fresh browser context without cached Google sessions

## Workaround

Manually delete the entire browser profile directory before re-login:

```bash
rm -rf ~/.notebooklm/browser_profile ~/.notebooklm/storage_state.json
notebooklm login
```

## Environment

- notebooklm-py: v0.3.4
- Python: 3.14
- OS: macOS (Darwin 24.6.0)
- Playwright browser: Chromium (bundled)

## Interested in Contributing

Happy to submit a PR for this if the maintainers agree on the approach (e.g., adding a `logout` command or a `--fresh` flag to `login`).

Halmurat-Uyghur:
## Summary

Fixes #246 — Cannot switch Google accounts: login reuses cached browser profile + crashes on new account.

- **`--fresh` flag on `login`**: Deletes the persistent browser profile (`~/.notebooklm/profiles/<profile>/browser_profile/`) before launching Playwright, giving users a clean browser session with no cached Google cookies
- **`auth logout` command**: Clears both `storage_state.json` and `browser_profile/` for the active profile, providing CLI symmetry with `login`
- **`TargetClosedError` handling**: Catches page destruction in both the initial navigation retry loop and the cookie-forcing section, recovering via `context.new_page()` with a browser-specific error message when recovery fails

## Test plan

- [x] 13 new unit tests covering all three features
- [x] All 2025 tests pass (0 failures, 10 skipped)
- [x] `ruff format` + `ruff check` + `mypy` all clean
- [x] Existing retry tests (`ERR_CONNECTION_CLOSED`, `ERR_CONNECTION_RESET`, `Navigation interrupted`) unaffected

## Changes

| File | Lines | What |
|------|-------|------|
| `src/notebooklm/cli/session.py` | +141/-9 | `--fresh` flag, `auth logout`, `_recover_page` helper, `TargetClosedError` handling |
| `tests/unit/cli/test_session.py` | +363 | 13 new tests across `TestLoginCommand` and `TestAuthLogoutCommand` |

## Summary by CodeRabbit

* **New Features**
  * Added `--fresh` flag to login to clear cached browser profile before authentication (warns if cookie-only mode is used).
  * Added `auth logout` command to remove saved session state and cached browser profile with clear user messaging.

* **Bug Fixes**
  * Improved recovery and retry behavior when the browser/page closes during login/navigation; clearer guidance when recovery fails.

* **Tests**
  * Added CLI tests for `--fresh`, logout, and recovery/failure scenarios.

### DIFF
--- a/src/notebooklm/cli/session.py
+++ b/src/notebooklm/cli/session.py
@@ -18,11 +18,15 @@ import time
 from collections.abc import Iterator
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import click
 import httpx
 from rich.table import Table
+
+if TYPE_CHECKING:
+    from playwright.sync_api import BrowserContext, Page
+    from rich.console import Console
 
 from ..auth import (
     ALLOWED_COOKIE_DOMAINS,
@@ -60,9 +64,9 @@ NOTEBOOKLM_HOST = "notebooklm.google.com"
 # Retryable Playwright connection errors
 RETRYABLE_CONNECTION_ERRORS = ("ERR_CONNECTION_CLOSED", "ERR_CONNECTION_RESET")
 LOGIN_MAX_RETRIES = 3
-# Playwright TargetClosedError substring — verified against Playwright >= 1.40.
-# If a future Playwright version changes this message, the error will propagate
-# unhandled (safe fallback) rather than silently ignored.
+# Playwright TargetClosedError substring — matches the default message from
+# Playwright's TargetClosedError class (introduced in v1.41). If a future
+# version changes this message, the error will propagate unhandled (safe fallback).
 TARGET_CLOSED_ERROR = "Target page, context or browser has been closed"
 BROWSER_CLOSED_HELP = (
     "[red]The browser window was closed during login.[/red]\n"
@@ -350,21 +354,28 @@ def _ensure_chromium_installed() -> None:
         )
 
 
-def _recover_page(context, console):
+def _recover_page(context: "BrowserContext", console: "Console") -> "Page":
     """Get a fresh page from a persistent browser context.
 
     Used when the current page reference is stale (TargetClosedError).
     A new page in a persistent context inherits all cookies and storage.
 
     Returns a new Page, or raises SystemExit if the context/browser is dead.
+    Raises the original PlaywrightError for non-TargetClosed failures.
     """
     from playwright.sync_api import Error as PlaywrightError
 
     try:
         return context.new_page()
     except PlaywrightError as exc:
-        console.print(BROWSER_CLOSED_HELP)
-        raise SystemExit(1) from exc
+        error_str = str(exc)
+        if TARGET_CLOSED_ERROR in error_str:
+            logger.error("Browser context is dead, cannot recover page: %s", error_str)
+            console.print(BROWSER_CLOSED_HELP)
+            raise SystemExit(1) from exc
+        # Not a TargetClosedError — don't mask the real problem
+        logger.error("Failed to create new page for recovery: %s", error_str)
+        raise
 
 
 def register_session_commands(cli):
@@ -443,8 +454,9 @@ def register_session_commands(cli):
                 shutil.rmtree(browser_profile)
                 console.print("[yellow]Cleared cached browser session (--fresh)[/yellow]")
             except OSError as exc:
+                logger.error("Failed to clear browser profile %s: %s", browser_profile, exc)
                 console.print(
-                    "[red]Cannot clear browser profile — it may be in use by another process.[/red]\n"
+                    f"[red]Cannot clear browser profile: {exc}[/red]\n"
                     "Close any open browser windows and try again.\n"
                     f"If the problem persists, manually delete: {browser_profile}"
                 )
@@ -845,6 +857,13 @@ def register_session_commands(cli):
           notebooklm auth logout           # Clear auth for active profile
           notebooklm -p work auth logout   # Clear auth for 'work' profile
         """
+        # Warn if env-based auth will remain active after logout
+        if os.environ.get("NOTEBOOKLM_AUTH_JSON"):
+            console.print(
+                "[yellow]Note: NOTEBOOKLM_AUTH_JSON is set — env-based auth will "
+                "remain active after logout. Unset it to fully log out.[/yellow]"
+            )
+
         storage_path = get_storage_path()
         browser_profile = get_browser_profile_dir()
 
@@ -855,20 +874,22 @@ def register_session_commands(cli):
             try:
                 storage_path.unlink()
                 removed_any = True
-            except OSError:
+            except OSError as exc:
+                logger.error("Failed to remove auth file %s: %s", storage_path, exc)
                 console.print(
-                    "[red]Cannot remove auth file — it may be in use by another process.[/red]\n"
+                    f"[red]Cannot remove auth file: {exc}[/red]\n"
                     "Close any running notebooklm commands and try again.\n"
                     f"If the problem persists, manually delete: {storage_path}"
                 )
-                raise SystemExit(1) from None
+                raise SystemExit(1) from exc
 
         # Remove browser profile directory
         if browser_profile.exists():
             try:
                 shutil.rmtree(browser_profile)
                 removed_any = True
-            except OSError:
+            except OSError as exc:
+                logger.error("Failed to remove browser profile %s: %s", browser_profile, exc)
                 partial = (
                     "[yellow]Note: Auth file was removed, but browser profile "
                     "could not be deleted.[/yellow]\n"
@@ -877,11 +898,11 @@ def register_session_commands(cli):
                 )
                 console.print(
                     f"{partial}"
-                    "[red]Cannot remove browser profile — it may be in use by another process.[/red]\n"
+                    f"[red]Cannot remove browser profile: {exc}[/red]\n"
                     "Close any open browser windows and try again.\n"
                     f"If the problem persists, manually delete: {browser_profile}"
                 )
-                raise SystemExit(1) from None
+                raise SystemExit(1) from exc
 
         if removed_any:
             console.print("[green]Logged out.[/green] Run 'notebooklm login' to sign in again.")
--- a/src/notebooklm/cli/session.py
+++ b/src/notebooklm/cli/session.py
@@ -527,7 +527,10 @@ def register_session_commands(cli):
 
                             backoff_seconds = attempt  # Linear backoff: 1s, 2s
                             logger.debug(
-                                f"Retryable error on attempt {attempt}/{LOGIN_MAX_RETRIES}: {error_str}"
+                                "Retryable error on attempt %d/%d: %s",
+                                attempt,
+                                LOGIN_MAX_RETRIES,
+                                error_str,
                             )
                             if is_target_closed:
                                 console.print(
@@ -545,8 +548,9 @@ def register_session_commands(cli):
                         elif is_target_closed:
                             # Exhausted retries on browser-closed errors
                             logger.error(
-                                f"Browser closed during login after {LOGIN_MAX_RETRIES} attempts. "
-                                f"Last error: {error_str}"
+                                "Browser closed during login after %d attempts. Last error: %s",
+                                LOGIN_MAX_RETRIES,
+                                error_str,
                             )
                             console.print(BROWSER_CLOSED_HELP)
                             raise SystemExit(1) from exc
--- a/src/notebooklm/cli/session.py
+++ b/src/notebooklm/cli/session.py
@@ -11,6 +11,7 @@ import asyncio
 import json
 import logging
 import os
+import shutil
 import subprocess
 import sys
 import time
@@ -59,6 +60,17 @@ NOTEBOOKLM_HOST = "notebooklm.google.com"
 # Retryable Playwright connection errors
 RETRYABLE_CONNECTION_ERRORS = ("ERR_CONNECTION_CLOSED", "ERR_CONNECTION_RESET")
 LOGIN_MAX_RETRIES = 3
+# Playwright TargetClosedError substring — verified against Playwright >= 1.40.
+# If a future Playwright version changes this message, the error will propagate
+# unhandled (safe fallback) rather than silently ignored.
+TARGET_CLOSED_ERROR = "Target page, context or browser has been closed"
+BROWSER_CLOSED_HELP = (
+    "[red]The browser window was closed during login.[/red]\n"
+    "This can happen when switching Google accounts in a persistent browser session.\n\n"
+    "Try:\n"
+    "  1. Run: notebooklm login --fresh\n"
+    "  2. Or run: notebooklm auth logout && notebooklm login"
+)
 CONNECTION_ERROR_HELP = (
     "[red]Failed to connect to NotebookLM after multiple retries.[/red]\n"
     "This may be caused by:\n"
@@ -198,9 +210,7 @@ def _login_with_browser_cookies(storage_path: Path, browser_name: str) -> None:
             storage_path.chmod(0o600)
     except OSError as e:
         logger.error("Failed to save authentication to %s: %s", storage_path, e)
-        console.print(
-            f"[red]Failed to save authentication to {storage_path}.[/red]\n" f"Details: {e}"
-        )
+        console.print(f"[red]Failed to save authentication to {storage_path}.[/red]\nDetails: {e}")
         raise SystemExit(1) from None
 
     console.print(f"\n[green]Authentication saved to:[/green] {storage_path}")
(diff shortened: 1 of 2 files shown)


src/notebooklm/cli/session.py
```diff
@@ -466,22 +517,41 @@ def login(storage, browser, browser_cookies):
...
-                            # Retryable error with attempts remaining: retry
+                        if (is_retryable or is_target_closed) and attempt < LOGIN_MAX_RETRIES:
+                            # For TargetClosedError, get a fresh page reference
+                            if is_target_closed:
+                                page = _recover_page(context, console)
+
                             backoff_seconds = attempt  # Linear backoff: 1s, 2s
                             logger.debug(
-                                f"Retryable connection error on attempt {attempt}/{LOGIN_MAX_RETRIES}: {error_str}"
+                                f"Retryable error on attempt {attempt}/{LOGIN_MAX_RETRIES}: {error_str}"
```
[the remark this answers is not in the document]
Halmurat-Uyghur: Addressed in e53d686: switched to `%` formatting for lazy evaluation.

src/notebooklm/cli/session.py
```diff
@@ -466,22 +517,41 @@ def login(storage, browser, browser_cookies):
...
+                                    f"(attempt {attempt}/{LOGIN_MAX_RETRIES}). "
+                                    f"Retrying in {backoff_seconds}s...[/yellow]"
+                                )
+                                time.sleep(backoff_seconds)
+                        elif is_target_closed:
+                            # Exhausted retries on browser-closed errors
+                            logger.error(
+                                f"Browser closed during login after {LOGIN_MAX_RETRIES} attempts. "
+                                f"Last error: {error_str}"
                             )
```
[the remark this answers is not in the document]
Halmurat-Uyghur: Addressed in e53d686: switched to `%` formatting for lazy evaluation.

src/notebooklm/cli/session.py
```diff
@@ -745,6 +828,62 @@ def auth_group():
...
+                    "[red]Cannot remove browser profile — it may be in use by another process.[/red]\n"
+                    "Close any open browser windows and try again.\n"
+                    f"If the problem persists, manually delete: {browser_profile}"
+                )
+                raise SystemExit(1) from None
+
+        if removed_any:
+            console.print("[green]Logged out.[/green] Run 'notebooklm login' to sign in again.")
+        else:
+            console.print("[yellow]No active session found.[/yellow] Already logged out.")
```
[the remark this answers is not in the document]
Halmurat-Uyghur: Intentionally kept separate — `context.json` stores notebook context (not credentials) and is owned by the `clear` command. Clearing it on logout would be an unexpected side effect since a user's active notebook selection is independent of which Google account is logged in. The existing `notebooklm clear` command handles this when needed.

Halmurat-Uyghur:
@teng-lin Thanks for the detailed root cause analysis and recommended fix in #246 — this PR follows your approach closely:

1. **`--fresh` flag** — `shutil.rmtree(browser_profile)` before `launch_persistent_context`, guarded with `if exists()` for first login and `except OSError` for locked profiles (Windows)
2. **`auth logout`** — clears both `storage_state.json` and `browser_profile/` with partial-state messaging if `unlink` succeeds but `rmtree` fails
3. **`TargetClosedError` handling** — caught in both Site A (retry loop) and Site B (cookie-forcing) with page recovery via `context.new_page()`. Uses `_recover_page` helper consistently, including for initial page creation. Shows browser-specific error message (not network error) when retries exhaust.

All edge cases from your comment are handled: first login with `--fresh`, logout when already logged out, Windows `PermissionError`/`OSError`, and the `--browser-cookies` interaction (warns that `--fresh` has no effect).

13 new tests, all 2025+ passing across macOS/Ubuntu/Windows CI. Ready for review!

src/notebooklm/cli/session.py
```diff
@@ -340,6 +350,23 @@ def _ensure_chromium_installed() -> None:
...
+    Used when the current page reference is stale (TargetClosedError).
+    A new page in a persistent context inherits all cookies and storage.
+
+    Returns a new Page, or raises SystemExit if the context/browser is dead.
+    """
+    from playwright.sync_api import Error as PlaywrightError
+
+    try:
+        return context.new_page()
+    except PlaywrightError as exc:
```
teng-lin, line 370: **[High]** `PlaywrightError` is the base class for all Playwright errors. If `context.new_page()` fails for a reason other than a dead context (resource exhaustion, browser crash, internal Playwright bug), the user sees "browser window was closed during login" which is misleading.

Suggestion — narrow the catch to only show `BROWSER_CLOSED_HELP` when it's actually a target-closed error:

```python
def _recover_page(context, console):
    from playwright.sync_api import Error as PlaywrightError

    try:
        return context.new_page()
    except PlaywrightError as exc:
        error_str = str(exc)
        if TARGET_CLOSED_ERROR in error_str:
            logger.error("Browser context is dead, cannot recover page: %s", error_str)
            console.print(BROWSER_CLOSED_HELP)
            raise SystemExit(1) from exc
        # Not a TargetClosedError — don't mask the real problem
        logger.error("Failed to create new page for recovery: %s", error_str)
        raise
```

This also adds `logger.error()` which is currently missing (every other terminal error path in this file logs before exiting).

*Found by: Claude Silent Failure Hunter, Codex*

src/notebooklm/cli/session.py
```diff
@@ -745,6 +832,62 @@ def auth_group():
...
+        browser_profile = get_browser_profile_dir()
+
+        removed_any = False
+
+        # Remove storage_state.json
+        if storage_path.exists():
+            try:
+                storage_path.unlink()
+                removed_any = True
+            except OSError:
```
teng-lin: **[High]** Three concerns with the OSError handling in `auth logout`:

1. This `except OSError:` doesn't bind the exception — the error details are truly discarded
2. `from None` at line 864 explicitly suppresses the exception chain
3. No `logger.error()` call — failure is invisible in logs

The user always sees "may be in use by another process" regardless of actual cause (could be `PermissionError`, `ReadOnlyFilesystemError`, or even `FileNotFoundError` from a TOCTOU race).

Suggestion:
```python
except OSError as exc:
    logger.error("Failed to remove auth file %s: %s", storage_path, exc)
    console.print(
        f"[red]Cannot remove auth file: {exc}[/red]\n"
        "Close any running notebooklm commands and try again.\n"
        f"If the problem persists, manually delete: {storage_path}"
    )
    raise SystemExit(1) from exc
```

Same pattern applies to the `shutil.rmtree` handler at line 871.

*Found by: Claude Silent Failure Hunter*

src/notebooklm/cli/session.py
```diff
@@ -745,6 +832,62 @@ def auth_group():
         """Authentication management commands."""
         pass
 
+    @auth_group.command("logout")
+    def auth_logout():
```
teng-lin, line 848: **[Medium]** `login` blocks when `NOTEBOOKLM_AUTH_JSON` is set (line 416), but `auth logout` has no equivalent check. After `auth logout`, the env var auth remains active — the user sees "Logged out" but the very next CLI command still works with the env var credentials.

Suggestion — mirror `login`'s guard, or at minimum warn:
```python
if os.environ.get("NOTEBOOKLM_AUTH_JSON"):
    console.print(
        "[yellow]Note: NOTEBOOKLM_AUTH_JSON is set — env-based auth will "
        "remain active after logout. Unset it to fully log out.[/yellow]"
    )
```

*Found by: Codex (gpt-5.4)*

src/notebooklm/cli/session.py
```diff
@@ -59,6 +60,17 @@
 # Retryable Playwright connection errors
 RETRYABLE_CONNECTION_ERRORS = ("ERR_CONNECTION_CLOSED", "ERR_CONNECTION_RESET")
 LOGIN_MAX_RETRIES = 3
+# Playwright TargetClosedError substring — verified against Playwright >= 1.40.
```
teng-lin: **[Low]** Playwright's `TargetClosedError` class was introduced in version 1.41 (not 1.40). Consider updating the version reference or replacing it with a source-of-truth reference:

```python
# Playwright TargetClosedError substring — matches the default message from
# Playwright's TargetClosedError class. If a future version changes this
# message, the error will propagate unhandled (safe fallback).
```

*Found by: Claude Comment Analyzer*

src/notebooklm/cli/session.py
```diff
@@ -393,12 +426,30 @@ def login(storage, browser, browser_cookies):
...
 
         storage_path = Path(storage) if storage else get_storage_path()
         browser_profile = get_browser_profile_dir()
+
+        if fresh and browser_profile.exists():
+            try:
+                shutil.rmtree(browser_profile)
+                console.print("[yellow]Cleared cached browser session (--fresh)[/yellow]")
+            except OSError as exc:
+                console.print(
```
teng-lin, line 458: **[Medium]** This error path is missing `logger.error()`. Every other terminal error path in this file logs before exiting. The same issue applies to `_recover_page()` at line 366.

Suggestion:
```python
except OSError as exc:
    logger.error("Failed to clear browser profile %s: %s", browser_profile, exc)
    console.print(...)
```

*Found by: Claude Silent Failure Hunter*

tests/unit/cli/test_session.py
```diff
@@ -392,6 +392,255 @@ def goto_side_effect(url, **kwargs):
...
+                result = runner.invoke(cli, ["login"])
+
+        assert result.exit_code == 1
+        # Should show browser-closed message, NOT network error message
+        assert "browser" in result.output.lower() and "closed" in result.output.lower()
+        assert "Network connectivity" not in result.output
+
 
 # =============================================================================
 # USE COMMAND TESTS
```
teng-lin, line 702: **[Medium]** Missing test: the cookie-forcing double-failure path (`session.py:592-595`) where the recovered page *also* raises `TargetClosedError` is never exercised. This is the final safety net — a regression here would give users an unhandled traceback instead of `BROWSER_CLOSED_HELP`.

Suggestion — add a test where `mock_context.new_page()` returns a page whose `goto()` also raises `TargetClosedError` for the cookie-forcing URLs (calls 2+).

*Found by: Claude Test Analyzer*

teng-lin (commented):
Thanks for this well-crafted PR, @Halmurat-Uyghur! The approach to `TargetClosedError` recovery is architecturally sound — particularly the insight that `context.storage_state()` captures cookies at the context level regardless of page lifecycle. The `--fresh` flag and `auth logout` give users clean escape hatches for account switching.

I ran this through multiple reviewers (Claude, Codex/gpt-5.4, Gemini 2.5 Pro) and compiled findings below. Two high-priority items and a few suggestions — overall this is in great shape.

## Summary

| Priority | Issue | Severity |
|----------|-------|----------|
| 1 | `_recover_page` catches all `PlaywrightError` broadly — misleading message for non-target-closed errors | **High** |
| 2 | `auth logout` OSError handlers discard exception details (`from None`, no logging, unbound exc) | **High** |
| 3 | `auth logout` doesn't check `NOTEBOOKLM_AUTH_JSON` env var (unlike `login`) | **Medium** |
| 4 | Missing test: cookie-forcing double-failure path (recovered page also dies) | **Medium** |
| 5 | Missing `logger.error()` in `_recover_page` and `--fresh` rmtree handler | **Medium** |
| 6 | Comment says "Playwright >= 1.40" but `TargetClosedError` was introduced in 1.41 | **Low** |

## Strengths

- Error message UX is excellent — every path gives actionable guidance
- `--fresh` + `--browser-cookies` interaction properly warns instead of silently ignoring
- TargetClosedError vs network error distinction in retry messaging prevents confusion
- Partial cleanup communication in `auth logout` tells user what succeeded/failed
- Logger calls in new code use `%s`-style formatting (lazy evaluation)
- Safe fallback design — if Playwright changes the error message, errors propagate unhandled rather than being silently swallowed

See inline comments for details on each finding.

teng-lin:
### Unaddressed review comments from other reviewers

I checked all comments from CodeRabbit, Gemini, and Copilot. Most were addressed or correctly rejected. Two sound items remain:

**1. Type hints for `_recover_page` (CodeRabbit, flagged twice)**

Both CodeRabbit review rounds flagged the missing type annotations on `_recover_page`. Since this project runs `mypy` as a pre-commit check, this is a reasonable suggestion. CodeRabbit's proposed fix using `TYPE_CHECKING` guard is correct:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from playwright.sync_api import BrowserContext, Page
    from rich.console import Console

def _recover_page(context: "BrowserContext", console: "Console") -> "Page":
```

Priority: **Low** — the function works fine without it, but it aligns with the project's direction.

**2. Inconsistent f-string logger formatting (CodeRabbit round 2)**

Lines 559-562 and 567 still use f-strings in `logger.error`/`logger.debug`, while commit e53d686 fixed the same pattern at lines 529-534. However, these are **pre-existing lines** not modified by this PR, so leaving them alone is arguably correct per project guidelines ("don't modify code you didn't change"). Not a blocker.

**3. Gemini's `context.json` clearing on logout — correctly rejected** ✅

@Halmurat-Uyghur's reasoning is sound: `context.json` stores notebook selection (not credentials) and is owned by the `clear` command. One edge case worth noting: after switching accounts via `auth logout` + `login`, the stale `context.json` will reference notebook IDs from the old account, which would fail on the next API call. But that's better handled by graceful error recovery in commands that read context, not by coupling cleanup to logout.

src/notebooklm/cli/session.py
Halmurat-Uyghur: Addressed in 4ac30ee: narrowed catch to only show BROWSER_CLOSED_HELP when TARGET_CLOSED_ERROR is in the error string. Other PlaywrightError types now re-raise. Added logger.error() for both paths.

src/notebooklm/cli/session.py
Halmurat-Uyghur: Addressed in 4ac30ee: bound OSError as exc, switched from `from None` to `from exc` for proper chaining, added logger.error(), and now show the actual error in the user message (e.g. `Cannot remove auth file: {exc}`).

src/notebooklm/cli/session.py
Halmurat-Uyghur: Addressed in 4ac30ee: added NOTEBOOKLM_AUTH_JSON warning at the top of auth logout, mirroring login's guard pattern.

src/notebooklm/cli/session.py
Halmurat-Uyghur: Addressed in 4ac30ee: updated comment to reference v1.41 and removed the specific version claim in favor of a source-of-truth description.

src/notebooklm/cli/session.py
Halmurat-Uyghur: Addressed in 4ac30ee: added logger.error() to the --fresh OSError path and both paths in _recover_page.

tests/unit/cli/test_session.py
Halmurat-Uyghur: Addressed in 4ac30ee: added test_login_cookie_forcing_double_failure_shows_browser_closed — exercises the inner Site B path where the recovered page also raises TargetClosedError. 2026 tests now passing.

Halmurat-Uyghur:
@teng-lin All 7 review items addressed in 4ac30ee. Ready for re-review when you get a chance. Thanks!

teng-lin:
Thanks @Halmurat-Uyghur — excellent work on this. The `_recover_page` helper is clean, coverage of both error sites plus the double-failure edge case is thorough, and you iterated responsively on Copilot, Greptile, Gemini, and review feedback. CI green across 15 OS/Python combos. Merging now. 🎉

PR #262, opened 2026-04-09
feat(source): add timeout to add-research imports

### LINKED ISSUE #194: Add timeout option to source add-research import retries

voidborne-d:
## Summary
- add a `--timeout` option to `source add-research` so `--import-all` can use an explicit retry budget
- keep the existing 1800s default for backwards-compatible behavior
- add CLI regression coverage and document the new flag

## Testing
- python3 -m pytest tests/unit/cli/test_source.py -k add_research

Closes #194

## Summary by CodeRabbit

* **New Features**
  * Added a configurable timeout for the "source add-research" command so import-all operations can use a custom maximum duration (default 1800s).

* **Documentation**
  * CLI reference updated to include the new --timeout flag, adjusted flag formatting, and updated examples showing --import-all with a custom timeout.

* **Tests**
  * Added a unit test verifying custom timeout is honored during import-all operations.


teng-lin:

PR #276, opened 2026-04-13
fix(auth): replace raw Cookie header with httpx cookie jar to fix #273

### LINKED ISSUE #273: Auth fails after ~10 min: raw Cookie header dropped on cross-domain redirects
## Bug

Authentication fails after ~10 minutes with "Authentication expired or invalid. Redirected to accounts.google.com" even though all session cookies (SID, HSID, etc.) are valid for 395+ days.

## Root Cause

`fetch_tokens()` in `auth.py` and `ClientCore.open()` in `_core.py` send cookies as a raw `Cookie:` header string:

```python
# auth.py line 605-610
cookie_header = "; ".join(f"{k}={v}" for k, v in cookies.items())
async with httpx.AsyncClient() as client:
    response = await client.get(
        "https://notebooklm.google.com/",
        headers={"Cookie": cookie_header},
        ...
    )
```

Google's `__Secure-1PSIDRTS` and `__Secure-3PSIDRTS` cookies have a ~10 minute TTL. When they expire, Google redirects `notebooklm.google.com` → `accounts.google.com/ServiceLogin` → `accounts.google.com/SetOSID` → back to `notebooklm.google.com`. 

**httpx drops raw `Cookie:` headers on cross-domain redirects** (by design — headers are per-request, not per-domain). So the redirect to `accounts.google.com` arrives with no cookies, fails, and lands on the login page.

## Proof

Same cookies, same moment — raw header fails, cookie jar works:

```python
# FAILS: raw Cookie header
async with httpx.AsyncClient() as client:
    r = await client.get("https://notebooklm.google.com/",
        headers={"Cookie": cookie_header}, follow_redirects=True)
    # Lands on accounts.google.com/signin (3 redirects)

# WORKS: httpx cookie jar
jar = httpx.Cookies()
for c in storage_state["cookies"]:
    jar.set(c["name"], c["value"], domain=c["domain"].lstrip("."))
async with httpx.AsyncClient(cookies=jar, follow_redirects=True) as client:
    r = await client.get("https://notebooklm.google.com/")
    # Lands on notebooklm.google.com (0 redirects)
```

## Additional Issue

`load_auth_from_storage()` filters cookies to `ALLOWED_COOKIE_DOMAINS` which excludes `accounts.google.com`. Cookies like `LSID`, `__Host-GAPS`, and `ACCOUNT_CHOOSER` (domain: `accounts.google.com`) are needed for the cross-domain redirect to succeed.

## Suggested Fix

Replace raw `Cookie:` headers with `httpx.Cookies` cookie jar in:
1. `auth.py:fetch_tokens()` — use `httpx.AsyncClient(cookies=jar)` instead of `headers={"Cookie": cookie_header}`
2. `_core.py:ClientCore.open()` — same change
3. `auth.py:extract_cookies_from_storage()` — include `accounts.google.com` domain cookies

This is a one-line change in each location. The cookie jar correctly sends domain-appropriate cookies on cross-domain redirects, which is how real browsers work.

## Environment

- notebooklm-py 0.3.4
- httpx 0.28.1
- Windows 11, Python 3.10
- Google account region: Canada

## Workaround

Monkey-patch `fetch_tokens` and `ClientCore.open` to use `httpx.Cookies` before importing the CLI. Happy to share the patch if helpful.

AmanSuryavanshi-1:
Closes #273

Fixes Issue #273 where short-lived 10 minute Google cookies triggering a 302 redirect to accounts.google.com would drop the raw Cookie headers and kill the session. Migrated the auth fetchers and the RPC ClientCore to use a persistent `httpx.Cookies` jar.

## Summary

**Problem:** Auth fails after ~10 minutes due to raw Cookie header being dropped on cross-domain redirects (Google's cookies expire every 10-15 minutes and trigger a 302 to `accounts.google.com` to refresh them).

**Solution:** Replace raw `Cookie:` header with `httpx.Cookies()` jar which properly handles cross-domain redirects.

### Changes Made

1. **`src/notebooklm/auth.py`** - `fetch_tokens()`: Uses `httpx.Cookies()` jar with domain=".google.com"
2. **`src/notebooklm/_core.py`** - `ClientCore.open()`: Uses cookies jar for HTTP client
3. **`src/notebooklm/_core.py`** - `update_auth_headers()`: Replaces cookies jar on token refresh
4. Added `_build_cookies_jar()` helper method

## Test Plan

- [x] All unit tests pass (179 tests)
- [x] All integration tests pass
- [x] Pre-commit checks pass (ruff, mypy)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of authentication cookies across domains to avoid session loss and preserve existing cookies during updates
  * Broadened Google auth domain support to reduce redirect/login failures
  * Persist refreshed cookies back to storage so auth updates survive restarts

* **Tests**
  * Added integration tests verifying cookies are preserved and merged (not replaced) during auth updates

### DIFF
--- a/src/notebooklm/_core.py
+++ b/src/notebooklm/_core.py
@@ -10,7 +10,7 @@ from urllib.parse import urlencode
 
 import httpx
 
-from .auth import AuthTokens
+from .auth import AuthTokens, build_cookie_jar, save_cookies_to_storage
 from .rpc import (
     BATCHEXECUTE_URL,
     AuthError,
@@ -128,6 +128,8 @@ class ClientCore:
         """Open the HTTP client connection.
 
         Called automatically by NotebookLMClient.__aenter__.
+        Uses httpx.Cookies jar to properly handle cross-domain redirects
+        (e.g., to accounts.google.com for auth token refresh).
         """
         if self._http_client is None:
             # Use granular timeouts: shorter connect timeout helps detect network issues
@@ -138,12 +140,19 @@ class ClientCore:
                 write=self._timeout,
                 pool=self._timeout,
             )
+            # Build cookies jar for cross-domain redirect support
+            # Use pre-built jar if available, otherwise build one
+            cookies = self.auth.cookie_jar or build_cookie_jar(
+                cookies=self.auth.cookies,
+                storage_path=self.auth.storage_path,
+            )
             self._http_client = httpx.AsyncClient(
                 headers={
                     "Content-Type": "application/x-www-form-urlencoded;charset=UTF-8",
-                    "Cookie": self.auth.cookie_header,
                 },
+                cookies=cookies,
                 timeout=timeout,
+                follow_redirects=True,
             )
 
     async def close(self) -> None:
@@ -152,6 +161,8 @@ class ClientCore:
         Called automatically by NotebookLMClient.__aexit__.
         """
         if self._http_client:
+            # Sync any newly refreshed tokens back to disk so they survive CLI restarts
+            save_cookies_to_storage(self._http_client.cookies, self.auth.storage_path)
             await self._http_client.aclose()
             self._http_client = None
 
@@ -161,17 +172,22 @@ class ClientCore:
         return self._http_client is not None
 
     def update_auth_headers(self) -> None:
-        """Update HTTP client headers with current auth tokens.
+        """Refresh auth metadata without resetting the live cookie jar.
 
         Call this after modifying auth tokens (e.g., after refresh_auth())
         to ensure the HTTP client uses the updated credentials.
+
+        The httpx client's cookie jar is authoritative once the session is
+        open. Re-injecting startup cookies here can overwrite cookies refreshed
+        during redirects to accounts.google.com.
 
         Raises:
             RuntimeError: If client is not initialized.
         """
         if not self._http_client:
             raise RuntimeError("Client not initialized. Use 'async with' context.")
-        self._http_client.headers["Cookie"] = self.auth.cookie_header
+
+        self.auth.cookie_jar = self._http_client.cookies
 
     def _build_url(self, rpc_method: RPCMethod, source_path: str = "/") -> str:
         """Build the batchexecute URL for an RPC call.
--- a/src/notebooklm/_core.py
+++ b/src/notebooklm/_core.py
@@ -161,10 +161,15 @@ class ClientCore:
         Called automatically by NotebookLMClient.__aexit__.
         """
         if self._http_client:
-            # Sync any newly refreshed tokens back to disk so they survive CLI restarts
-            save_cookies_to_storage(self._http_client.cookies, self.auth.storage_path)
-            await self._http_client.aclose()
-            self._http_client = None
+            try:
+                # Sync refreshed cookies only when auth came from an explicit file.
+                if self.auth.storage_path is not None:
+                    save_cookies_to_storage(self._http_client.cookies, self.auth.storage_path)
+            except Exception as e:
+                logger.warning("Failed to sync refreshed cookies during close: %s", e)
+            finally:
+                await self._http_client.aclose()
+                self._http_client = None
 
     @property
     def is_open(self) -> bool:
--- a/src/notebooklm/auth.py
+++ b/src/notebooklm/auth.py
@@ -31,9 +31,10 @@ import json
 import logging
 import os
 import re
+import tempfile
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any
+from typing import Any, TypeAlias
 
 import httpx
 
@@ -41,6 +42,11 @@ from ._url_utils import contains_google_auth_redirect, is_google_auth_redirect
 from .paths import get_storage_path
 
 logger = logging.getLogger(__name__)
+
+CookieKey: TypeAlias = tuple[str, str]
+DomainCookieMap: TypeAlias = dict[CookieKey, str]
+FlatCookieMap: TypeAlias = dict[str, str]
+CookieInput: TypeAlias = DomainCookieMap | FlatCookieMap
 
 # Minimum required cookies (must have at least SID for basic auth)
 MINIMUM_REQUIRED_COOKIES = {"SID"}
@@ -53,6 +59,7 @@ ALLOWED_COOKIE_DOMAINS = {
     ".notebooklm.google.com",
     "notebooklm.google.com",
     ".googleusercontent.com",
+    "accounts.google.com",  # Required for token refresh redirects
 }
 
 # Regional Google ccTLDs where Google may set auth cookies
@@ -148,14 +155,26 @@ class AuthTokens:
     """Authentication tokens for NotebookLM API.
 
     Attributes:
-        cookies: Dict of required Google auth cookies
+        cookies: Dict of required Google auth cookies keyed by (name, domain)
         csrf_token: CSRF token (SNlM0e) extracted from page
         session_id: Session ID (FdrFJe) extracted from page
+        storage_path: Path to the storage_state.json file, if file-based auth was used
+        cookie_jar: Domain-preserving httpx.Cookies jar. Preferred over flat cookies dict
+            for HTTP operations as it retains original cookie domains (e.g.,
+            .googleusercontent.com vs .google.com).
     """
 
-    cookies: dict[str, str]
+    cookies: DomainCookieMap
     csrf_token: str
     session_id: str
+    storage_path: Path | None = None
+    cookie_jar: httpx.Cookies | None = None
+
+    def __post_init__(self) -> None:
+        """Normalize legacy flat cookie mappings into domain-keyed mappings."""
+        self.cookies = normalize_cookie_map(self.cookies)
+        if self.cookie_jar is None:
+            self.cookie_jar = build_cookie_jar(cookies=self.cookies, storage_path=self.storage_path)
 
     @property
     def cookie_header(self) -> str:
@@ -164,7 +183,18 @@ class AuthTokens:
         Returns:
             Semicolon-separated cookie string (e.g., "SID=abc; HSID=def")
         """
-        return "; ".join(f"{k}={v}" for k, v in self.cookies.items())
+        return "; ".join(f"{k}={v}" for k, v in self.flat_cookies.items())
+
+    @property
+    def flat_cookies(self) -> FlatCookieMap:
+        """Return a legacy name→value cookie mapping.
+
+        When the same cookie name exists on multiple domains, the base
+        ``.google.com`` value wins for compatibility with the previous flat
+        representation. Domain-aware HTTP operations should use ``cookie_jar``
+        or ``cookies`` directly instead.
+        """
+        return flatten_cookie_map(self.cookies)
 
     @classmethod
     async def from_storage(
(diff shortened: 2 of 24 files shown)


teng-lin (changes requested):
Thanks for taking this on — the cookie-jar approach in `ClientCore.open()` is the right direction, and `build_httpx_cookies_from_storage` + adding `accounts.google.com` to `ALLOWED_COOKIE_DOMAINS` are solid building blocks. Before this can land there are a few concrete blockers and one structural suggestion.

## Blocking

**1. CI failure — unused import.** `fetch_tokens` is imported in `src/notebooklm/cli/helpers.py:26` but no longer called (the PR switched to a local import of `fetch_tokens_with_domains`). Ruff F401 is why the Code Quality check is currently red. Drop it from the top-level import.

**2. `save_cookies_to_storage` drops 0o600 permissions.** The write path in `src/notebooklm/auth.py` does `temp_path.write_text(...)` → `temp_path.replace(path)`. The new inode is created with the process umask (typically 0o644), so the `chmod 0o600` that `_login_with_browser_cookies` applies at `src/notebooklm/cli/session.py:197` is silently lost after the first `.close()` that syncs refreshed cookies. Sensitive tokens become world-readable. Fix: `os.chmod(temp_path, 0o600)` before `replace`, or use `os.open(..., flags, mode=0o600)` + `os.fdopen`. Also consider `tempfile.NamedTemporaryFile(dir=path.parent, delete=False)` to avoid collisions when two processes close at once.

**3. `update_auth_headers` re-flattens to `.google.com`.** `src/notebooklm/_core.py:186-189` iterates `self.auth.cookies` (a flat name→value dict) and calls `.set(name, value, domain=".google.com")` regardless of the cookie's original scope. After a redirect where Google has just set a fresh `__Host-GAPS` / `LSID` / `ACCOUNT_CHOOSER` on `accounts.google.com`, this creates a second `.google.com`-scoped entry in the jar (httpx will send both on subsequent `accounts.google.com` requests), and `save_cookies_to_storage` can then serialize the stale value because the match at `src/notebooklm/auth.py:776-778` accepts either the exact domain *or* `.google.com`. This directly undermines the PR's stated goal of preserving original per-cookie domains. Suggest: don't re-inject from the flat dict at all — the jar is authoritative between refreshes. If a refresh happens, update the jar from the response cookies, not from `auth.cookies`.

**4. Env-var `NOTEBOOKLM_AUTH_JSON` regression.** `src/notebooklm/cli/helpers.py:get_client` now does `if not storage_path: storage_path = get_storage_path()` *before* calling `load_auth_from_storage(storage_path)`. The env-var precedence documented at `src/notebooklm/auth.py:529-556` only kicks in when `path is None` — by synthesizing a concrete path up-front, this bypass the env-var branch entirely. CI / container users whose auth comes from `NOTEBOOKLM_AUTH_JSON` with no physical file will now fail with `FileNotFoundError` (both in `load_auth_from_storage` and again in `build_httpx_cookies_from_storage`). Revert the path synthesis; pass `None` through when no `--storage` flag was given, and teach `fetch_tokens_with_domains` to fall back to the flat-cookie path when there's no physical storage file.

**5. Fix is not wired through all token-fetch call sites.** The Python-API entry point `AuthTokens.from_storage()` at `src/notebooklm/auth.py:204` still calls the old `fetch_tokens(cookies)`, which collapses every cookie to `.google.com` at `src/notebooklm/auth.py:822-824`. `src/notebooklm/cli/download.py:168` and `:807` also still call the old function. These are exactly the call sites that hit the 10-minute refresh window, so the bug returns for any caller not going through `cli/helpers.get_client`.

## Structural suggestion

Several of the findings above are symptoms of the same thing: `AuthTokens.cookies` is still `dict[str, str]`, so every writer into the jar (`update_auth_headers`, `fetch_tokens`, `save_cookies_to_storage`, `_build_cookies_jar` fallback) has to re-invent domain context — and each reinvention reintroduces the collapse-to-`.google.com` bug in a new place. Rather than maintaining two parallel paths (`fetch_tokens` + `fetch_tokens_with_domains`, `extract_cookies_from_storage` + `extract_cookies_with_domains`), consider:

- Change `AuthTokens.cookies` to `dict[tuple[str, str], str]` keyed on `(name, domain)`. Derive `cookie_header` as a `@property` for any legacy consumer.
- Make `fetch_tokens` take the jar (or `AuthTokens`) directly and delete `fetch_tokens_with_domains`. Route `AuthTokens.from_storage`, `cli/download.py:168`, `cli/download.py:807`, `cli/session.py:852`, and `cli/helpers.get_client` through the single implementation.
- `ClientCore._build_cookies_jar` then reads from `AuthTokens.cookies` directly without re-parsing `storage_state.json` on every `open()` (avoids the double I/O at `src/notebooklm/_core.py:201-204`).

This collapses three code paths into one and makes domain preservation a type-level invariant rather than a runtime discipline.

## Tests

The two new tests in `TestCrossDomainCookiePreservation` only verify that in-memory jar entries survive `update_auth_headers()`. They don't exercise the scenario the PR is actually fixing. To prove the fix:

- Use `pytest-httpx` to simulate a 302 from `notebooklm.google.com` → `accounts.google.com`, have the mock set a `Set-Cookie` scoped to `accounts.google.com`, then assert that a subsequent request on the `accounts.google.com` hop carries that cookie.
- Add a round-trip test for `save_cookies_to_storage` that asserts file permissions remain `0o600` after sync.
- Update `tests/unit/test_auth.py:544` — `test_fetch_tokens_includes_cookie_header` still locks in the old raw-`Cookie`-header shape. It passes today because httpx derives the header from the jar, but it no longer validates what the PR changes. Replace with a cross-domain-redirect assertion.

## uv.lock

The `uv.lock` churn is fine — `ruff==0.8.6` and the `rookiepy` optional extra already exist on `main` (see `bc8751d`, `cea0566`, `da55b46`). This is a stale-branch resync, not unrelated scope creep. Not a blocker.

---

Happy to discuss any of this, especially the `AuthTokens` shape change if that feels too invasive for one PR. We could also split the structural refactor off as a prep commit and rebase the cookie-jar fix on top. Either way, thanks again for the careful analysis in the description — it matched the wire-level diagnosis from #114 exactly.

aayushprsingh (commented):
Thanks @AmanSuryavanshi-1 — the direction is solid. To unblock CI:

**Fix the unused import** in src/notebooklm/cli/helpers.py:
The old fetch_tokens import at line ~26 is no longer needed since you're now using fetch_tokens_with_domains locally. Just remove fetch_tokens from that import.

Happy to help iterate if you want me to take over the PR!

AmanSuryavanshi-1:
@teng-lin Thank you for the thorough review. I've addressed all 5 points and implemented the structural suggestion:

**Bug 1 (file permissions):** \save_cookies_to_storage\ now uses \	empfile.NamedTemporaryFile(dir=path.parent)\ + \os.chmod(temp_path, 0o600)\ before \
eplace()\. Permissions are preserved.

**Bug 2 (_is_allowed_cookie_domain):** Picked up via rebase onto main. Used in \save_cookies_to_storage\ to filter the cookie jar before serialization.

**Bug 3 (update_auth_headers re-flattening):** \update_auth_headers()\ no longer re-injects from the flat dict. The live httpx cookie jar is authoritative during the session. The method now simply syncs \self.auth.cookie_jar = self._http_client.cookies\.

**Bug 4 (NOTEBOOKLM_AUTH_JSON regression):** \rom_storage()\ now checks \os.environ.get(NOTEBOOKLM_AUTH_JSON)\ before synthesizing a storage path. \get_client\ passes \None\ through when no \--storage\ flag is given, preserving the env-var branch.

**Bug 5 (unwired call sites):** \download.py:168\ and \:807\ now use \AuthTokens.from_storage(storage_path)\ directly, which goes through the domain-preserving \_fetch_tokens_with_jar\ path.

**Structural suggestion:** Implemented. \AuthTokens.cookies\ is now \dict[tuple[str, str], str]\ keyed on \(name, domain)\. \__post_init__\ normalizes legacy flat dicts via \
ormalize_cookie_map()\. \cookie_header\ uses a \lat_cookies\ property for backward compat. Single fetch implementation via \_fetch_tokens_with_jar()\.

Squashed the commit history from 12 → 2 clean commits and rebased onto latest main.

Miles0sage:
+1, confirmed fix on Linux headless VPS. Reproduced #273 exactly: lib 0.3.3 cookies (full set incl. SID/HSID/__Secure-1PSID/__Secure-3PSID/SAPISID) → after one redirect chain notebooklm.google.com → accounts.google.com → notebooklm.google.com, raw `Cookie:` header was dropped on the second hop and subsequent calls 401'd until manual re-login. Behaviour observed: ~every 10 min (matches `__Secure-1PSIDRTS` TTL).

Pinned this PR head (`be87002e45a`) into prod via `pip install --break-system-packages git+https://github.com/AmanSuryavanshi-1/notebooklm-py@be87002e45aa82cbb22df934aa477595ae896e8b` (becomes 0.3.4). Both first call and a follow-up call (which deliberately exercises the cross-domain redirect path) now succeed. 68 notebooks listed, two consecutive runs, zero re-auth needed.

One small adoption note for downstream MCPs: `extract_cookies_with_domains()` is correctly stricter than the old `extract_cookies_from_storage()` — it raises `ValueError: Missing required cookies: {'SID'}` instead of silently auth-failing later. Anyone bumping should expect that. The new multi-profile layout migration (`~/.notebooklm/profiles/<name>/...`) ran cleanly on first import.

LGTM.

AmanSuryavanshi-1:
@teng-lin Pushed `4c0b46d` - a 6-line deletion that deduplicates the import block in `src/notebooklm/cli/helpers.py` introduced by the merge-main commit (`82a89f5`).

**Root cause:** The merge concatenated both the old main imports and our new imports instead of deduplicating, creating 5 ruff errors (F811x3, F401x1, I001x1).

**Fix:** Removed the stale duplicate block (old `fetch_tokens`, redundant `AuthTokens`, `load_auth_from_storage`, `RPCTimeoutError` re-imports). Kept `NetworkError` and `RPCError` which are used by `import_with_retry()`.

**Local verification:**
- `ruff check src/ tests/` -> All checks passed
- `ruff format --check src/ tests/` -> 138 files already formatted
- `pre-commit run --all-files` -> ruff pass, ruff-format pass
- `pytest tests/unit/ -x -q` -> 1544 passed, 3 skipped, 0 failed
- `mypy src/notebooklm/cli/helpers.py` -> 0 errors

CI should go green once the workflow run is approved. No auth logic was touched - purely import cleanup.

AmanSuryavanshi-1:
Rebased onto latest `main` (`2a9a3a4`) and resolved the merge conflict in `helpers.py`.

**Additional fix:** 5 tests in `test_note.py` and `test_notebook.py` (from PRs #259 and #328) were mocking `notebooklm.cli.helpers.fetch_tokens` which no longer exists after our import cleanup. Updated them to mock `notebooklm.auth.fetch_tokens_with_domains` to match every other test in those files.

**Clean commit history** (2 commits on top of main):
- `5c3f325` fix(auth): replace raw Cookie header with domain-preserving httpx cookie jar
- `4bacf1c` test: add cross-domain cookie preservation and redirect tests

**Full local verification:**
- `ruff check src/ tests/` -> All checks passed
- `ruff format --check` -> 140 files already formatted
- `pre-commit run --all-files` -> ruff pass, ruff-format pass
- `pytest tests/unit/ -x -q` -> **1595 passed**, 3 skipped, 0 failed
- No merge conflicts remaining

teng-lin:
Maintainer update: I took over finalization of this PR and rebased it onto current `main` (including #334). No behavioral changes were added beyond the existing two commits; this refreshes the branch so it can be reviewed/merged against the latest auth cookie-domain changes.\n\nVerification run from an isolated worktree:\n\n- `uv sync --extra dev --extra browser`\n- `uv run pytest tests/unit/test_auth.py tests/unit/test_client.py tests/integration/test_core.py -q` -> 247 passed\n- `uv run pytest tests/unit -q` -> 1619 passed, 5 skipped\n- `uv run pytest -q` -> 2166 passed, 9 skipped\n- `uv run ruff check src/ tests/`\n- `uv run ruff format --check src/ tests/`\n- `uv run mypy src/notebooklm`\n\nThe remaining `CHANGES_REQUESTED` review state appears stale relative to the current branch; the previously requested follow-ups are addressed in the rebased commits.

teng-lin:
Ran the requested polish pass across Codex, Claude, and Gemini, then reconciled the findings against the branch.

Confirmed and fixed:
- explicit `--storage` now wins over `NOTEBOOKLM_AUTH_JSON` when building the live cookie jar
- in-memory `AuthTokens(...)` no longer syncs cookies into the user's default storage file on `close()`
- `close()` still closes the `httpx` client even if cookie sync fails
- `.accounts.google.com` cookies normalized by `http.cookiejar` now round-trip through storage extraction
- removed the stale `fetch_tokens` alias in `session.py` and updated the tests to patch the real helper

I left the non-blocking style suggestions alone where they would just churn public/internal helper names or duplicate existing behavior.

Verification run from `.worktrees/takeover-276`:

```text
uv run pytest tests/unit/test_auth.py tests/unit/cli/test_helpers.py tests/unit/cli/test_session.py::TestLoginBrowserCookies tests/integration/test_core.py -q
315 passed

uv run ruff check src/ tests/
All checks passed!

uv run ruff format --check src/ tests/
140 files already formatted

uv run mypy src/notebooklm
Success: no issues found in 48 source files

uv run pytest -q
2170 passed, 9 skipped, 4 warnings
```

teng-lin:
Live-tested the auth/cookie path from the takeover worktree.

Notes:
- the existing default `~/.notebooklm/profiles/default/storage_state.json` had a SID cookie but was expired for real token fetch; NotebookLM redirected to Google sign-in
- refreshed credentials into a temporary storage file using local Chrome cookies via `uv run --extra cookies notebooklm login --browser-cookies chrome --storage <tmp>` so the real profile file was not modified

Live checks against the temp storage:

```text
fetch_tokens_with_domains(<tmp>) -> token fetch succeeded
AuthTokens.from_storage(<tmp>) -> succeeded
NotebookLMClient(auth, storage_path=<tmp>).notebooks.list() -> 24 notebooks
storage mode after token/client sync -> 0600
cookie domains preserved -> .google.com, .notebooklm.google.com, accounts.google.com, notebooklm.google.com
```

Also verified explicit storage precedence with a bogus env auth value:

```text
NOTEBOOKLM_AUTH_JSON='{"cookies":[]}' notebooklm --storage <tmp> list --json
```

That succeeded and returned 24 notebooks, confirming the CLI used the explicit storage file rather than the env var cookie jar.

teng-lin:
Follow-up from live testing: I noticed the e2e auth fixture was still manually building a flat `Cookie` header from `load_auth_from_storage()` and fetching tokens through a bare `httpx.AsyncClient`. That bypassed the new domain-preserving auth path this PR is meant to validate.

Updated `tests/e2e/conftest.py` so e2e auth now uses:

```python
AuthTokens.from_storage()
NotebookLMClient(auth_tokens, storage_path=auth_tokens.storage_path)
```

This means future e2e runs exercise the same domain-aware cookie jar and storage-path behavior as production code.

Verification:

```text
uv run ruff check src/ tests/
uv run ruff format --check src/ tests/
uv run mypy src/notebooklm
uv run pytest -q
2170 passed, 9 skipped, 4 warnings

NOTEBOOKLM_AUTH_JSON=<refreshed temp storage json> \
  uv run pytest tests/e2e/test_notebooks.py::TestNotebookOperations::test_list_notebooks -q
1 passed
```

teng-lin:
Thanks @AmanSuryavanshi-1 for the original fix and test coverage here, and thanks to everyone who helped review and validate the cross-domain cookie behavior. This now has the original implementation, polish fixes, live auth verification, updated e2e coverage, and a fully green CI matrix.

Merging this into `main`.

PR #257, opened 2026-04-06
fix: dedupe research import retries after timeout

voidborne-d:
## Summary
- dedupe the retry batch in `import_with_retry()` by checking which source URLs are already present in the notebook after a timeout
- stop retrying early when every source has already landed, instead of resubmitting the whole batch
- add regression tests covering partial and complete retry deduplication

## Testing
- `PYTHONPATH=src python3 -m pytest tests/unit/cli/test_helpers.py tests/unit/cli/test_research.py tests/unit/cli/test_source.py -q`

Closes #241.

## Summary by CodeRabbit

* **Bug Fixes**
  * Smarter retry behavior for importing sources: deduplicates and incrementally shrinks retry batches by reconciling with existing notebook sources, stops retrying when nothing remains, and continues retrying if reconciliation fails.

* **Tests**
  * Expanded async test coverage validating timeout/retry reconciliation, deduplication, early stop behavior, and resilience when source-listing fails.

PR #291, opened 2026-04-17
fix(cli): dedupe sources before IMPORT_RESEARCH retry (closes #241)

### LINKED ISSUE #241: Bug: research wait --import-all imports duplicate sources on timeout retry

lovebuilt:
## Summary

When `import_with_retry` in `cli/helpers.py` hits an `RPCTimeoutError`, it currently re-submits the full sources list unchanged. In practice the NLM server has often completed the import by then (the RPC timeout is client-side, not server-side cancellation), so each retry produces another full duplicate set. This is Issue #241.

This PR applies the same approach as PR #286: between retries, list the notebook's existing sources and filter `pending_sources` by URL. If nothing remains, return early. If listing fails, fall back to retrying the current pending list.

## Empirical evidence

Reproduced in an isolated venv, query: "Tailscale subnet router setup on Hetzner VPS", against the same Google account. Deep research, `research wait --import-all`.

| | v0.3.3 (no retry loop) | v0.3.4 (current main) | v0.3.4 + this PR |
|---|---|---|---|
| URLs discovered | 42 | 56–70 typical | 54 |
| Sources imported | 41 | 351–370 typical | 64 |
| Duplicates | 0 | 234–300 | 9 |
| Dupe ratio | 1.0× | 3–6× | 1.17× |

The residual 9 dupes in the patched run are a server-side race (sources in flight when the retry fires) rather than a client-side retry loop, so they'd need a different fix (e.g., server-side idempotency on IMPORT_RESEARCH). The ~25–33× reduction in client-induced duplication is what this PR delivers.

## Test results

4/4 existing `TestImportWithRetry` tests still pass with no modification. 1471 other unit tests pass; the 8 pre-existing Playwright-related failures and 11 errors in `tests/unit/cli/test_session.py` and `tests/unit/test_windows_compatibility.py` are identical on unpatched `main` in the same environment — unrelated to this change.

## Context / acknowledgment

Issue #241 has been open since 2026-04-03 and has bitten every user running deep research — in our case, deduplication was taking ~40% of every notebook-build session. PRs #286 and #257 proposed similar fixes earlier and were closed unmerged. #286 was closed with concern that the submitter was a burner account; that concern doesn't apply here (this is a long-standing personal account, and we've been running the fork in production).

Happy to adjust the diff — is there specific guidance on what blocked those earlier PRs? I kept the approach close to #286 because it was the cleanest and the filter-then-retry pattern reads naturally against the current CLI-only retry policy comment already in the function.

Closes #241.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source import reliability with smarter retry logic that detects and skips already-imported sources, reducing redundant processing.
  * Enhanced timeout handling during source imports with clearer status messages about remaining sources to process.

### DIFF
--- a/src/notebooklm/cli/helpers.py
+++ b/src/notebooklm/cli/helpers.py
@@ -90,34 +90,53 @@ async def import_with_retry(
 ) -> list[dict[str, str]]:
     """Retry research import on RPC timeouts with exponential backoff.
 
+    On retry, already-imported sources are filtered out to prevent duplicates.
+
     This is intentionally CLI-only policy. Library consumers calling
     `client.research.import_sources()` directly still get one-shot behavior.
     """
     started_at = time.monotonic()
     delay = initial_delay
     attempt = 1
+    pending_sources = list(sources)
 
     while True:
         try:
-            return await client.research.import_sources(notebook_id, task_id, sources)
+            return await client.research.import_sources(notebook_id, task_id, pending_sources)
         except RPCTimeoutError:
             elapsed = time.monotonic() - started_at
             remaining = max_elapsed - elapsed
             if remaining <= 0:
                 raise
 
+            # Filter out already-imported sources to prevent duplicates
+            try:
+                existing = await client.sources.list(notebook_id)
+                existing_urls = {s.url for s in existing if s.url}
+                pending_sources = [s for s in pending_sources if s.get("url") not in existing_urls]
+                if not pending_sources:
+                    logger.info("All sources already imported; no retry needed")
+                    if not json_output:
+                        console.print("[green]All sources already imported.[/green]")
+                    return []
+            except Exception as exc:
+                # If listing fails, retry with current pending list
+                logger.debug("Failed to list existing sources for deduplication: %s", exc)
+
             sleep_for = min(delay, max_delay, remaining)
             logger.warning(
-                "IMPORT_RESEARCH timed out for notebook %s; retrying in %.1fs (attempt %d, %.1fs elapsed)",
+                "IMPORT_RESEARCH timed out for notebook %s; retrying in %.1fs "
+                "(attempt %d, %.1fs elapsed, %d sources remaining)",
                 notebook_id,
                 sleep_for,
                 attempt + 1,
                 elapsed,
+                len(pending_sources),
             )
             if not json_output:
                 console.print(
                     f"[yellow]Import timed out; retrying in {sleep_for:.0f}s "
-                    f"(attempt {attempt + 1})[/yellow]"
+                    f"(attempt {attempt + 1}, {len(pending_sources)} remaining)[/yellow]"
                 )
             await asyncio.sleep(sleep_for)
             delay = min(delay * backoff_factor, max_delay)


teng-lin:
Thanks for the thorough PR body — the empirical before/after table (351–370 → 64 sources) made the triage much easier, and the proactive context about why this isn't a burner-account situation was appreciated.

Heads-up: PR #236 (@wanderernupur-arch) is also open for #241 and takes a slightly different approach. Rather than merge both, I'd like to pick the strongest one — so I thought I'd ping you and invite a look at #236, since you're clearly active and running this in production.

**How the two PRs compare:**

| | #291 (this PR) | #236 |
|---|---|---|
| Error handling on dedup probe failure | ✅ `try/except` around `sources.list()` falls back to original list | ❌ missing — network blip mid-retry aborts the whole import (I [reproduced this](https://github.com/teng-lin/notebooklm-py/pull/236#issuecomment-4267353449)) |
| New tests for dedup path | 0 | 7 including regression test for the timeout-retry path |
| Scope | only `import_with_retry` | also pre-import dedup in `source add-research` / `research wait` (makes `--import-all` idempotent) |
| URL normalization (trailing slash, UTM, etc.) | ❌ raw-string match | ❌ raw-string match |

Your empirical result of "9 residual duplicates in the patched run ... a server-side race" — when I reproduced it, at least some of those are actually client-side: NotebookLM's server sometimes canonicalizes URLs (trailing slash, query params), so raw-string dedup misses the match. A small `_normalize_url()` helper in the existing `_url_utils.py` would close that gap in both PRs.

**What would help me decide:**

Would you be open to looking at #236 and one of:
1. Porting its test suite (the regression test in particular) onto this PR, or
2. Porting your `try/except` guard onto #236 via a review comment there, so the winner has both.

Either way, responsiveness and iteration quality over the next few days will drive the final call. No rush — just being transparent about where things stand.

lovebuilt:
Hey — thanks so much for taking the time to write this up. Really awesome engagement, and the #236 comparison is super helpful.

Real talk: I'm not an OSS regular. This was my first drive-by PR, and I wouldn't have known about #236 without you pointing it out. Looking at it, their test suite is clearly better than what we shipped (we shipped zero) and the broader scope makes sense. I'm not trying to "win" this — I just want the bug gone for everyone.

On your suggestions:

1. **URL normalization**: makes sense. Our ~9 residual dupes being partly client-side is news to me but totally tracks. Happy to help get a `_normalize_url()` helper landed wherever it makes the most sense.

2. **Which PR wins**: honestly, whichever one you think is the stronger foundation. If you lean toward #236, I'd rather help improve it (leaving your `try/except` guard as a review comment over there) than try to bolt tests onto mine just to compete. If you want to go with #291, I'll happily port their regression test over.

3. **No pressure either way** — we're running the fork in production and it's stable. Take the time you need.

Just let me know which direction you'd prefer and I'll go that way. Appreciate you being so transparent about the decision process. Whatever you think is best!

teng-lin:
Closing this as superseded by #319, which has now merged and addresses #241 with server-side import state verification plus regression coverage. Thanks for helping move the duplicate-import issue forward.

lovebuilt:
Thanks for closing this out cleanly and for the kind words on the PR — glad #319 landed. Will pull the server-side fix into our pipeline when we upgrade.

PR #298, opened 2026-04-20
feat(auth): auto-refresh cookies via NOTEBOOKLM_REFRESH_CMD on expiry

### LINKED ISSUE #297: Auth expiry UX: let users plug in a cookie-refresh hook instead of re-running `notebooklm login`
## Problem

Google SID cookies expire frequently — I hit `notebooklm login` prompts multiple times a day in long-running agent workflows and automation. Every expiry today forces the human back to the browser, which breaks unattended use cases (cron jobs, `tmux`/CI sessions, Claude Code agents, etc.).

Current behavior in `src/notebooklm/auth.py`:

```python
async def fetch_tokens(cookies: dict[str, str]) -> tuple[str, str]:
    ...
    if is_google_auth_redirect(final_url):
        raise ValueError(
            "Authentication expired or invalid. "
            "Redirected to: " + final_url + "\n"
            "Run 'notebooklm login' to re-authenticate."
        )
```

There's no hook for the many legitimate refresh sources a user might already have — e.g. a browser-extension cookie vault (CookieCloud), a secrets manager that holds fresh `storage_state.json`, a custom `rookiepy`-based refresher. Users who *do* have one must either shell-wrap every CLI call with retry glue, or add fragile editor-level hooks that trigger only after the first failure.

## Proposal

Add an opt-in escape hatch via an env var, `NOTEBOOKLM_REFRESH_CMD`:

- If set and `fetch_tokens` raises a recognized auth-expiry `ValueError`, run the command, reload cookies from the default storage path, and retry once.
- If unset → behavior unchanged (no API break, no surprise for existing users).
- Single attempt guarded by a process-scoped flag so a broken refresh script can't cause an infinite loop.
- `fetch_tokens` is the single chokepoint every CLI path hits (`cli/helpers.py`, `cli/download.py`, `AuthTokens.from_storage`, etc.), so one wrap covers every command without touching each call site.

## Example user setup

```bash
# User already has a script that pulls fresh Google cookies from a vault
# and rewrites ~/.notebooklm/storage_state.json
export NOTEBOOKLM_REFRESH_CMD=/usr/local/bin/my-cookie-sync

notebooklm list   # auto-refreshes on expiry, no manual login
```

## Alternatives considered

1. **Auto-trigger `notebooklm login`** — no: login is interactive (opens browser), wrong for headless/CI.
2. **New public API** (e.g. `AuthTokens.from_storage_with_refresh`) — more surface area for the same outcome; requires every call site to migrate.
3. **Editor/shell hooks that retry after failure** — what I did before, and what I want to replace. Fragile (depends on matching error strings in the right wrapper), only covers one caller, and the first command still fails before the hook kicks in.

Happy to open a PR implementing the above — I have it working locally with ~100 LOC change + 4 unit tests. Let me know if you'd prefer a different shape (e.g. callable-based API instead of env-var based, or gated behind a config setting) and I'll adjust.

JaggerH:
Closes #297.

## Summary

- Adds opt-in auto-refresh of Google auth cookies when `fetch_tokens` detects an expired session.
- Opt in by setting `NOTEBOOKLM_REFRESH_CMD` to a shell command that rewrites the default storage file (e.g. a sync script that reads from a browser-extension cookie vault, a secrets manager, or a `rookiepy` exporter).
- When unset, behavior is unchanged — no API break.
- One wrap in `fetch_tokens` covers every CLI path: `AuthTokens.from_storage`, `cli/helpers.get_client`, `cli/download.py`, etc.
- Retry is one-shot per process (process-scoped env flag) so a broken refresh script can't cause an infinite loop.

## Why at `fetch_tokens`?

It is the single point every auth-using caller funnels through. Wrapping here avoids editing five CLI call sites, keeps the public API stable, and makes the opt-in apply uniformly to notebook ops, chat, downloads, and research.

## Behavior matrix

| Env var | fetch_tokens outcome | Behavior |
|---|---|---|
| unset | success | identical to today |
| unset | auth redirect | identical to today (raises `ValueError`) |
| set   | success | identical to today (no extra calls) |
| set   | auth redirect → refresh succeeds | runs `$NOTEBOOKLM_REFRESH_CMD`, reloads cookies from disk, retries once (logs a `WARNING`) |
| set   | auth redirect → refresh still stale | re-raises the original `ValueError` after one retry (no loop) |
| set   | refresh command exits non-zero | `RuntimeError` with captured stderr |

## In-place cookie mutation

On refresh, the passed `cookies` dict is updated in place so callers who kept a reference (`AuthTokens.from_storage`, `cli/helpers.get_client`, `cli/download.py`) see fresh values without needing code changes. This is documented in the updated docstring. I considered returning cookies instead but that would widen the signature and require touching every call site.

## Tests

Four new unit tests in `TestFetchTokensAutoRefresh`:

- `test_no_refresh_when_env_unset` — regression check: absent env var = exact current behavior
- `test_refresh_retries_once_and_succeeds` — happy path: mock auth failure → run real shell script that rewrites storage → retry succeeds, cookies mutated in place
- `test_refresh_does_not_loop` — safety: refresh that fails to fix auth must not loop, must propagate original error
- `test_refresh_cmd_nonzero_exit_becomes_runtime_error` — failed refresh surfaces as RuntimeError with captured stderr

Full auth suite passes (144/144), `ruff check` and `ruff format --check` clean.

## Test plan

- [x] `pytest tests/unit/test_auth.py` — 144 passed
- [x] `ruff check src/ tests/` — clean
- [x] `ruff format --check src/ tests/` — clean
- [x] End-to-end smoke test: installed the same patch in my local pipx venv, corrupted SID, ran `notebooklm list` with `NOTEBOOKLM_REFRESH_CMD=/path/to/cookie-sync` → refresh triggered, cookies replaced on disk, command succeeded without manual `notebooklm login`
- [ ] Maintainer review for API shape preference (env-var vs. callable, inline mutation vs. return value)

Happy to iterate if you'd prefer a different surface (e.g. gating via config file instead of env var, or exposing as `AuthTokens.from_storage(auto_refresh=True)` keyword).

## Summary by CodeRabbit

* **New Features**
  * Added opt-in automatic session refresh when authentication expires. Configure the `NOTEBOOKLM_REFRESH_CMD` environment variable to specify a custom refresh command. When triggered, the system updates authentication state and automatically retries operations once.

* **Tests**
  * Added comprehensive test coverage for session refresh scenarios, including success cases, retry logic, and error handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Auto-refresh on auth expiry** - `fetch_tokens` now optionally runs a user-provided refresh command when a Google session cookie has expired, reloads cookies from storage, and retries once. Opt in by setting the `NOTEBOOKLM_REFRESH_CMD` environment variable to a script that rewrites `~/.notebooklm/storage_state.json` (e.g. a sync script reading from a cookie vault). Covers every CLI entry point without changing the public API. A process-scoped flag prevents retry loops.
+
 ## [0.3.4] - 2026-03-12
 
 ### Added
--- a/src/notebooklm/auth.py
+++ b/src/notebooklm/auth.py
@@ -31,6 +31,8 @@ import json
 import logging
 import os
 import re
+import shlex
+import subprocess
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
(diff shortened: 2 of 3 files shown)


teng-lin:
Good idea. Windows CI is failing because the test creates a `refresh.sh` script that Windows cannot execute (`[WinError 2]`). Please make the test portable — either invoke `sys.executable -c "..."` as the refresh command, or write a `.bat` on Windows and `.sh` on POSIX behind `os.name`. Once Windows is green this is in good shape.

teng-lin:
Thank you @JaggerH — your design in #297 (opt-in refresh hook via env var, single retry guard, profile/storage env wiring exposed to refresh scripts) is the core of what's about to ship. Closing this in favor of #336, which keeps your design and adds the surrounding pieces:

- rebased onto the post-#276 jar-based auth flow with `asyncio.to_thread(subprocess.run)` for Windows portability under `WindowsSelectorEventLoopPolicy`, plus `asyncio.Lock` + a per-storage-path generation counter + a `ContextVar` for concurrent-fetch safety;
- the proactive layers your refresh hook makes recoverable: per-call `CheckCookie` poke (closes #312), `notebooklm auth refresh` for OS-scheduled keepalive, `examples/refresh_browser_cookies.py` as a turnkey `NOTEBOOKLM_REFRESH_CMD` target, and cross-OS scheduler recipes (launchd / systemd / cron / Windows Task Scheduler / k8s CronJob) in `docs/troubleshooting.md`;
- a follow-up #341 layers an opt-in periodic background poke on `NotebookLMClient` for long-lived sessions.

Thanks for the well-scoped issue and the clean draft to take over.

PR #310, opened 2026-04-24
fix(chat): emit refs only from winning chunk (fixes #300)

### LINKED ISSUE #300: result.references has more than 600 elements although the chat response contains about 20 references
## Description
 result.references is inflated.

## Steps to Reproduce

1. nb = await client.notebooks.get(notebook_id)
2. result = await client.chat.ask(nb.id, f"Write a summary of the chapter {chap_title}")
3. accessing result.references...

## Expected Behavior
result.references should include only as many elements as are referenced in the chat answer. For each element ref.citation_number should correspond to the actual number in the chat answer (eg. [1], [5], etc.)

## Actual Behavior
result.references contains about 600 elements, and ref.citation_number corresponds to the index of the element in result.references + 1. So [ref.citation_number for ref in result.references] produces a list of 1, 2, 3, ... 600.

ltianyi992:
## Problem

`result.references` returned 600+ entries when a chat response contained only ~20 cited sources (issue #300).

**Root cause:** `_parse_ask_response_with_references` processes multiple streaming chunks via `process_chunk`. Each chunk that contains an answer also includes a full copy of the reference list. The old code called `all_references.extend(refs)` for *every* chunk, so references multiplied by chunk count:

```
6 chunks × 100 refs per chunk = 600 entries in result.references
```

Citation numbers were then assigned sequentially across all accumulated refs (1, 2, 3 … 600), not based on actual citations in the answer.

## Fix

Track refs alongside each best-answer candidate instead of accumulating from all chunks. Only the refs from the chunk that produced the longest marked answer (the "winning" chunk) are returned.

```python
# Before (in process_chunk):
all_references.extend(refs)   # accumulated from every chunk

# After:
if is_answer and len(text) > len(best_marked_answer):
    best_marked_answer = text
    best_marked_refs = refs   # refs travel with their answer candidate
```

The final `return` now uses `final_refs` (from the winning chunk) instead of `all_references`.

## Tests

Added `TestMultiChunkReferenceInflation` with three regression tests:

- `test_multi_chunk_does_not_inflate_references` — 2 chunks → 1 ref, not 2
- `test_many_chunks_still_single_ref_set` — 6 chunks → 1 ref, not 6 (mirrors the bug report numbers)
- `test_refs_from_longest_winning_chunk` — refs belong to the longer (winning) chunk, not the shorter one

All 21 existing chat-reference tests still pass; the 8 pre-existing `playwright`-related failures in `test_session.py` and `test_windows_compatibility.py` are unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed reference tracking so only citations from the selected answer candidate are returned, preventing duplicated or inflated references when responses arrive in multiple streamed chunks.

* **Tests**
  * Added unit tests covering streamed multi-chunk scenarios to ensure the parser selects the winning chunk and returns a single, correct reference list.

### DIFF
--- a/src/notebooklm/_chat.py
+++ b/src/notebooklm/_chat.py
@@ -448,20 +448,22 @@ class ChatAPI:
 
         lines = response_text.strip().split("\n")
         best_marked_answer = ""
+        best_marked_refs: list[ChatReference] = []
         best_unmarked_answer = ""
-        all_references: list[ChatReference] = []
+        best_unmarked_refs: list[ChatReference] = []
         server_conv_id: str | None = None
 
         def process_chunk(json_str: str) -> None:
-            """Process a JSON chunk, updating best answers and all_references."""
-            nonlocal best_marked_answer, best_unmarked_answer, server_conv_id
+            """Process a JSON chunk, updating best answer candidates and their refs."""
+            nonlocal best_marked_answer, best_marked_refs, best_unmarked_answer, best_unmarked_refs, server_conv_id
             text, is_answer, refs, conv_id = self._extract_answer_and_refs_from_chunk(json_str)
             if text:
                 if is_answer and len(text) > len(best_marked_answer):
                     best_marked_answer = text
+                    best_marked_refs = refs
                 elif not is_answer and len(text) > len(best_unmarked_answer):
                     best_unmarked_answer = text
-            all_references.extend(refs)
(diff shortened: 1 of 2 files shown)


teng-lin:
Thanks @ltianyi992 for the focused fix and the regression coverage for the streaming reference inflation case. Merged.

PR #319, opened 2026-05-02
fix(cli): verify server-side state on import timeout to prevent duplicate inflation

### LINKED ISSUE #241: Bug: research wait --import-all imports duplicate sources on timeout retry

jmenzler:
## Summary

Closes #241.

`import_with_retry` re-submits the entire source list on every RPC timeout, which inflates the in-notebook source count by the retry attempt count. In practice this routinely turns a 60–80-source deep-research import into 300+ sources across 5–6 retries, because the server has typically *already* written the sources by the time the 30s client deadline fires.

This change snapshots the notebook's source URLs before importing and, on `RPCTimeoutError`, probes `sources.list` to compute the delta. If the delta covers the requested URLs (either as a subset match, or by count — the latter is robust to server-side URL normalization), the timeout is treated as a verified success and the retry loop exits. The probe is wrapped in a defensive `try/except`, so any failure (e.g. transient server error) falls back cleanly to the original retry path.

## Approach vs. issue's suggested fix

Issue #241 suggests filtering `pending_sources` to remove already-imported URLs and then retrying with the smaller set. This works but still issues an extra IMPORT_RESEARCH RPC even when the first call succeeded.

This PR takes the slightly more aggressive route: when the post-timeout probe shows the requested URLs are present, **skip the retry entirely**. In the common case (server succeeded, RPC client timeout fired) this collapses to one import + one list call, with zero duplication. The partial-import case (rare in practice; not observed in the reporter's reproduction or in my own) still retries via the existing path because the subset/delta check fails.

If maintainers prefer the partial-retry variant, I'm happy to switch — let me know.

## Reproduction

A 64-source deep-research import on a fresh notebook reliably inflated to ~300 sources before this patch (6 retry attempts × ~50 sources each, with some server-side dedup), confirmed against the live API. Same scenario after the patch: ~64 sources, single import attempt, the verified-success notice prints, and source-list dedup is no longer needed downstream.

## Tests

- 4 existing tests in `TestImportWithRetry` continue to pass. They use a vanilla `MagicMock` for `client`, so the new pre-import `sources.list` snapshot raises `TypeError` on the un-async-mocked attribute, hits the defensive `except Exception`, and falls back to the original retry behavior — matching prior expectations.
- 3 new tests cover the new behavior:
  - `test_skips_retry_when_server_state_shows_import_succeeded` — happy path: URL-set match → no retry, no sleep, single import attempt, synthesized success list.
  - `test_skips_retry_when_delta_size_matches_requested` — URL-normalization-safe path: server stores a normalized URL, delta count match still treats as success.
  - `test_retries_when_server_state_shows_no_progress` — fallback: when the probe shows no progress, retry as before.

`pytest tests/unit/cli/` → 574 passed locally. `ruff check` and `ruff format --check` clean.

## Test plan

- [x] Existing CLI unit tests pass (574/574)
- [x] New tests added for verified-success, URL-normalization, and no-progress cases
- [x] `ruff check` clean
- [x] `ruff format --check` clean
- [x] Reproduced the bug on the live API and verified the patch resolves it

## Notes

- AI-assisted contribution per CONTRIBUTING.md — I reviewed, ran, and tested everything before submitting; happy to revise on any point.
- The bug only affects the CLI helper; library consumers calling `client.research.import_sources()` directly are unchanged (no retry policy there).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI import resilience: on RPC timeouts the tool now verifies server-side import state to avoid duplicate entries and unnecessary retries, and surfaces a warning/notice when imports are confirmed.

* **Tests**
  * Added unit tests covering timeout handling, server-state verification branching, exclusion of pre-existing sources, inclusion of non-URL sources, and retry fallback.

### DIFF
--- a/src/notebooklm/cli/helpers.py
+++ b/src/notebooklm/cli/helpers.py
@@ -105,18 +105,26 @@ async def import_with_retry(
 
     requested_urls = {url for s in sources if isinstance((url := s.get("url")), str) and url}
 
-    # Snapshot baseline source URLs so we can detect server-side success on
-    # timeout by computing the delta after each failed import RPC.
+    # Snapshot baseline source URLs AND IDs so we can detect server-side success
+    # on timeout by computing the delta after each failed import RPC. URLs drive
+    # the detection condition (was the import seen on the server?); IDs filter
+    # the return value to only the truly new sources, which avoids reporting
+    # pre-existing rows as "imported" and also captures non-URL sources such as
+    # research-report entries.
+    baseline_urls: set[str] | None
+    baseline_ids: set[str] | None
     try:
         baseline = await client.sources.list(notebook_id)
-        baseline_urls: set[str] | None = {src.url for src in baseline if src.url}
-    except Exception as snapshot_exc:
+        baseline_urls = {src.url for src in baseline if src.url}
+        baseline_ids = {src.id for src in baseline}
+    except Exception as snapshot_exc:  # noqa: BLE001 - probe is best-effort; any failure falls back to legacy retry
         logger.debug(
             "Pre-import sources.list snapshot failed for %s: %s",
             notebook_id,
             snapshot_exc,
         )
         baseline_urls = None
+        baseline_ids = None
 
     while True:
         try:
@@ -128,11 +136,12 @@ async def import_with_retry(
             # Verify server-side state before retrying. The IMPORT_RESEARCH RPC
             # frequently times out at the client (30s) after a successful
             # server-side write; retrying then duplicates every source.
-            if baseline_urls is not None and requested_urls:
+            if baseline_urls is not None and baseline_ids is not None and requested_urls:
                 try:
                     current = await client.sources.list(notebook_id)
                     current_urls = {src.url for src in current if src.url}
                     delta_urls = current_urls - baseline_urls
+                    new_ids = {src.id for src in current} - baseline_ids
                     # Either all requested URLs are now visible, or the delta
                     # size matches what we tried to import. The latter is
                     # robust to server-side URL normalization.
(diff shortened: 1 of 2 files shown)


jmenzler:
Thanks for the careful review @gemini-code-assist + @coderabbitai — both bots converged on the same real bug: the URL-subset check could spuriously match when a requested URL was already in the baseline, and the return filter dropped non-URL sources (research reports). Pushed [4cad509](https://github.com/jmenzler/notebooklm-py/commit/4cad509) which:

- Snapshots `baseline_ids` alongside `baseline_urls` and computes `new_ids = current_ids - baseline_ids`
- Keeps URL match as the *detection* condition but uses `src.id in new_ids` as the *return filter*
- Adds two regression tests: pre-existing-URL false-success, and non-URL (report) inclusion
- Adds `# noqa: BLE001` with rationale on the two best-effort probe `except Exception` blocks (probes are intentionally catch-all so they fall back to legacy retry rather than propagating)

`pytest tests/unit/cli/`: 576 passed. Lint + format clean.

Independent confirmation that the patch is doing useful work in the wild: I've been hammering the CLI today on a multi-notebook research run, and the `Import RPC timed out, but server-side verified N new sources — skipping retry.` notice has fired several times — turning what would have been ~6× duplicate inflation into clean single imports. There's still a corner case where the baseline `sources.list` returns `NoneType` (warning: "Sources data ... is not a list (type=NoneType)") on freshly-created notebooks; in that case the patch correctly falls back to legacy retry. Probably worth a separate issue + fix to retry the snapshot probe with a small backoff for newly-created notebooks, but that's out of scope for this PR.

teng-lin:
Merged — thanks @jmenzler! Great repro and clear write-up. The duplicate-inflation case you hit was a real production pain point.

A multi-model review (Claude + Codex + Gemini) flagged some edge cases in the verification heuristic — chiefly that `requested_urls.issubset(current_urls)` and `len(delta_urls) >= len(requested_urls)` can both false-positive when a requested URL pre-existed or when concurrent activity adds unrelated sources during the timeout window. None of those are worse than pre-#319 behavior, so I shipped this as-is and tracked the tightening in #320 with a recommended fix sketch.

If you'd like to take the follow-up, it's all yours — otherwise I or another maintainer will pick it up. Either way: nice fix.

PR #404, opened 2026-05-12
test: add e2e composition test for variant-aware CAS rejection

### LINKED ISSUE #387: test: add end-to-end composition test for variant-aware rejected-baseline preservation
Follow-up from PR #363 (the snapshot/delta cookie-save machinery).

## Context

`advance_cookie_snapshot_after_save` (in `src/notebooklm/auth.py`) is variant-aware when preserving the baseline for CAS-rejected keys (handles leading-dot domain variants). Unit-tested at `tests/unit/test_auth_cookie_save_race.py::test_rejected_variant_preserves_original_baseline_variant`, which drives the helper directly with hand-built snapshots.

The variant-aware CAS path is separately unit-tested at `test_cas_protects_across_leading_dot_variant`.

## Gap

There's no test exercising the **composition** of both pieces: a save that actually CAS-rejects on a variant key end-to-end through the real `from_storage` / `save_cookies` plumbing, followed by a second save that successfully proceeds without re-clobbering the sibling write.

## Suggested test

Wire `AuthTokens.from_storage` → `ClientCore` → two saves where:

1. Disk row uses one domain variant (e.g., `accounts.google.com`).
2. Local jar uses the leading-dot variant (`.accounts.google.com`).
3. A sibling writes a different value to disk between our open and our first save.
4. First save CAS-rejects on the variant match → returns False, baseline preserved variant-aware.
5. Second save (after local jar's value matches disk via Set-Cookie) successfully proceeds.

Asserts:

- First save: `result.ok is False`, sibling's value survives on disk.
- Second save: rejected-key baseline was preserved through the variant lookup, save advances cleanly.

## Priority

Low. The unit tests for each component would already catch a regression in either piece individually. This is defense-in-depth, not an actionable correctness hole.

## Refs

Flagged by the Claude code-reviewer agent during the review of commit 1494e33 as a low-confidence suggestion explicitly marked \"not reported as a defect\".

teng-lin:
## Summary

Adds the end-to-end composition test issue #387 asked for. Existing unit tests cover the variant-aware CAS path (`test_cas_protects_across_leading_dot_variant`) and the variant-aware baseline preservation (`test_rejected_variant_preserves_original_baseline_variant`) individually by driving the helpers directly. This new test wires the full plumbing.

`TestCASVariantAware::test_variant_aware_cas_rejection_then_recovery_e2e`:

1. Disk row for `OSID` uses the bare-host variant (`accounts.google.com`).
2. `AuthTokens.from_storage` loads the jar; the mocked token fetch re-keys `OSID` to the leading-dot variant (`.accounts.google.com`) and a sibling process rewrites disk's bare-host row to a third value between snapshot and save.
3. The pre-client save runs through real `save_cookies_to_storage`: variant-aware CAS spots the disk drift, rejects the dotted delta, sibling write survives on disk. Real `advance_cookie_snapshot_after_save` then preserves the bare-host baseline.
4. A real `ClientCore` stands up (inheriting the variant-aware preserved baseline), a Set-Cookie aligns the in-memory jar with disk, and a second `ClientCore.save_cookies` runs through the real lock + `to_thread` + baseline-advance plumbing. It also CAS-rejects (because disk still diverges from the preserved baseline) — and the preserved baseline must stay at its original value rather than absorbing the rejected delta.

After triple-model review (codex/gemini/claude), the final post-save assertion was tightened from \`get(bare_key) is not None\` to \`.value == \"OLD\"\` so a regression that silently absorbs a rejected delta into the baseline would be caught.

Closes #387.

## Test plan

- [x] \`ruff format --check .\` clean
- [x] \`ruff check .\` clean
- [x] \`mypy src/notebooklm --ignore-missing-imports\` clean
- [x] New test passes: \`pytest tests/unit/test_auth_cookie_save_race.py -x\` → 49 passed
- [x] Full unit suite passes: \`pytest tests/unit/\` → 1928 passed, 5 skipped

## Summary by CodeRabbit

* **Tests**
  * Added an async regression test validating cookie-domain variant handling and CAS rejection/recovery across token refresh and client save cycles, ensuring on-disk cookie consistency through concurrent in-memory and on-disk updates.

* **Bug Fixes**
  * Improved snapshot/delta merge so a sibling-process write is treated as a CAS conflict only when on-disk value truly differs from both the snapshot and the intended update, allowing convergence to proceed when disk already matches the update.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/404)

### DIFF
--- a/src/notebooklm/auth.py
+++ b/src/notebooklm/auth.py
@@ -1642,7 +1642,10 @@ def _merge_cookies_with_snapshot(
             # CAS-guard for value updates: if our snapshot had this key in any
             # leading-dot variant and disk's current value differs from the
             # snapshot value, a sibling process has rewritten the row between
-            # our open and our save. Preserve their write rather than clobber.
+            # our open and our save. Preserve their write rather than clobber,
+            # unless disk has already converged to our current value; in that
+            # case the save intent is satisfied and the caller may advance its
+            # baseline.
             # Variant-aware lookup mirrors the delta match above: if the snapshot
             # was keyed on ``accounts.google.com`` but the matched delta key is
             # the leading-dot variant, a plain ``.get(matched_delta_key)`` would
@@ -1656,7 +1659,11 @@ def _merge_cookies_with_snapshot(
                 None,
             )
             stored_value = stored_cookie.get("value")
-            if snapshot_entry is not None and stored_value != snapshot_entry.value:
+            if (
+                snapshot_entry is not None
+                and stored_value != snapshot_entry.value
+                and stored_value != matched_delta_cookie.value
+            ):
                 logger.debug(
                     "Skipped CAS-guarded value update of %s on %s: disk value "
                     "differs from snapshot (sibling write preserved)",
--- a/tests/unit/test_auth_cookie_save_race.py
+++ b/tests/unit/test_auth_cookie_save_race.py
@@ -1517,10 +1517,13 @@ class TestCASVariantAware:
                 "divergence through the leading-dot variant"
             )
             assert core._loaded_cookie_snapshot is not None
-            assert core._loaded_cookie_snapshot.get(bare_key) is not None, (
-                "After the second save the variant-aware baseline preservation "
-                "must keep the bare-host entry so subsequent rotations can "
-                "still be CAS-checked against the original snapshot variant"
+            assert core._loaded_cookie_snapshot[bare_key].value == "OLD", (
+                "After the second save (which also CAS-rejects, because disk "
+                "still diverges from baseline) the variant-aware baseline "
+                "preservation must keep the bare-host entry AT its original "
+                "value — a regression that absorbed the rejected dotted "
+                "delta's value into the baseline would silently disable CAS "
+                "protection for the next rotation"
             )
         finally:
             await core.close()
--- a/tests/unit/test_auth_cookie_save_race.py
+++ b/tests/unit/test_auth_cookie_save_race.py
@@ -1408,8 +1408,10 @@ class TestCASVariantAware:
         )
 
     @pytest.mark.asyncio
-    async def test_variant_aware_cas_rejection_then_recovery_e2e(self, tmp_path, monkeypatch):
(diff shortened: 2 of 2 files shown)


teng-lin:
Addressed coderabbitai's nitpick in commit 4844f49: renamed `test_variant_aware_cas_rejection_then_recovery_e2e` → `test_variant_aware_cas_rejection_then_recovery_through_real_plumbing`. The `_e2e` suffix conflicted with the repo's convention where `tests/e2e/` is reserved for `@pytest.mark.e2e` tests that hit real APIs. This test mocks at the fetch boundary, so it correctly stays under `tests/unit/` and the new name makes that consistent.

teng-lin:
Fixed in commit 24c8e7b.

The issue was that a prior CAS rejection left the baseline at the original bare-host value (`OLD`). When a later Set-Cookie aligned the in-memory dotted cookie to the sibling value already on disk (`SIBLING`), the save still treated disk-vs-baseline divergence as another CAS rejection, so the baseline never recovered and future rotations remained stuck.

The merge path now treats `disk value == current cookie value` as converged success, even when disk differs from the old snapshot. That allows `ClientCore.save_cookies()` to advance the baseline after the aligned save. I also tightened the composition test so it asserts the recovered baseline (`SIBLING`) and then verifies a follow-up `NEXT` rotation persists through the variant-aware lookup.

Validation run locally:
- `uv run --extra dev python -m pytest tests/unit/test_auth_cookie_save_race.py::TestCASVariantAware::test_variant_aware_cas_rejection_then_recovery_through_real_plumbing -q`
- `uv run --extra dev python -m pytest tests/unit/test_auth_cookie_save_race.py -q`
- `uv run --extra dev python -m pytest tests/unit -q`
- `uv run --extra dev ruff check src/notebooklm/auth.py tests/unit/test_auth_cookie_save_race.py`
- `uv run --extra dev ruff format --check src/notebooklm/auth.py tests/unit/test_auth_cookie_save_race.py`

PR #478, opened 2026-05-14
feat(migration): single-writer migrate_to_profiles() under filelock (T5.B)

teng-lin:
## Summary

Two `notebooklm` CLI invocations starting under a fresh home (e.g., a container start-up race) previously both ran the copy + delete migration concurrently, producing partial state visible to either. This PR wraps the migration body in a cross-process `filelock` rooted at `<home>/.migration.lock` and re-checks the `.migration_complete` marker inside the lock so the loser of the race no-ops cleanly.

- New exception `MigrationLockTimeoutError(RuntimeError)` for the >30s wait case; raw `filelock.Timeout` is wrapped so callers see a domain-specific failure.
- Legacy-source unlink/rmtree is now tolerant of already-removed paths so a concurrent migration that crashed mid-run between copy and marker write does not crash the next runner.
- Marker re-check uses `default_dir / _MIGRATION_MARKER` (the path where `migration.py` actually writes the marker), not the home root.

Closes the T5.B item in `.sisyphus/plans/tier-5-implementation.md`.

## Test plan

- [x] `uv run ruff format . && uv run ruff check . && uv run mypy src/notebooklm --ignore-missing-imports`
- [x] `uv run pytest tests/unit -x -q` (2562 passed, 5 skipped)
- [x] New `tests/unit/test_migration_lock.py` covers:
  - Two-thread race produces exactly one `True` + one `False`, profile dir + marker intact, originals removed exactly once.
  - Pre-existing partial state (legacy sources already unlinked, marker not yet written) does not crash the loser.
  - Subsequent invocations are clean no-ops (lock state from first run does not poison second acquisition; marker survives).
  - Held lock surfaces `MigrationLockTimeoutError` with the underlying `filelock.Timeout` preserved as `__cause__`.
  - `MigrationLockTimeoutError` is a `RuntimeError` subclass (matches the existing `RuntimeError` pattern in `auth.py`).

### DIFF
--- a/src/notebooklm/migration.py
+++ b/src/notebooklm/migration.py
@@ -8,6 +8,8 @@ The migration is:
 - Idempotent: safe to run multiple times
 - Crash-safe: uses copy-then-delete with marker file
 - Non-destructive: originals kept until all copies succeed
+- Single-writer: cross-process ``filelock`` serializes concurrent CLI
+  invocations so a container start-up race cannot interleave copies.
 """
 
 import json
@@ -15,16 +17,29 @@ import logging
 import shutil
 import sys
 
+from filelock import FileLock, Timeout
+
 from ._atomic_io import atomic_update_json
 from .paths import get_config_path, get_home_dir
 
 logger = logging.getLogger(__name__)
 
 _MIGRATION_MARKER = ".migration_complete"
+_MIGRATION_LOCK = ".migration.lock"
+_MIGRATION_LOCK_TIMEOUT = 30.0
 
 # Legacy files that should be moved into profiles/default/
 _LEGACY_FILES = ["storage_state.json", "context.json"]
 _LEGACY_DIRS = ["browser_profile"]
+
+
+class MigrationLockTimeoutError(RuntimeError):
+    """Raised when ``migrate_to_profiles()`` cannot acquire the migration lock.
+
+    Surfaces the underlying :class:`filelock.Timeout` so callers can
+    distinguish a stuck peer from other migration failures (e.g., a CI
+    runner whose previous job left a stale process holding the lock).
+    """
 
 
 def _has_legacy_files(home) -> bool:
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
**Scope addendum from codex review of plan v3:**

Plan v3 (`.sisyphus/plans/tier-5-implementation.md`) calls out a crash gap that the current implementation may miss:

`migration.py:54-56` short-circuits when `profiles/` exists and no legacy files remain — BEFORE the lock would be acquired. A process that copied + deleted legacy files but crashed before writing `.migration_complete` would today look 'migrated' via this heuristic, so the next run never gets a chance to write the marker.

Could you:
1. **Move the 'already-migrated' short-circuit INSIDE the lock**, making marker presence the authoritative check: `marker = default_dir / _MIGRATION_MARKER; if marker.exists(): return False`.
2. Add a **crash-after-delete-no-marker recovery test**: leave `profiles/default/` present + legacy files deleted + `_MIGRATION_MARKER` MISSING. Calling `migrate_to_profiles()` must acquire the lock and write the marker idempotently without erroring.

Today the line-54 heuristic falsely short-circuits this recovery path.

teng-lin:
Thanks @claude — addressed the actionable items in commit 27b5ac5:

- **Item 1 (double `get_home_dir()`)** — collapsed to a single `home = get_home_dir(create=True)` call. The ordering intent is preserved by a single `create=True` invocation.
- **Item 5 (thread-vs-process layer)** — added a docstring note to `test_concurrent_threads_only_one_migrates` acknowledging that the thread-based test exercises `filelock`'s internal `threading.Lock` rather than the OS-level `fcntl`/`LockFileEx` layer. The single-writer invariant is identical in both cases; the comment helps future readers see the unit-test tradeoff.

Leaving the following as-is:

- **Item 2 (two `_has_legacy_files()` calls)** — the duplicate guard preserves the original code path; the I/O cost is negligible (two `stat()` calls on a no-op return) and unifying the conditional risks subtly changing semantics in the fresh-install branch.
- **Item 3 (`MigrationLockTimeoutError` propagation through `ensure_profiles_dir`)** — explicit design choice per the PR description. A 30s timeout fires only when a sibling process is genuinely stuck; surfacing the failure (rather than logging-and-continuing into a half-migrated state) is the intended behavior.
- **Item 4 (fresh install does not write marker)** — pre-existing and out of scope for this PR. The `profiles_dir.exists()` guard handles it correctly.
- **Item 6 (`clear=True` in `patch.dict`)** — matches the existing convention in `tests/unit/test_migration.py` (e.g., line 34). Consistency with the sibling test file outweighs the micro-purity gain here.

PR #500, opened 2026-05-14
feat(cli): add --json to chat configure (I3)

teng-lin:
## Summary
- Adds `--json` to `notebooklm configure` (alias `chat configure`) so the command produces a stable, machine-readable shape on both code paths (`--mode` and `--persona` / `--response-length`).
- The `goal` field in the JSON output is exposed as the lowercase `ChatGoal` enum name (e.g. `"custom"`) rather than the underlying RPC integer, so the JSON contract does not leak an internal detail Google could change.
- Non-JSON output is byte-for-byte unchanged.

## Plan Reference
- Task: `.sisyphus/plans/cli-ux-remediation/phase-2.md` § P2.T5
- Audit row: I3 (CLI UX remediation — `--json` rollout)

## Acceptance
- `chat configure --json` emits structured JSON.
- Smoke tests in `tests/unit/cli/test_chat.py::TestConfigureJsonOutput` cover both the mode path and the persona/length path.
- Coverage gate (≥90%) holds; project total at 92.60%.

## JSON shape
```jsonc
// configure --mode learning-guide --json
{ "notebook_id": "nb_123", "mode": "learning-guide", "configured": true }

// configure --persona "..." --response-length longer --json
{
  "notebook_id": "nb_123",
  "mode": null,
  "goal": "custom",
  "persona": "...",
  "response_length": "longer",
  "configured": true
}
```

## Test plan
- [x] `uv run ruff check .` — All checks passed
- [x] `uv run mypy src/notebooklm --ignore-missing-imports` — Success: no issues found in 58 source files
- [x] `uv run pytest --cov=src/notebooklm --cov-fail-under=90` — 3464 passed, 11 skipped; coverage 92.60%
- [x] Manual: `notebooklm configure --help` shows `--json` flag
- [x] Manual diff inspection: no behavioural change to non-JSON output

## Summary by CodeRabbit

* **New Features**
  * Added a `--json` flag to `configure` (alias `chat configure`) to emit stable machine-readable JSON including notebook_id, mode or custom persona/response settings, and configured: true for automation.

* **Tests**
  * Added unit tests validating JSON output for mode selection, custom persona/response-length, and no-flag scenarios.

* **Documentation**
  * CLI reference updated to document `configure --json` usage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/500)


teng-lin:
@claude review

teng-lin:
Thanks for the review @claude. Addressed in commit b8176d4:

**1. docs/cli-reference.md gap** — Added a `configure --json` row right after the existing `configure` row, parallel to how `ask --json` is documented (line 101). Doc table now matches the shipped feature.

**2. No-flags JSON test** — Added `test_configure_no_flags_json` in `TestConfigureJsonOutput`. Asserts the all-null shape (`{notebook_id, mode: null, goal: null, persona: null, response_length: null, configured: true}`) so the contract is locked in for automation scripts.

**3. JSON shape asymmetry** — Not changing. Push-back rationale:

The two paths describe genuinely different state. The mode-path JSON deliberately omits `goal`/`persona`/`response_length` because `--mode` selects a *bundle* (e.g. `learning-guide` => `goal=LEARNING_GUIDE` + `response_length=LONGER`) and exposing the unbundled internals there would (a) duplicate information already encoded in `mode`, and (b) lock the JSON contract to today's bundle composition, which would break the moment Google rebalances `learning-guide` to e.g. `response_length=DEFAULT`. The `mode` discriminator (string vs `null`) tells consumers which schema variant they got — the same pattern Click uses for tagged-union outputs and the same pattern the existing `chat.set_mode` vs `chat.configure` API split uses.

If a downstream caller needs uniform key access, the cheap workaround is `data.get("goal")` instead of `data["goal"]`. I'd rather keep the JSON honest about which RPC was actually called than paper over the asymmetry.

Nit on module-level `import json` — left as-is to match the established pattern in this file (e.g. `test_history_json_outputs_valid_json:194`). Worth a separate sweep across the test suite if the team wants to standardize, but out of scope for this isolated task.

PR #501, opened 2026-05-14
fix(audit): p2-t2-artifact-json review fallout (#496)

teng-lin:
## Summary
Follow-up to #496 (P2.T2). The original PR was merged before the bot-review fix commit landed; this PR carries the fixes forward.

Three findings folded into one commit (cherry-picked from the original branch):

1. **artifact delete --json no longer hangs / breaks JSON purity** (Claude medium / CodeRabbit major). `click.confirm()` would block on stdin even when `--json` was set, and the prompt itself wrote to stdout — clobbering JSON parseability. Fix: `--json` implies `--yes`, so the prompt is silently skipped. Interactive users who want a confirm omit `--json`. The dead JSON-cancellation payload is removed since the cancel branch is now unreachable in JSON mode.
2. **artifact poll --json uses direct attribute access** (Claude low). `getattr(status, "task_id", task_id)` etc. is replaced with direct `status.task_id` access on the typed `GenerationStatus` dataclass, matching how `artifact wait` consumes the same object.
3. **artifact get --json includes notebook_id** (Claude minor). Mirrors the wrapper used by `artifact list --json` so cached responses share one schema across the two commands. Both the normal and the not-found payload carry it.

Plus one test addition:
- `test_artifact_get_json_not_found` exercises the `art is None` JSON path that #496 introduced but didn't test (Claude nit).

## Push-backs (logged here for the record)
- **Claude low — `exported: false` + exit 0**: pre-existing non-JSON behavior; in scope for I14 (Phase 3), not P2.T2.
- **CodeRabbit nit — JSON-cancel regression test for `delete`**: fix #1 makes the cancel branch unreachable when `--json` is set, so there is nothing left to regress against in the targeted code path.

## Phase / plan
- Audit follow-up to Phase 2 / P2.T2 from `.sisyphus/plans/cli-ux-remediation/phase-2.md` (audit row I3).
- Strict file isolation: only `src/notebooklm/cli/artifact.py` and `tests/unit/cli/test_artifact.py`.

## Test plan
- [x] `uv run ruff format --check .`
- [x] `uv run ruff check .`
- [x] `uv run pytest --cov=src/notebooklm --cov-fail-under=90` — 3469 passed, coverage 92.61%, `cli/artifact.py` at 93%.
- [x] 32 tests in `tests/unit/cli/test_artifact.py`, including the new `test_artifact_get_json_not_found`.

## Summary by CodeRabbit

* **New Features**
  * `artifact get --json` response now includes `notebook_id` field.

* **Bug Fixes**
  * `artifact get --json` properly indicates when an artifact is not found.
  * `artifact delete --json` suppresses interactive prompts for automated workflows.
  * `artifact poll --json` provides more consistent status reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/501)

### DIFF
--- a/src/notebooklm/cli/artifact.py
+++ b/src/notebooklm/cli/artifact.py
@@ -153,9 +153,12 @@ def artifact_get(ctx, artifact_id, notebook_id, json_output, client_auth):
 
             if json_output:
                 if art is None:
-                    json_output_response({"id": resolved_id, "found": False})
+                    json_output_response(
+                        {"notebook_id": nb_id_resolved, "id": resolved_id, "found": False}
+                    )
                     return
                 data = {
+                    "notebook_id": nb_id_resolved,
                     "id": art.id,
                     "title": art.title,
                     "type": get_artifact_type_display(art).split(" ", 1)[-1],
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

PR #504, opened 2026-05-14
fix(cli)!: get commands exit 1 on not-found (C1)

teng-lin:
## Summary

- BREAKING: \`source get\` / \`artifact get\` / \`note get\` now exit 1 with a typed JSON error envelope when the requested ID does not resolve to an existing item (was exit 0 with a "not found" line on stdout). Matches the rest of the CLI's user-error convention so scripts can branch on the exit code without parsing output text.
- Covers BOTH not-found code paths: input IDs ≥20 chars (skip the partial-resolve list round-trip in \`_resolve_partial_id\`) AND the rare race where partial-resolve succeeds but the subsequent \`get\` returns \`None\` because the row was deleted between the two calls.
- Routes both branches through \`error_handler._output_error\` so \`--json\` emits the canonical \`{"error": true, "code": "NOT_FOUND", "message": "...", "id": "...", "notebook_id": "..."}\` envelope and text-mode messages land on stderr (not stdout). The pre-existing partial-resolve "no match" branch (already exit 1 via \`ClickException\`) is unchanged.

## Task

Phase 3, Task P3.T1 of the internal cli-ux-remediation plan ([phase plan](.sisyphus/plans/cli-ux-remediation/phase-3.md)). Closes audit row C1.

## Test plan

- [x] TDD red phase — added 9 new tests + flipped 3 existing exit-0 placeholder tests; all failed as expected before implementation.
- [x] TDD green phase — implementation passes all 12 (new + flipped) tests.
- [x] \`uv run ruff format .\` — clean.
- [x] \`uv run ruff check .\` — clean.
- [x] \`uv run mypy src/notebooklm --ignore-missing-imports\` — Success: no issues found in 58 source files.
- [x] \`uv run pytest --cov=src/notebooklm --cov-fail-under=90\` — 3532 passed, 11 skipped, coverage 92.78% (gate ≥90%).

## Migration notes for callers

Shell scripts that relied on exit-0-on-not-found (e.g. \`notebooklm source get X | grep -q '<title>' && do_something\`, or any \`set -e\` pipeline that used \`get\` as a probe) must switch to:

\`\`\`sh
if ! notebooklm source get "\$ID"; then
    handle_missing "\$ID"
fi
\`\`\`

JSON consumers that branched on the old \`{"found": false}\` placeholder must switch to branching on the exit code or on the \`code: "NOT_FOUND"\` field in the typed error envelope. See the updated \`docs/cli-exit-codes.md\` migration note for shell + Python recipes.

## Deferred polish findings

None — the spec is tightly scoped to C1 and the comments inline (in source.py / artifact.py / note.py) already explain the \`raise AssertionError("unreachable")\` mypy-narrowing pattern that avoids touching \`error_handler.py\` (which the spec explicitly forbids).

## Summary by CodeRabbit

* **Breaking Changes**
  * `source get`, `artifact get`, and `note get` now exit with code 1 (instead of 0) when items are not found
  * Non-JSON error messages now appear on stderr instead of stdout

* **Changes**
  * Not-found errors with `--json` now return a structured error envelope with `code: "NOT_FOUND"` and metadata

* **Documentation**
  * Updated CLI reference and migration guides to reflect new exit codes and error behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/504)


teng-lin:
@claude review

PR #506, opened 2026-05-14
feat(cli): warn when source add path-shaped arg doesn't exist (I8)

teng-lin:
## Summary

Audit row I8 from the CLI UX remediation plan.

`notebooklm source add ./missing.md` previously fell through silently to inline-text ingestion — the success line (`Added source: src_xxx`) was indistinguishable from a real file upload, so a typo became invisible to both humans and AI agents reading the CLI back-channel.

**Behavior change (additive, advisory):**
- When the input is path-shaped (contains `/` / `\` OR ends in a curated document extension — `.pdf`, `.md`, `.markdown`, `.txt`, `.html`, `.htm`, `.doc`, `.docx`, `.rtf`, `.odt`, `.csv`, `.tsv`, `.epub`) but the path does not exist on disk, the CLI now emits a stderr advisory warning before adding the source as text.
- Source is still added as text — inferred-text behavior is preserved, no breaking exit-code change.
- Explicit `--type text` suppresses the warning (user has stated intent).
- Help text on `source add` updated to surface the new behavior and to fix the contradicted `./doc.md` example (which previously claimed file content was uploaded "as text" — a lie that was the original I8 finding).

**Extension allowlist** is intentionally narrow: only formats that are unambiguously documents in the NotebookLM source domain. `.json` / `.yaml` / etc. are deliberately excluded to avoid false positives on legitimate inline content (e.g. pasting a config snippet as a note).

**Out of scope (deferred per plan):** `--strict-path` flag for promoting the warning to an exit-1 error.

**TDD trace:** new tests committed as part of the implementation pass; verified red→green in the worktree before commit.

## Test plan

- [x] New regression suite `tests/unit/cli/test_source.py::TestSourceAddPathShapedMissing` covers all four required cases (path-shaped + missing → warn, explicit `--type text` suppresses, pure text → no warn, existing path → no warn) plus an extra extension-only case (`missing.md` no slash).
- [x] Full `tests/unit/cli/test_source.py` (115 tests) green.
- [x] `tests/unit/cli/test_help_text.py` snapshot still green after help-text edit.
- [x] Pre-commit gate (`ruff format`, `ruff check`, `mypy`) clean.
- [x] Full suite `uv run pytest --cov=src/notebooklm --cov-fail-under=90`: 3553 passed, 11 skipped, total coverage **92.81%** (gate 90%).

## Summary by CodeRabbit

* **New Features**
  * `source add` now emits an advisory stderr warning when given a path-like argument that doesn't exist; the input is still ingested as text. Use `--type text` to suppress the warning.

* **Documentation**
  * Help text and release notes updated to document the advisory and suppression option.

* **Tests**
  * Added unit tests covering warning emission, suppression, and existing-path handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/506)


teng-lin:
@claude review

teng-lin:
Addressed in commit 0aaad08: tightened the two or assertions in TestSourceAddPathShapedMissing to the full-phrase match "looks like a path but does not exist" per @claude review nit.

The other findings were classified as "acceptable / consistent with existing style" (multi-word + extension false positive — suppressible via --type text; comment verbosity — matches the in-practice file style; warning: prefix — divergence noted, no action requested) so leaving those as-is. Gemini quota was exhausted today, so no second-bot pass on this PR.

PR #507, opened 2026-05-14
fix(cli)!: cinematic-video --format raises on non-cinematic value (I2)

teng-lin:
## Summary

Audit row **I2** (cli-ux-remediation phase 4). `notebooklm generate cinematic-video --format mp4` (or any other non-`cinematic` value) used to be a *silent override* — the value was thrown away and `cinematic` was substituted with no warning. This PR makes the conflict explicit: it now raises `click.UsageError` and exits `2`, mirroring the existing `--style-prompt cannot be used with cinematic video` rejection one block down (`generate.py:519`).

- **Detection**: `ctx.get_parameter_source("video_format") == ParameterSource.COMMANDLINE` distinguishes "user explicitly passed `--format`" from "user did not pass `--format`". The implicit default path stays a no-op coercion to `cinematic`; only the explicit conflict raises.
- **Help text**: the `cinematic-video` alias `--help` block now documents the lock explicitly so the constraint is discoverable without running into the error first.
- **Backward-compat**: the canonical happy path (`generate cinematic-video` with no flag, and `generate cinematic-video --format cinematic`) is unchanged. The only behavior change is the previously-silent override now becoming a hard error — which is the whole point of the fix.

### BREAKING note (in CHANGELOG)
Any script that was passing `--format <non-cinematic>` to `generate cinematic-video` was already getting cinematic output (it was being silently overridden) — drop the flag, or switch to `generate video --format <value>` if a non-cinematic format was actually intended.

## Test plan
- [x] **TDD red**: `test_generate_cinematic_video_rejects_non_cinematic_format` (parametrised over `explainer` and `brief`) failed against the unmodified code path (silent override let the test reach the network layer).
- [x] **TDD green**: After the fix, all 8 `TestGenerateCinematicVideo` cases pass — the 4 pre-existing tests (default format, with description, ignores `--style`, rejects `--style-prompt`) PLUS the 4 new ones (rejects `--format explainer`/`brief`, accepts explicit `--format cinematic`, accepts no-flag default, help text mentions `--format` + `cinematic`).
- [x] Pre-commit gate: `uv run ruff format . && uv run ruff check . && uv run mypy src/notebooklm --ignore-missing-imports && uv run pytest --cov=src/notebooklm --cov-fail-under=90` — **3552 passed, 11 skipped, coverage 92.81%**.
- [x] File scope respected: only `src/notebooklm/cli/generate.py`, `tests/unit/cli/test_generate.py`, and a single `[Unreleased] → Changed` bullet in `CHANGELOG.md`.

## Release Notes

* **Bug Fixes**
  * The `generate cinematic-video` command now enforces format restrictions: passing a non-cinematic `--format` value returns an error with exit code 2 instead of silently overriding to cinematic output. Omitting `--format` still defaults to cinematic behavior.

* **Documentation**
  * Updated `cinematic-video` help documentation to clarify that `--format` is locked to cinematic values and to direct users to `generate video` for other formats.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/507)


teng-lin:
@claude review

PR #509, opened 2026-05-14
docs(cli): clarify artifact poll vs wait ID kind (C2)

teng-lin:
## Decision: document (not unify)

Audit row C2 flagged that `artifact poll` and `artifact wait` have ambiguous `--help` about ID kind. After reading both commands and the underlying client layer:

- Per `_artifacts.py::_parse_generation_result` docstring (`src/notebooklm/_artifacts.py:2455-2457`): **"The API returns a single ID that serves as both the task_id (for polling during generation) and the artifact_id (once complete)."** So the IDs are *literally the same identifier*, not two distinct kinds.
- But the **commands** have a real operational split that's load-bearing:
  - **`poll`**: one-shot non-blocking check. Passes the ID through unchanged (does NOT call `resolve_artifact_id`). This works *immediately* after `generate <type>` returns, before the artifact appears in `artifact list`. `poll_status` even returns `status="not_found"` for that case so callers can disambiguate "genuinely pending" from "removed" (`_artifacts.py:1769-1771`).
  - **`wait`**: blocks with exponential backoff until terminal. Calls `resolve_artifact_id` (prefix-matches against `artifact list`).
- Unifying would lose either the poll-before-list-population capability OR the partial-ID ergonomics on `wait`. Both are real value, so **document** wins.

## Summary

- Rewrote both `--help` docstrings (`src/notebooklm/cli/artifact.py`) with explicit examples of where each ID comes from:
  - `poll`: from the `task_id` field of `notebooklm generate <type>` output (or after `Started:` in human-readable mode).
  - `wait`: from `notebooklm artifact list` (full UUID or unique prefix).
  - Both note: same underlying identifier, the split is operational not semantic.
- Added a **"Common confusion: `poll` vs `wait` ID kind"** callout in `docs/cli-reference.md` near the artifact command table; labelled the ID source inline in the table rows.
- Added `tests/unit/cli/test_artifact.py::TestPollWaitIdKindHelp` — five snapshot tests pinning the canonical disambiguation phrasing in both `--help` outputs (`task_id`, `generate`, `artifact list`, `same identifier`, plus a runnable example) so future docstring rewrites can't silently regress this.
- CHANGELOG entry under `[Unreleased] → Changed`.

## File scope (per plan)

Touched only the four allowed files:
- `src/notebooklm/cli/artifact.py`
- `tests/unit/cli/test_artifact.py`
- `docs/cli-reference.md`
- `CHANGELOG.md` (one bullet under `[Unreleased] → Changed`)

## Test plan

- [x] `uv run mypy src/notebooklm --ignore-missing-imports` — clean
- [x] `uv run pytest --cov=src/notebooklm --cov-fail-under=90` — 3553 passed, 11 skipped, coverage 92.80%
- [x] New snapshot suite `TestPollWaitIdKindHelp` (5 tests) all green
- [x] Manually verified `notebooklm artifact poll --help` and `notebooklm artifact wait --help` render cleanly with proper paragraph breaks, examples, and the cross-reference

## Summary by CodeRabbit

* **Documentation**
  * Clarified `artifact poll` vs `artifact wait` help and CLI docs, including when to use each and that they accept the same identifier form in different lifecycle contexts.
  * Documented `--json` output option for both commands and added a “Common confusion” guidance section with examples.

* **Tests**
  * Added help-text validation tests to lock in the canonical help wording and examples.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/509)


teng-lin:
@claude review

PR #520, opened 2026-05-15
test(concurrent): reusable assertion helpers (T7.A2)

teng-lin:
## Summary

Tier 7 foundation — adds `tests/integration/concurrency/helpers.py` with three plain-function helpers consumed by Phase 2 fix tests:

- **`assert_no_concurrent_overlap(events)`** — replays an enter/exit event stream and asserts peak depth ≤ 1. Used for serialization regression tests (per-conversation lock T7.F1, polling-dedupe leader T7.E2).
- **`assert_unique_outputs(make_coro, n_concurrent)`** — N concurrent invocations must produce N distinct results. Used for monotonic/unique-ID tests (`next_reqid` regression coverage).
- **`with_simulated_cancel(coro, delay, *, label=None)`** — schedules `task.cancel()` after `delay` and **returns** (not raises) the result or captured exception. Used for `asyncio.shield` regressions (T7.C1–C4): cancel one waiter mid-flight, assert sibling work survives.

Plain functions (not fixtures) so per-bug tests can `from .helpers import ...` from anywhere under `tests/integration/concurrency/`.

## Design notes

- `with_simulated_cancel` returns the exception rather than raising — Phase 2 tests typically want "leader survived; canceled follower received `CancelledError`" as two assertions, not one `pytest.raises` block.
- `assert_no_concurrent_overlap` raises with the offending event index + timestamp so a failing test points at the offending coroutine in its log capture.
- `assert_unique_outputs` requires hashable outputs; non-hashable cases stay inline (no over-design here).
- All three reject misuse (negative delays, `n_concurrent < 1`, unknown event kinds) with `ValueError`/`AssertionError` and clear messages.

## Test plan

- [x] `uv run pytest tests/integration/concurrency -v` — 18 tests pass in 0.21s
- [x] `uv run pytest -W error::RuntimeWarning` — clean (caught one un-awaited-coroutine on misuse path; fixed by `coro.close()` in the validation branch)
- [x] Pre-commit suite: `ruff format`, `ruff check`, `mypy src/notebooklm`, full `pytest` (3640 passed, 11 skipped, 0 new warnings)

## Out of scope

- Tests for specific Tier 7 bugs (those go in per-fix PRs in Phase 2; the helpers here are dependencies of those tests).
- Helpers for fixture composition / barrier orchestration (already in `conftest.py` from T7.A1).

Refs: T7.A2 of the Tier 7 thread-safety remediation arc; sequential after #514.

## Summary by CodeRabbit

* **Tests**
  * Added new concurrency test helpers to validate serialized/overlap behavior, enforce uniqueness of concurrent outputs, and simulate cancellations safely.
  * Added comprehensive integration tests covering helpers, edge cases (invalid inputs, duplicate results, leak prevention), and cancellation propagation/cleanup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/520)

### DIFF
--- a/tests/integration/concurrency/helpers.py
+++ b/tests/integration/concurrency/helpers.py
@@ -216,6 +216,13 @@ async def with_simulated_cancel(
     "the leader survived; the canceled follower received
     ``CancelledError``" — returning makes that two assertions, not
     one ``pytest.raises`` block per scenario.
+
+    Footgun: if ``T`` is itself a ``BaseException`` subclass (a
+    coroutine that legitimately RETURNS an exception object rather
+    than raising), the caller cannot distinguish "coro returned an
+    exception value" from "coro raised". Phase 2 tests that need
+    that distinction should write a custom assertion rather than
+    rely on ``isinstance(result, BaseException)``.
     """
     if delay < 0:
         # Close the coro so misuse doesn't trigger
@@ -240,6 +247,13 @@ async def with_simulated_cancel(
     try:
         return await task
     except BaseException as exc:  # noqa: BLE001 — we explicitly return any exception
+        # Outer-cancellation safety: if THIS helper's invoking task is
+        # cancelled (vs. the inner ``task`` being cancelled by our own
+        # canceller), propagate the cancel into ``task`` so it doesn't
+        # become a stray background task. The exception is still
+        # returned per the helper's contract.
+        if isinstance(exc, asyncio.CancelledError) and not task.done():
+            task.cancel()
         return exc
     finally:
         # Best-effort cleanup; the canceller has either fired or is
@@ -248,7 +262,7 @@ async def with_simulated_cancel(
             canceller.cancel()
             try:
                 await canceller
-            except (asyncio.CancelledError, BaseException):  # noqa: BLE001
+            except BaseException:  # noqa: BLE001 — cleanup must not propagate
                 pass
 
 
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
@claude — LGTM noted, all four findings addressed in 73f5862:

1. ✅ Simplified `except (asyncio.CancelledError, BaseException)` → `except BaseException`. Cleaner; same semantics.

2. ✅ Outer-cancellation propagation. The helper now does `task.cancel()` on the inner task when it observes a `CancelledError` on its own await — so the inner task can never become a stray background task if a test teardown cancels the helper itself. Added a regression test (`test_with_simulated_cancel_propagates_outer_cancellation_to_inner`) that explicitly creates the helper, cancels it, and asserts the inner coroutine never reached its `inner_finished = True` line.

3. ✅ `T | BaseException` ambiguity noted in the docstring's Notes section as a known footgun for coroutines that legitimately RETURN a `BaseException` value. Phase 2 authors writing such tests are pointed at custom assertions.

4. ✅ `test_unique_outputs_returns_outputs_for_caller_assertions` now uses `sorted()` + `set()` instead of positional equality. Test is now stable under uvloop or any other event-loop implementation that doesn't preserve creation order for zero-await coroutines.

Pre-commit re-run after the changes: ruff format, ruff check, mypy, and `pytest tests/integration/concurrency -v -W error::RuntimeWarning` (19 tests in 0.22s) all green.

Thanks for the careful review.

PR #554, opened 2026-05-15
feat(tests): re-derive chunked byte-counts post-scrub (I9, T8.D7)

teng-lin:
## Summary

- New helper `recompute_chunk_prefix` in `tests/cassette_patterns.py` walks Google's `<count>\n<payload>\n` chunked response bodies and rewrites the byte-count headers when sanitization changes payload length.
- Wired into `tests/vcr_config.py::scrub_response` for both `str` and `bytes` code paths, so post-scrub cassettes stay self-consistent on the byte-count axis.
- Empirically: across the 75 cassettes currently on `main`, the helper corrects 343 chunk byte-count mismatches and leaves zero remaining.

## Task

`.sisyphus/plans/tier-8-rpc-vcr-remediation/phase-1.md#T8.D7` — promoted from phase 4 to phase 1 wave 4 because T8.B6's bulk re-scrub script needs this utility before running.

## Coordination notes for the in-flight T8.A1 / A2 / A4

- **T8.A4** (canonical sanitization registry): this PR creates `tests/cassette_patterns.py` as a thin home for the chunk-prefix helper only. When T8.A4 merges it will populate the same file with the `SENSITIVE_PATTERNS` registry — the two patches edit disjoint sections and should rebase cleanly.
- **T8.A1** (default `rpcids` matcher) and **T8.A2** (`f.req` body matcher): both modify the `match_on` list / matcher functions in `vcr_config.py`. This PR only adds a new helper-loader at the top of the file and a few lines in `scrub_response`, so all three patches edit disjoint sections.

## Invariants preserved

- Decoder tolerance behavior (`src/notebooklm/rpc/decoder.py:217-237`) is intentionally **unchanged**. The decoder continues to warn-and-parse on mismatched byte counts; this helper only ensures cassettes don't trigger that warning during replay.
- Production scrub behavior is unchanged beyond the new byte-count re-derivation step.
- Helper is conservative: bodies without a `<digits>\n<non-digits>` pair (HTML, plain JSON, CRLF-only bodies) pass through unchanged.
- Helper is idempotent — re-running on an already-correct body yields identical output.

## Test plan

- [x] 12 new unit + integration tests in `tests/unit/test_vcr_config.py` (empty body, XSSI prefix preservation, multi-chunk rewrite, UTF-8 byte counting for emoji, idempotence, dangling-header safety, consecutive-digit-line safety, digits-inside-payload safety, `scrub_response` integration for both `str` and `bytes`)
- [x] Existing `tests/unit/test_cookie_redaction.py` (which imports `vcr_config.py` via `importlib`) still passes after the new sibling-load helper.
- [x] `uv run ruff format . && uv run ruff check . && uv run mypy src/notebooklm --ignore-missing-imports && uv run pytest` — 3700 passed, 11 skipped, 1 warning.

## Deferred polish findings

- **OPTIONAL (gemini)** Large cassette performance: list-split + join on multi-MB bodies is fine for the test-time / one-shot rescrub use case; reconsider only if T8.B6 hits noticeable latency.
- **OPTIONAL (ultrathink)** Shape-lint character vs UTF-8 byte mismatch on non-ASCII payloads. Currently zero affected cassettes on `main`; would require a separate change in `tests/unit/test_cassette_shapes.py` (T8.A3 territory) to align on byte count.

## Summary by CodeRabbit

* **Tests**
  * Enhanced test infrastructure for handling and validating cassette data, improving test reliability and consistency across different response formats.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/554)

### DIFF
--- a/tests/cassette_patterns.py
+++ b/tests/cassette_patterns.py
@@ -35,8 +35,9 @@ _XSSI_PREFIX = ")]}'\n\n"
 # A "chunk header" line is a line consisting of ONLY ASCII digits — that's the
 # advertised byte count for the next payload line. Restricting to ASCII digits
 # avoids accidentally treating a JSON payload line that happens to start with a
-# digit-like character as a header.
-_CHUNK_HEADER_RE = re.compile(r"\A\d+\Z")
+# digit-like character as a header. ``fullmatch`` anchors at both ends so we
+# don't need explicit ``\A`` / ``\Z`` (claude-bot review on PR #554).
+_CHUNK_HEADER_RE = re.compile(r"\d+")
 
 
 def recompute_chunk_prefix(body: str) -> str:
@@ -122,8 +123,8 @@ def recompute_chunk_prefix(body: str) -> str:
         #  - JSON payloads that happen to be a single integer literal
         #    immediately preceded by another digit-only line (unlikely in
         #    practice but we'd rather be conservative)
-        is_header = _CHUNK_HEADER_RE.match(line) is not None
-        has_payload = i + 1 < len(lines) and not _CHUNK_HEADER_RE.match(lines[i + 1])
+        is_header = _CHUNK_HEADER_RE.fullmatch(line) is not None
+        has_payload = i + 1 < len(lines) and not _CHUNK_HEADER_RE.fullmatch(lines[i + 1])
         if is_header and has_payload:
             payload = lines[i + 1]
             new_count = len(payload.encode("utf-8"))
--- a/tests/cassette_patterns.py
+++ b/tests/cassette_patterns.py
@@ -1,0 +1,137 @@
+"""Cassette sanitization helpers shared between VCR config and the bulk re-scrub script.
+
+This module is the canonical home for cassette-mutating utilities. T8.A4 will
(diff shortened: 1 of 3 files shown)


teng-lin:
@claude review

teng-lin:
Addressed the three claude-bot review nits in commit 7f57b58:

1. `_CHUNK_HEADER_RE`: switched to `re.fullmatch(r"\d+", ...)` — drops the redundant `\A`/`\Z` anchors that paired with `re.match()`.
2. `scrub_response` str-path: introduced the same `rederived` intermediate variable the bytes-path uses, so both branches read identically.
3. Test comment in `test_recompute_chunk_prefix_corrects_synthetic_shrinkage`: rewrote the misleading "16 bytes" line to clarify the post-scrub payload is 22 bytes (JSON brackets count, not just the inner 16-char placeholder). Assertion was already correct.

The CRLF observation is documented as intentional pass-through in the helper's docstring (no current cassette uses CRLF endings); no code change there.

Re-ran the full pre-commit chain — all 3078 unit tests still green.

PR #562, opened 2026-05-15
fix(core): shield shared refresh task from waiter cancellation (T7.C1)

teng-lin:
## Summary

- Wrap `await refresh_task` in `_core.py:_await_refresh` with `asyncio.shield(...)` so a caller cancelled via `asyncio.wait_for(timeout=...)` (or any outer cancellation) unwinds locally without taking down the shared single-flight refresh task.
- `self._refresh_task` is left intact across the cancellation and replaced only on the next refresh wave once the current task transitions to `done()` — pinned by a new regression test.
- 1-line behavioral change in `_core.py`, ~10 lines of invariant docstring, and 2 new red-green regression tests in `tests/integration/concurrency/`. Dedupe semantics and lock structure unchanged.

## Task

- Spec: `.sisyphus/plans/tier-7-thread-safety-concurrency/phase-2.md#T7.C1`
- Audit reference: §4 (cancellation propagation cluster)
- Wave 2 of the Tier 7 thread-safety / concurrency remediation arc

## The bug

Before this PR, the join in `_await_refresh` was an unshielded `await self._refresh_task`. If any caller was cancelled while joined to the shared task — for example, a `wait_for(timeout=...)` firing while the refresh callback was still in flight — the `CancelledError` propagated into the *underlying task itself*, cancelling the single in-flight refresh and surfacing `CancelledError` to every sibling waiter joined to it.

## The fix

```python
await asyncio.shield(refresh_task)
```

The `asyncio.shield` wrapper future is what gets cancelled; the underlying task continues to produce a value for non-cancelled siblings. The shield is placed *outside* the `_refresh_lock` block (matching the pre-existing comment) so a cancelled waiter never tears down the lock.

## Tests

`tests/integration/concurrency/test_refresh_cancellation_propagation.py` — 2 regression tests, both confirmed red on `origin/main` and green with the shield applied:

1. **`test_waiter_cancellation_does_not_kill_shared_refresh`** — behavioral invariant: two concurrent waiters, one cancelled via `wait_for(timeout=0.01)`, asserts the surviving waiter still receives the successful refresh result and the callback fired exactly once (single-flight preserved).
2. **`test_refresh_task_slot_not_cleared_on_waiter_cancellation`** — structural invariant: pins that `self._refresh_task` retains the same task object identity across a waiter cancellation and the task is still alive (`not done()`) when the cancellation lands.

All 3 existing `test_refresh_state_machine.py` tests (single-flight, failure-propagation, second-wave-distinct) still pass — dedupe semantics preserved.

## Test plan

- [x] `uv run mypy src/notebooklm --ignore-missing-imports` — clean
- [x] `uv run pytest tests/integration/concurrency/ tests/unit/test_refresh_state_machine.py` — 43 passed
- [x] Full suite: `uv run pytest` — 3798 passed, 12 skipped, 102 xfailed, 0 failed

## Deferred polish findings

None — the change is the minimal surgical shield. The harness duplication of `make_core` as a local `_opened_core` helper is intentional (per the file docstring) because `tests/unit/conftest` is not importable from `tests/integration/concurrency/`; consolidating shared test infra is out of scope and risks colliding with other Tier 7 waves.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where cancelling a refresh operation could inadvertently interrupt other concurrent refresh requests, ensuring independent operation of concurrent refresh calls.

* **Tests**
  * Added integration tests to verify proper handling of concurrent refresh cancellation scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/562)

### DIFF
--- a/src/notebooklm/_core.py
+++ b/src/notebooklm/_core.py
@@ -1004,6 +1004,16 @@ class ClientCore:
         401s on the same client triggers exactly one token refresh. The lock
         protects task-creation only; the await on the task itself happens
         outside the lock so other callers can join.
+
+        The join is wrapped in :func:`asyncio.shield` (T7.C1, audit §4) so
+        that a caller cancelled while waiting — e.g. via
+        ``asyncio.wait_for(..., timeout=...)`` — unwinds locally without
+        propagating the ``CancelledError`` into the *shared* refresh task.
+        Without the shield, one cancelled waiter would cancel the
+        underlying task, taking down every sibling joined to the same
+        single-flight refresh. The slot at ``self._refresh_task`` is left
+        intact across the cancellation and is replaced only on the next
+        refresh wave once the current task transitions to ``done()``.
         """
         assert self._refresh_callback is not None
         assert self._refresh_lock is not None
@@ -1017,7 +1027,7 @@ class ClientCore:
                 self._refresh_task = asyncio.create_task(coro)
                 refresh_task = self._refresh_task
 
-        await refresh_task
+        await asyncio.shield(refresh_task)
 
     async def query_post(
         self,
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

PR #566, opened 2026-05-15
test(core): regression for close-cancellation transport leak (T7.C2; subsumed by B4)

teng-lin:
## Summary

T7.C2's spec asked us to audit whether the `asyncio.shield` around `self._http_client.aclose()` in `ClientCore.close` survives a cancel that lands while `aclose` is in flight via the `__aexit__` surface.

**Audit conclusion**: PR #526 (T7.B4, sha `d8b5bd6`) already shipped the shield at the correct location. The source change for C2 is a deliberate no-op — the value of this PR is the **complementary regression test** exercising a different cancel-injection site than B4's `test_cancel_mid_close_does_not_leak_transport`.

## What the new test exercises

- Opens client with `keepalive=...` so a background poke task is alive and close has to drive a real keepalive teardown.
- Monkeypatches `_rotate_cookies` to hang on an unset `asyncio.Event` so the keepalive task is genuinely parked when close starts cancelling it.
- Slows `aclose` with `await asyncio.sleep(0.2)` so the outer `wait_for(timeout=0.1)` reliably cancels the **shielded** await rather than letting close finish in microseconds.
- Drives `__aexit__` through `asyncio.wait_for(timeout=0.1)` and asserts `http_client_ref.is_closed` afterwards (the saved reference is needed because successful close nulls `_core._http_client`).

## Verification

- The test was verified to fail loudly when the B4 shield is removed (cancelled-mid-`aclose` path leaves `is_closed=False`), proving the shield is doing real work and the test pins the invariant.
- ✅ `uv run ruff format . && uv run ruff check . && uv run mypy src/notebooklm --ignore-missing-imports`
- ✅ `uv run pytest tests/integration/concurrency/test_close_cancellation_leak.py` → 1 passed

## References

- `src/notebooklm/_core.py:586-601` (the shielded close path)
- PR #526 / `d8b5bd6` (T7.B4 — actual shield deployment)
- Audit §7

## Test plan

- [x] New regression test passes locally
- [x] Regression confirmed: test fails if B4 shield is reverted
- [x] No source changes (test-only)

## Release Notes

* **Tests**
  * Added new integration regression test for concurrent client closure scenarios with keepalive enabled.
  * Added test fixture for keepalive authentication configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/566)

### DIFF
--- a/tests/integration/concurrency/test_close_cancellation_leak.py
+++ b/tests/integration/concurrency/test_close_cancellation_leak.py
@@ -156,7 +156,7 @@ async def test_close_during_keepalive_cancel_does_not_leak_transport(
         http_client_ref = client._core._http_client
         assert http_client_ref is not None, "open() must have installed a transport"
 
-        # Slow down ``aclose()`` so the outer ``wait_for(timeout=0.5)``
+        # Slow down ``aclose()`` so the outer ``wait_for(timeout=0.1)``
         # below reliably injects a ``CancelledError`` while the shielded
         # close is in flight. Without this the entire close path
         # finishes in microseconds (mock transport, no real connections
--- a/tests/integration/concurrency/test_close_cancellation_leak.py
+++ b/tests/integration/concurrency/test_close_cancellation_leak.py
@@ -120,11 +120,11 @@ async def test_close_during_keepalive_cancel_does_not_leak_transport(
         With keepalive's poke stuck here, ``_core.close()`` has to
         cancel the keepalive task and ``gather()`` it before reaching
         ``save_cookies`` and the shielded ``aclose``. The outer
-        ``wait_for(timeout=0.5)`` will inject a cancel somewhere along
-        that path — at the gather, the save_cookies ``to_thread``, or
-        the shielded aclose itself. Wherever it lands, the shield in
-        the outer ``finally`` must still drive ``aclose()`` to
-        completion; that's what the assertion below proves.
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude review

teng-lin:
@claude review

teng-lin:
Addressed in 86210f3: updated `_hanging_rotate`'s docstring to reference `wait_for(timeout=0.1)` and clarified that the cancel is injected inside the shielded `aclose` via the patched `_slow_aclose` (not at the gather / save_cookies await points the older draft mentioned). No behavior change.

Thanks for the careful review @claude — the timing-chain table was especially helpful.

teng-lin:
Addressed in 2375bf7: also fixed the second stale `0.5` reference in the inline comment at the top of the `_slow_aclose` block. (The first fix in 86210f3 caught the docstring; this one catches the comment block right before `original_aclose = http_client_ref.aclose`.) All four prose mentions of `wait_for(timeout=...)` now consistently say `0.1`.

Re: the timing-margin observation — agreed that gather/save_cookies on this fixture should be near-instant (no storage_path → save_cookies is a no-op short-circuit, and gather on a cancelled task with cleanly-propagating CancelledError is one event-loop tick). The `aclose_started` invariant is the right guard if a future CI quirk slows that down.

PR #569, opened 2026-05-15
fix(tests): scrub Drive tokens in source cassettes (I17, T8.B5)

teng-lin:
## Summary

Re-scrubs the two Drive-flow cassettes called out by phase-1 audit
finding I17 so they pass the canonical leak guard and can come off the
phase-2 repair allowlist:

- ``tests/cassettes/sources_add_drive.yaml``
- ``tests/cassettes/sources_check_freshness_drive.yaml``

The leaks that disappear:

- **Drive AONS tokens** (``AONS...{20,}+``) — both files carried a
  literal AONS permission/ACL token in the WRB payload of the
  ``izAoDd`` response. Now ``SCRUBBED_AONS``.
- **``/ogw/`` avatar URLs** (``lh3.googleusercontent.com/ogw/<tok>``) —
  seven instances in ``sources_add_drive.yaml`` collapsed to
  ``SCRUBBED_AVATAR_URL`` via the T8.A6a scrubber.

Mechanically: load each cassette as YAML, apply
``tests.cassette_patterns.scrub_string`` to every request URI,
request/response body string, and header value, then run the scrubbed
response body through ``recompute_chunk_prefix`` so the chunked
``)]}'``-framed byte-count headers track the now-shorter payload (the
AONS token shrinks the chunk by ~92 bytes). Re-serialize with
``yaml.safe_dump`` — VCR's serializer is the same, so replay matchers
keep matching.

Two follow-on cleanups required to land this:

- ``tests/cassette_patterns.py`` + ``tests/unit/test_cassette_shapes.py``:
  three Google Sans language variants (Japanese, Korean, Simplified
  Chinese) and the ``Microsoft Edge`` browser-brand string added to
  ``DISPLAY_NAME_FALSE_POSITIVES``. The escaped-display-name detector
  fires on these CSS / Sec-CH-UA values otherwise — same shape A6a
  already handles for the four pre-existing Google Sans entries. No new
  scrubber regex; allowlist-only extension.
- ``tests/scripts/cassette_repair_allowlist.txt``: both cassettes
  dropped; the spec-explicit-entries comment count adjusted from 10 → 8.
- ``tests/unit/test_cassette_shapes.py:AUDIT_REPAIR_LIST``: both
  ``Phase 2 T8.B5 will fix`` entries removed.
- ``tests/unit/test_cassette_sanitizer.py:test_python_guard_repo_allowlist_is_explicit_basename_list``:
  required-basename tuple updated so the assertion tracks the new state.

## Test plan

- [x] ``uv run python tests/scripts/check_cassettes_clean.py --strict tests/cassettes/sources_add_drive.yaml tests/cassettes/sources_check_freshness_drive.yaml`` returns 0 leaks
- [x] ``uv run python tests/scripts/check_cassettes_clean.py`` (default allowlist): 4 scanned, 71 allow-listed, 0 leaks
- [x] ``uv run pytest tests/unit/test_cassette_shapes.py``: 10 pass / 70 xfail (the two T8.B5 entries no longer xfail)
- [x] ``uv run pytest tests/integration -k 'drive or freshness'``: 15 pass — replay against the scrubbed cassettes still matches
- [x] ``uv run pytest -m vcr``: 81 pass / 4 skip / 30 xfail
- [x] Full ``uv run pytest``: 3838 pass / 12 skip / 100 xfail
- [x] ``uv run ruff format . && uv run ruff check . && uv run mypy src/notebooklm --ignore-missing-imports`` all green


teng-lin:
@claude review

PR #574, opened 2026-05-15
fix(auth): offload storage saves to thread to unblock event loop (T7.D1)

teng-lin:
## Summary
- Wrap the two `save_cookies_to_storage` call sites at `src/notebooklm/auth.py:507` (`AuthTokens.from_storage`) and `:3204` (`fetch_tokens_with_domains`) with `await asyncio.to_thread(...)` so the synchronous atomic-replace + fsync + flock no longer freezes the event loop on slow filesystems.
- `save_cookies_to_storage` itself is correctly synchronous and unchanged — only the callers were wrong.
- Stdlib only; `asyncio` was already imported.

## Task
- Plan: `.sisyphus/plans/tier-7-thread-safety-concurrency/phase-2-wave-3.md#T7.D1`
- Audit references: §6, §27
- Wave-3 task T7.D1 (D-series blocking-I/O offload)

## Test plan
- [x] New red test `tests/integration/concurrency/test_auth_load_blocks_loop.py` — monkeypatches `save_cookies_to_storage` to `time.sleep(0.5)`; a concurrently scheduled 50 ms heartbeat must tick ≥ 5 times during the save window. Pre-fix: 0 ticks (loop frozen). Post-fix: ~10 ticks. Both call sites exercised so a regression on either one trips the suite.
- [x] `uv run ruff format .`
- [x] `uv run ruff check .`
- [x] `uv run pytest tests/integration/concurrency/ tests/unit/test_auth.py` — **339 passed**, including the 2 new regression cases.

## Deferred polish findings
None. The change is a 10-line semantic delta plus a focused regression test; no refactor surface to clean up (the spec explicitly forbids touching `save_cookies_to_storage` itself).

## Wave-2 invariant carried forward
T7.D1 introduces no new `asyncio.create_task` for fire-and-forget cleanup — only `asyncio.to_thread` await sites — so the C4 strong-reference task-GC pattern (module-level `set` + `add_done_callback(set.discard)`) does not apply here. Flagged preemptively per wave-3 dispatch notes; no action needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication token refresh operations no longer cause responsiveness delays during credential persistence.

* **Tests**
  * Added integration tests to verify token operations maintain system responsiveness under concurrent workloads.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/574)

### DIFF
--- a/src/notebooklm/auth.py
+++ b/src/notebooklm/auth.py
@@ -503,8 +503,13 @@ class AuthTokens:
         # fails, carry the old baseline into the returned AuthTokens so a
         # later ClientCore can retry the delta instead of treating the mutated
         # jar as clean state.
+        # ``save_cookies_to_storage`` performs atomic-replace + fsync + flock
+        # under a synchronous file lock; offload to a worker thread so a
+        # slow filesystem (network FS, encrypted home, fcntl contention)
+        # can't freeze the event loop. T7.D1 / audit §6.
         post_save_snapshot = snapshot_cookie_jar(jar)
-        save_result = save_cookies_to_storage(
+        save_result = await asyncio.to_thread(
+            save_cookies_to_storage,
             jar,
             path,
             original_snapshot=snapshot,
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

PR #575, opened 2026-05-15
fix(sources): cancel sibling pollers on first wait_for_sources failure (T7.E1)

teng-lin:
## Summary
- `SourcesAPI.wait_for_sources` now drives its fan-out as explicit `asyncio.create_task` tasks, and on any exception cancels every pending task then drains them via `asyncio.gather(*tasks, return_exceptions=True)` before re-raising — so a slow sibling poll cannot keep ticking after the function returns to its caller (audit §5).
- Public signature unchanged: `async def wait_for_sources(notebook_id, source_ids, timeout=120.0, **kwargs) -> list[Source]`.
- Red test `tests/integration/concurrency/test_wait_for_sources_leak.py` asserts the slow task's `CancelledError` handler has fired by the time `wait_for_sources` returns, and total elapsed is well under the 1 s spec ceiling.

## Task
Plan: `.sisyphus/plans/tier-7-thread-safety-concurrency/phase-2-wave-3.md#T7.E1`
Wave 3 contention: `_sources.py` is shared with D2 (line 1353–1356) and D3 (lines 643, 1327–1356); this PR owns the disjoint surface around line 451 (post-format) — trivial rebase expected with either.

## Test plan
- [x] `uv run ruff format .` (clean)
- [x] `uv run ruff check .` (clean)
- [x] `uv run mypy src/notebooklm --ignore-missing-imports` (clean)
- [x] `uv run pytest tests/integration/concurrency/ tests/integration/test_sources.py` — 149 passed
- [x] Red test fails on `origin/main` baseline, passes after fix

## Deferred polish findings
none — surgical change scoped to the one call site at `_sources.py:451`. The pre-existing `cli_vcr/test_downloads.py` cassette mismatches on main are unrelated (T8 cassette work).

### DIFF
--- a/src/notebooklm/_sources.py
+++ b/src/notebooklm/_sources.py
@@ -445,10 +445,35 @@ class SourcesAPI:
                 nb_id, [s.id for s in sources]
             )
         """
+        # T7.E1: a bare ``asyncio.gather(*coros)`` propagates the first
+        # exception but does NOT await the sibling tasks it cancels. The
+        # cancelled siblings are left in a "cancellation requested but not
+        # yet observed" state — their ``finally`` blocks may not have run by
+        # the time we re-raise. That race lets a slow poll keep ticking
+        # against the network after ``wait_for_sources`` already raised to
+        # its caller (audit §5).
+        #
+        # Fix: drive the fan-out as explicit tasks. On any exception, cancel
+        # every pending task and drain them via ``return_exceptions=True``
+        # before re-raising the first failure. The drain guarantees each
+        # sibling has reached its ``except CancelledError`` block before we
+        # return control to the caller.
         tasks = [
-            self.wait_until_ready(notebook_id, sid, timeout=timeout, **kwargs) for sid in source_ids
+            asyncio.create_task(self.wait_until_ready(notebook_id, sid, timeout=timeout, **kwargs))
+            for sid in source_ids
         ]
-        return list(await asyncio.gather(*tasks))
+        try:
+            return list(await asyncio.gather(*tasks))
+        except BaseException:
+            for task in tasks:
+                if not task.done():
+                    task.cancel()
+            # Drain cancelled (and any already-failed) siblings before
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the reviews. Consolidated reply since there are no inline review threads:

**gemini-code-assist** — "no review comments". No action.

**coderabbitai** — hit hourly rate limit / out-of-credits before running. Acceptable per project's quota-terminal allowance for review bots (same policy applied to gemini in prior waves).

**claude[bot]** — LGTM with three observations, all framed as non-blocking:

1. *"`asyncio.gather` does not cancel siblings on non-`CancelledError` exception — it orphans them."* — Correct correction to my PR description's framing. The user-visible bug is the same (siblings keep running after `wait_for_sources` raises), but the precise mechanism is task-orphaning, not unawaited-cancellation. The fix is the same regardless. Noted; not changing the in-code comment because it now describes the post-fix invariant ("the drain guarantees each sibling has reached its `except CancelledError` block") which is accurate either way.

2. *"External-cancellation-during-drain replaces the original source error with `CancelledError`."* — Acknowledged as expected asyncio behavior, not a bug. Hardening the drain (e.g. `asyncio.shield`) would expand surgical scope beyond T7.E1's spec and risks masking legitimate caller-driven cancellation. Deferring.

3. *"Module docstring duplicates the function docstring."* — Cosmetic; keeping both because the module docstring documents the regression scenario (audit §5) for grep-from-failure-log discoverability, while the function docstring inside the test documents the assertion contract. Not changing.

Windows CI: 3/5 windows-matrix slots failed on pre-existing cassette/encoding issues (`test_check_freshness_drive` cassette mismatch, `test_cassette_shape` charmap decode) — same failures on the most recent merged PR #572 and actively being fixed by open PR #577 (`fix(tests): read cassette files with explicit UTF-8 encoding (Windows CI)`). Not caused by this PR.

Ready to merge once remaining test matrix slots finish.

PR #576, opened 2026-05-15
fix(tests): re-encode T8.B5 Drive cassettes as ASCII (Windows cp1252 fix)

teng-lin:
## Summary

Follow-up to #569 (T8.B5). The merged PR re-serialized the two Drive
cassettes with ``yaml.safe_dump(allow_unicode=True)`` so the 🔍 emoji
in ``sources_check_freshness_drive.yaml`` round-tripped as raw UTF-8
bytes (0xF0 0x9F 0x94 0x8D) instead of the original ASCII escape
``\U0001F50D``. That broke Windows CI on all five Python versions:

1. **VCR replay failure** (``test_check_freshness_drive``). vcrpy's
   ``FilesystemPersister.load_cassette`` opens cassettes via
   ``cassette_path.open()`` — text mode using the platform default
   encoding, which is cp1252 on Windows and cannot decode the UTF-8
   emoji. vcrpy silently catches the ``UnicodeDecodeError`` and treats
   the cassette as empty, so replay raises
   ``CannotOverwriteExistingCassetteException`` ("No match for the
   request") on the very first ``sources.list`` call.

2. **Shape-lint failure** (``test_cassette_shape[sources_check_freshness_drive.yaml]``).
   ``tests/unit/test_cassette_shapes.py`` reads cassettes via
   ``Path.read_text()`` — also cp1252-defaulted on Windows. Same byte,
   same crash, but louder (``UnicodeDecodeError`` bubbles up before the
   xfail check can run).

The fix:

- **Cassettes re-emitted from the origin/main pre-T8.B5 originals**
  through the same scrub_string + recompute_chunk_prefix pipeline, but
  with ``allow_unicode=False`` so non-ASCII codepoints are written as
  ``\uXXXX`` escapes. Both files are now byte-for-byte ASCII.
- **Belt-and-braces** ``encoding="utf-8"`` on the two
  ``Path.read_text()`` call sites in
  ``tests/unit/test_cassette_shapes.py`` (``_has_ogw_avatar`` and
  ``_load_cassette``) so the shape-lint stays Windows-safe even if a
  future cassette slips an unescaped non-ASCII byte through.

Cassettes are logically identical to the merged versions — same
interactions, same headers, same body content, same chunk byte counts.
The only change is the escape form (and an unchanged ``allow_unicode``
toggle in serialization).

## Test plan

- [x] Cassettes are pure ASCII (``non_ascii=0`` for both files)
- [x] ``uv run pytest tests/integration -k 'drive or freshness'``: 15 pass
- [x] ``uv run pytest tests/unit/test_cassette_shapes.py``: 14 pass / 64 xfail (the two Drive cassettes pass shape lint)
- [x] ``uv run ruff format . && uv run ruff check . && uv run mypy src/notebooklm --ignore-missing-imports`` all green
- [ ] CI Windows matrix returns to green (verified post-merge)


teng-lin:
@claude review

Issue #571, opened 2026-05-15
auth inspect / login --all-accounts misses accounts in non-Default Chrome user-profiles


teng-lin:
## Description

`notebooklm auth inspect --browser chrome` and `notebooklm login --browser-cookies chrome --all-accounts` only enumerate Google accounts visible to Chrome's `Default` user-data profile. Accounts signed in via other Chrome user-profiles (`Profile 1`, `Profile 2`, …) are silently dropped.

Both commands set user expectations of "show me everything":
- `auth inspect` advertises itself as *"List Google accounts visible to a browser's cookie store."*
- `--all-accounts` advertises itself as *"Extract every signed-in Google account into its own profile."*

In practice, with multiple Chrome user-profiles, both stop after the Default profile.

Related: #511 proposes `chrome::<profile-name>` syntax for explicit single-profile extraction. This issue is about the default / fan-out behavior — `auth inspect` and `--all-accounts` should auto-discover every Chrome user-profile without requiring the suffix.

## Steps to Reproduce

1. In Chrome, sign in to different Google accounts in different user-profiles:
   - `Default` → `alice@gmail.com`
   - `Profile 1` → `bob@gmail.com`
   - `Profile 2` → `carol@gmail.com`
2. Run:
   ```bash
   notebooklm auth inspect --browser chrome
   ```

## Expected Behavior

All three accounts (`alice`, `bob`, `carol`) listed. **Email is the primary identifier** — it is what `--account <email>` already selects on, what `email_to_profile_name` already uses for notebooklm profile names, and what users naturally think in. The Chrome user-profile a cookie came from is an implementation detail, kept out of the default view:

```
Browser: chrome
Found 3 signed-in Google account(s):
┏━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┓
┃ email             ┃ default ┃
┡━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━┩
│ alice@gmail.com   │    ✓    │
│ bob@gmail.com     │         │
│ carol@gmail.com   │         │
└───────────────────┴─────────┘
```

The Chrome user-profile name appears only in:
- `auth inspect -v` / `--verbose` (adds a `chrome user` column)
- `auth inspect --json` (adds a `browser_profile` field per account, for tooling/debugging)

And `notebooklm login --browser-cookies chrome --all-accounts` should write one notebooklm profile **per unique email**, regardless of which Chrome user-profile each account lives in.

[...]

PR #596, opened 2026-05-15
docs(skill): document auth refresh + auth inspect (#571 follow-up)

teng-lin:
## Summary

Documentation-only follow-up to #580 / #571. SKILL.md surfaced \`notebooklm login\` and \`auth refresh --quiet\` (cron keepalive) but not the in-place refresh paths that agents need for recovery, nor \`auth inspect\` for read-only account survey.

This PR adds:

- **Setup verification step 4** covering both \`auth refresh\` variants (server-side SIDTS refresh vs. \`--browser-cookies <browser>\` re-extract from a re-signed-in browser), the cron cadence (15–20 min), and the issue #571 fan-out behavior for multi-Chrome-profile setups.
- **\`auth inspect --browser <browser>\`** mention in step 3 so agents can survey accounts before picking one with \`--account EMAIL\`.
- **Quick-reference rows** for the new commands plus the existing \`--all-accounts\` and \`--browser-cookies\` flows that weren't in the table.
- **Autonomy-rules entries** for the read-only / non-destructive auth commands (\`inspect\`, \`refresh\`, \`refresh --browser-cookies\`) so agents can run them without prompting.

No code changes. Closes the user request raised during #571 review: "we need to document how to refresh the cookie using notebooklm auth refresh --browser-cookie browser".

## Test plan

- [x] \`uv run pytest tests/unit/test_claude_md_freshness.py\` passes (the doc-freshness regression tests).
- [x] Diff is +13 / -0 lines, no removed content.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -87,7 +87,12 @@ Before starting workflows, verify auth is in place. **Use `--test --json` (not b
 2. `notebooklm list --json` → expect valid JSON (may be empty for new accounts).
 3. **If auth fails or is missing → run `notebooklm login` first.** This is the primary auth path: opens a browser, the user signs in to Google once, and the resulting `storage_state.json` is reused on every subsequent run. Works on any environment with a display.
    - For headless contexts where opening a browser is not feasible, use `notebooklm login --browser-cookies <browser>` instead — extracts the user's already-logged-in cookies from Chrome/Firefox/etc. (requires the `[cookies]` extra; rookiepy may not install on Python 3.13+).
+   - To survey signed-in Google accounts before picking one: `notebooklm auth inspect --browser <browser>` (read-only; pass `-v` to see which Chromium user-profile each account came from, or `--json` for tooling).
    - Re-run step 1 after login to confirm.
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude review

teng-lin:
@claude addressed in 1fa7f54:

- **Factual inaccuracy ('running browser')** — fixed in both Step 4 and the Quick Reference row. Step 4 now says 'browser's on-disk cookie store' and explicitly notes 'The browser does NOT need to be running — rookiepy reads the cookies SQLite DB directly.' Quick Reference row label updated to match.
- **Autonomy-rule clarity** — fixed: both `auth refresh` entries now symmetrically describe what gets written (rotated cookies vs re-extracted cookies, both back to the same `storage_state.json` with the same `--profile`). No more apparent contradiction between 'no destructive writes' and 'rebuilds storage_state.json'.
- **`--browser-cookies` vs `--browser` flag asymmetry** — acknowledged as pre-existing; not addressing in this PR per your note.

PR #602, opened 2026-05-15
refactor(decoder): use safe_index in WRB traversal (I1, T8.D1)

teng-lin:
## Summary
- Migrate three decoder hot paths — `collect_rpc_ids`, `_find_wrb_status`, `extract_rpc_result` — from raw indexed access to `safe_index()`, unifying drift handling under the soft-strict policy already shared by `_notebooks`/`_artifacts`.
- Keep behavior on valid responses byte-for-byte identical; add explicit empty-chunk guards so the strict-mode `safe_index` does not raise on inputs the legacy short-circuit ignored.
- Add malformed-chunk resilience tests covering empty chunks, missing `item[5]`, non-list inner items, deeply nested malformed structures, and a `decode_response` end-to-end case that mixes malformed + valid chunks through `_find_wrb_status`.

## Why
Hand-rolled `try/except IndexError` and `len(item) < N` guards were drifting away from the shared `safe_index` policy (issue I1 / T8.D1). Routing through `safe_index` means: (a) one place owns the soft/strict toggle, (b) drift logs carry a consistent `source=` and `method_id=` for diagnostics, and (c) future call sites have a single idiom to copy.

## Test plan
- [x] `uv run ruff format .`
- [x] `uv run ruff check .`
- [x] `uv run pytest tests/unit` — 3357 passed
- [x] `uv run pytest tests/integration --ignore=tests/integration/cli_vcr/test_downloads.py` — 639 passed
- [x] `NOTEBOOKLM_STRICT_DECODE=1 uv run pytest tests/unit/test_decoder.py` — 81 passed (verifies strict-mode behavior)
- [ ] Note: `tests/integration/cli_vcr/test_downloads.py` has 7 pre-existing failures on `origin/main` (VCR cassette `rpcids` mismatch — unrelated to this change; verified by re-running on the base commit).

### DIFF
--- a/src/notebooklm/rpc/decoder.py
+++ b/src/notebooklm/rpc/decoder.py
@@ -298,19 +298,28 @@ def collect_rpc_ids(chunks: list[Any]) -> list[str]:
     Returns:
         List of RPC method IDs found in the response.
     """
+    source = "decoder.collect_rpc_ids"
     found_ids = []
     for chunk in chunks:
         if not isinstance(chunk, list):
             continue
 
-        items = chunk if (chunk and isinstance(chunk[0], list)) else [chunk]
+        # Preserve the truthy short-circuit on an empty chunk so safe_index
+        # (which would raise under NOTEBOOKLM_STRICT_DECODE=1) is only called
+        # when the index is structurally valid.
+        if not chunk:
+            continue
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

PR #611, opened 2026-05-15
docs(auth): document best-effort dedupe on warning flags (T7.G7)

teng-lin:
## Summary

- Add comment blocks above the two module-level warning flags in `auth.py` documenting the dedupe contract: **best-effort under threads, exactly-once on a single event loop**.
- Add `tests/unit/test_warning_dedupe.py` with two tests that gather 100 concurrent coroutines into each warning path and assert the warning fires exactly once.
- Architectural decision: Option B (documented best-effort) — locked in iter-1 of the plan. Threading is out of scope per the documented concurrency contract; spending a PR on exact-once cross-thread logging adds complexity without proven harm.

## Task

`.sisyphus/plans/tier-7-thread-safety-concurrency/phase-2-wave-5.md#T7.G7` (audit §24).

## Why this is correct on a single loop

Both flags share the same shape:

```python
global _SOME_FLAG
if not _SOME_FLAG:
    _SOME_FLAG = True
    logger.warning(...)
```

The asyncio scheduler can only switch coroutines at `await` points. The check-then-set is a single synchronous block with no intervening `await`, so concurrent coroutines on one loop cannot interleave between reading `_SOME_FLAG` as `False` and writing it as `True`. Under `asyncio.gather(*[100 coros])`, the warning fires exactly once. The unit test pins this.

Under genuine OS threads (which this library does NOT support — each client is bound to one event loop), two threads can both observe `False` before either has written `True`, causing a duplicate warning. That's the "best-effort" half of the contract.

## What we explicitly did NOT do (per spec MUST NOT DO)

- No `asyncio.Lock` for a logging warning.
- No re-architecture of logger handlers.
- No Option A (real cross-thread dedupe).

The comment blocks also explicitly debunk two tempting-but-broken alternatives:

- `functools.lru_cache` memoizes return values, not the `logger.warning` **side-effect**.
- `logging.LoggerAdapter` only rewrites records, it does not filter duplicates.

## Line drift note

The phase plan referenced `auth.py:1717` for the flock flag; in current `origin/main` the flag sits at `:1728` after T7.D1 (#574) added a few lines above it. Same flag, same call site — the drift is cosmetic.

## Test plan

- [x] `uv run ruff format .`
- [x] `uv run ruff check .`
- [x] `uv run pytest tests/unit/test_warning_dedupe.py` — 2 tests pass
- [x] `uv run pytest tests/unit/test_auth.py` — 292 tests pass (no regressions from comment edits)
- [x] `uv run pytest tests/unit/` — 3441 passed, 8 skipped

## Deferred polish findings

None.

## Summary by CodeRabbit

* **Documentation**
  * Enhanced inline documentation for warning deduplication behavior related to cookies and file-lock unavailability.

* **Tests**
  * Added comprehensive unit tests verifying warning deduplication under concurrent execution scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/611)

### DIFF
--- a/src/notebooklm/auth.py
+++ b/src/notebooklm/auth.py
@@ -135,6 +135,25 @@ _EXTRACTION_HINT = (
 
 # Tier 2 fires per cookie-load; a single CLI run can hit it 2-3 times across
 # the four loader entry points. One warning per process is enough signal.
+#
+# Dedupe contract (T7.G7): best-effort under threads, exactly-once on a single
+# event loop. The check-then-set at the call site (``_validate_required_cookies``
+# below) reads ``_SECONDARY_BINDING_WARNED`` and sets it to ``True`` in a single
+# synchronous block with no intervening ``await``. The asyncio scheduler can
+# only switch coroutines at ``await`` points, so concurrent coroutines on one
+# loop cannot interleave between the check and the set — the warning fires
+# exactly once per process. Under genuine OS threads (which this library does
+# NOT support per the documented concurrency contract — each client is bound
+# to one event loop), the pattern is racy: two threads can both observe
+# ``False`` before either has written ``True``, causing a duplicate warning.
+# We accept that as best-effort rather than introduce an ``asyncio.Lock``
+# (would not help threads) or a ``threading.Lock`` (re-architects for a use
+# case we don't support).
+#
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

PR #612, opened 2026-05-15
fix(client): canonicalize keepalive storage path for in-process dedupe (T7.G6)

teng-lin:
## Summary

- Canonicalize `keepalive_storage_path` via `Path(p).expanduser().resolve()` in `NotebookLMClient.__init__` so different syntactic representations of the same file (relative vs absolute, `~` shorthand, symlink components) hash to one dedupe entry. Mirrors the auth-refresh canonicalization at `auth.py:_fetch_tokens_with_refresh` (line ~2610).
- Without this, two clients pointing at the same physical file via different path syntaxes get distinct `_LAST_POKE_ATTEMPT_MONOTONIC` / `_POKE_LOCKS_BY_LOOP` / `_rotation_lock_path` keys and bypass the in-process dedupe — duplicate `RotateCookies` POSTs.
- Public path argument type (`str | Path | None`) and `auth.storage_path` are preserved as the caller supplied them; only the internal-derived `ClientCore._keepalive_storage_path` is canonicalized.

## Task

- Plan: `.sisyphus/plans/tier-7-thread-safety-concurrency/phase-2-wave-5.md#T7.G6`
- Audit reference: thread-safety-concurrency-audit §29 (`Keepalive rotation throttle keys raw Path objects (no canonicalization)`)

## Test plan

- [x] `uv run mypy src/notebooklm --ignore-missing-imports` — 60 files, no issues
- [x] `uv run pytest tests/integration/concurrency/ tests/unit/test_client_keepalive.py` — 98 passed
- [x] New regression test `tests/integration/concurrency/test_keepalive_path_canonicalize.py` covers four scenarios:
  - relative vs absolute paths to the same file share the dedupe key
  - `~`-prefixed path is expanded
  - public `storage_path` arg + `auth.storage_path` remain caller-supplied (not canonicalized)
  - `None` pass-through stays `None`
- [x] Full `uv run pytest` — 4091 passed; 7 pre-existing cassette-guard failures in `cli_vcr/test_downloads.py` are unrelated (`_rpcids_matcher` assertion, see Wave 2/3 lessons).

## Deferred polish findings

none — surgical 10-line change plus regression test.

## Release Notes

* **Bug Fixes**
  * Enhanced path resolution for storage configurations to ensure consistent behavior across different input formats, including relative, absolute, and home-directory references.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/612)

### DIFF
--- a/src/notebooklm/client.py
+++ b/src/notebooklm/client.py
@@ -153,6 +153,24 @@ class NotebookLMClient:
         if storage_path is not None and auth.storage_path != storage_path:
             auth = dataclasses.replace(auth, storage_path=storage_path)
 
+        # Canonicalize the keepalive storage path so different representations
+        # of the same physical file (relative vs absolute, ``~`` shorthand,
+        # symlink components) hash to the same key in the in-process rotation
+        # dedupe (``_get_poke_lock`` / ``_try_claim_rotation`` /
+        # ``_rotation_lock_path`` in auth.py). The auth refresh path already
+        # canonicalizes at ``auth.py:_fetch_tokens_with_refresh`` via
+        # ``Path(p).expanduser().resolve()``; this mirrors it so two clients
+        # pointing at the same file via different path syntaxes share one
+        # ``_LAST_POKE_ATTEMPT_MONOTONIC`` entry instead of bypassing dedupe
+        # and firing duplicate ``RotateCookies`` POSTs (audit §29 / T7.G6).
+        # NOTE: the public ``storage_path`` argument and ``auth.storage_path``
+        # are intentionally left as the caller provided them — only the
+        # internal-derived ``ClientCore._keepalive_storage_path`` is
+        # canonicalized.
+        keepalive_storage_path: Path | None = auth.storage_path
+        if keepalive_storage_path is not None:
+            keepalive_storage_path = Path(keepalive_storage_path).expanduser().resolve()
+
         # Pass refresh_auth as callback for automatic retry on auth failures
         # Note: refresh_auth calls update_auth_headers internally
         self._core = ClientCore(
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
## Review feedback addressed in 1c092fd

Thanks @coderabbitai and @claude for the reviews.

**@coderabbitai** — nitpick (symlink resolution test): applied. New `test_symlink_is_resolved` creates a symlink alongside the real file and verifies both clients yield the same canonical keepalive key.

**@claude[bot]** — `test_public_storage_path_argument_unchanged` weak-assertion finding: applied. The prior test compared two already-canonical absolute paths (since pytest's `tmp_path` is already resolved) and proved nothing. Switched to a genuinely non-canonical relative-from-cwd path via `monkeypatch.chdir`, plus an explicit `raw_path != target.resolve()` sanity guard so the test would FAIL if a future change accidentally canonicalized the public argument.

**@claude[bot]** — comment length: keeping the 14-line block. The rationale (audit §29 cross-reference, the asymmetric design choice, the explicit note that `auth.storage_path` is intentionally NOT canonicalized) is genuinely non-obvious; trimming it risks the next maintainer "fixing" the asymmetry. Per Claude bot's own assessment: "not blocking given the complexity of the invariant."

**@claude[bot]** — `_core.py:421-423` fallback trap: deferred. Out of scope for T7.G6 (the audit explicitly scoped the fix to `client.py` lines 124-135 / 224-236). `ClientCore` is an internal class; the only in-tree caller is `NotebookLMClient.__init__`, which now always passes a canonical path. If hardening that fallback later becomes desirable, it should be its own focused PR (audit §29 follow-up).

**@gemini-code-assist** — quota-exhausted on this PR run; nothing to address from that lens.

All 99 keepalive + concurrency tests still pass; ruff + mypy clean.

teng-lin:
## Windows portability fix in 6d5d894

The first push's Windows runner (Python 3.14) failed `test_tilde_path_is_expanded` because `Path.expanduser()` on Windows consults `USERPROFILE` (not `HOME`). Pushed a fix that overrides both `HOME` and `USERPROFILE`, clears the `HOMEDRIVE`/`HOMEPATH` fallback, and adds a `pytest.skip` guard so the test fails-closed if it can't override `~` hermetically on a future platform.

The 3 other failures on the same Windows run (`test_add_file_toctou`, two `test_download_blocks_loop` event-loop-blocking assertions) are pre-existing on main and unrelated to T7.G6 — outside this PR's scope per the Wave 2/3 lessons.

PR #619, opened 2026-05-15
fix(research): optional task_id discriminator in poll/import_sources (T7.F3)

teng-lin:
## Summary

Resolves the audit §26 research cross-wire: when two research tasks are in flight on the same notebook, `ResearchAPI.poll(notebook_id)` silently returned the *latest* task, so a caller that started task A could unknowingly act on task B's results.

Adds two non-breaking guards:

- **`poll()` discriminator** — new optional `task_id: str | None = None` filters `parsed_tasks` to the matched task when supplied; emits `DeprecationWarning` when `None` AND multiple tasks are in flight (the actually-broken case); preserves legacy single-task behavior with no warning.
- **`import_sources()` per-source validation** — any source whose `research_task_id` disagrees with the caller's `task_id` raises the new `ResearchTaskMismatchError` (subclass of `ValidationError`) before any RPC fires. Catches mis-attribution at the boundary instead of silently overriding caller intent.

The `research wait` CLI pins to the discovered `task_id` on subsequent poll iterations so a concurrent caller can't substitute a different task's results into the wait loop.

## Task

- Plan: `.sisyphus/plans/tier-7-thread-safety-concurrency/phase-2-wave-4.md#T7.F3`
- Architectural decision (locked): optional discriminator (NOT required) preserves back-compat; `None` default scheduled for removal in a future major release (Tier 8).

## Behavior matrix

| Caller passes | In-flight count | Behavior |
|---|---|---|
| `task_id="A"` | any | returns task A (or `no_research` if not found); no warning |
| `task_id=None` | 1 | returns that task; no warning (legacy path) |
| `task_id=None` | 2+ | returns latest + `DeprecationWarning` citing migration |

## Legacy test updates

Four existing `tests/unit/test_research.py` cases passed `task_id="task_123"` while their fixture sources carried `research_task_id="report_123"` — deep research has *two* IDs (task_id from `start()`, report_id from `poll()`). The legacy code silently overrode the caller's `task_id` with the source-derived one. The new contract makes the source-derived id authoritative — callers must pass `poll_result["task_id"]` (not `start_result["task_id"]`) when importing deep-research sources. Tests updated to match. The `research wait` CLI was already correct (it captures `task_id` from `poll`'s response).

## Test plan

- [x] `tests/integration/concurrency/test_research_task_crosswire.py` — 4 red-then-green scenarios (A explicit-id, B single-no-warn, C multi-warn, D import-mismatch).
- [x] `uv run pytest tests/integration/concurrency/ tests/unit/test_research.py` — 145 passed (1 expected warning in `test_poll_returns_all_tasks` exercising the back-compat path).
- [x] `uv run ruff format . && uv run ruff check . && uv run mypy src/notebooklm --ignore-missing-imports` — clean.
- [x] Full `uv run pytest` — 4091 passed; 7 failures pre-exist on `origin/main` in `tests/integration/cli_vcr/test_downloads.py` (cassette-guard, non-blocking per wave 2/3 lessons).

## Deferred polish findings

None.


teng-lin:
@claude review

teng-lin:
Addressed finding #1 in the latest push: `test_poll_returns_all_tasks` (legacy multi-task test) now wraps the `client.research.poll("nb_123")` call in `pytest.warns(DeprecationWarning, match="task_id")` so the new T7.F3 contract on the ambiguous case is pinned on this legacy code path too. Findings #2-#5 deferred per their classifications (style nit / nice-to-have / observations).

PR #620, opened 2026-05-15
fix(core): _auth_snapshot_lock + snapshot-aware _build_url (T7.F2)

teng-lin:
## Summary

Closes the torn-read window between `ClientCore._snapshot()` and `refresh_auth`'s csrf/sid mutation (audit §12).

- New `_auth_snapshot_lock` (NOT `_refresh_lock` — avoids reentrancy and refresh-task latency on RPCs); lazy-init mirrors `_reqid_lock`.
- `_snapshot()` becomes `async def` and acquires the lock for the four-scalar read; critical section is purely synchronous reads (no awaits inside).
- `_build_url()` refactored to consume `_AuthSnapshot` rather than reading `self.auth.session_id` / `authuser` / `account_email` LIVE — that live read was the actual hazard.
- `refresh_auth`'s csrf+sid mutation block wrapped in `async with self._core._get_auth_snapshot_lock()`; tiny critical section, no awaits.

## `_build_url` audit (per acceptance criteria)

Pre-fix the method read three scalars off `self.auth` directly:
- `self.auth.session_id` → `f.sid` URL param.
- `self.auth.authuser` → `authuser` URL param.
- `self.auth.account_email` → preferred over `authuser` index.

All three are already fields on `_AuthSnapshot`, so the refactor is a pure plumbing change — no new fields needed. New AST guard (`test_build_url_does_not_read_self_auth`) rejects any future `self.auth.<field>` access in this method.

## Tests

- `tests/integration/concurrency/test_auth_snapshot_torn_read.py` (new red→green): fans 50 RPCs while triggering a refresh halfway; tags csrf/sid/cookies with a monotonic generation counter under the lock; asserts every captured POST carries generation-coherent values across all three axes.
- `tests/unit/test_concurrency_refresh_race.py` (extended):
  - `test_build_url_does_not_read_self_auth` — AST guard, new.
  - `test_snapshot_acquires_auth_snapshot_lock` — AST guard, new.
  - Existing `test_perform_authed_post_has_no_await_before_post_per_iteration` still passes — the new `await self._snapshot()` lives outside the try block.

Test callers of `_build_url` (test_client.py × 3, test_env_base_url.py × 1, test_core.py × 3) updated to construct an `_AuthSnapshot` inline.

## Test plan

- [x] `uv run ruff format . && uv run ruff check .` clean.
- [x] `uv run mypy src/notebooklm --ignore-missing-imports` clean.
- [x] `uv run pytest tests/integration/concurrency/ tests/unit/test_concurrency_refresh_race.py` — 72 passed.
- [x] Touched suites: `tests/unit/test_client.py`, `tests/unit/test_env_base_url.py`, `tests/integration/test_core.py` — all green (183 total).
- [x] Full suite: 4090 passed; 7 pre-existing VCR-cassette failures in `tests/integration/cli_vcr/test_downloads.py` (confirmed on `origin/main` — out of scope).

## Architectural notes

- **Lock NOT reused**: `_refresh_lock` is owned by refresh-task creation and held across `await self._refresh_callback()`. Reusing it would (a) re-introduce the reentrancy ambiguity T7.F2 explicitly avoids, and (b) serialize all RPC snapshot reads to refresh-task latency. The new `_auth_snapshot_lock` is uncontested except during the two-scalar refresh write.
- **No awaits inside the lock**: both call sites (`_snapshot`'s read, `refresh_auth`'s write) are tiny sync critical sections.
- **Cookie axis**: `_AuthSnapshot` still doesn't carry cookies — cookie coherence with csrf/sid stays guaranteed by the existing no-await invariant between `_snapshot()` returning and `client.post(...)` (per `test_perform_authed_post_has_no_await_before_post_per_iteration`).

## Deferred polish findings

None — all polish-pass cleanups applied (removed unused locals in the red test, dropped dead `_TAG_PATTERN` constant, tightened docstrings).

Spec: `.sisyphus/plans/tier-7-thread-safety-concurrency/phase-2-wave-4.md#T7.F2`

### DIFF
--- a/src/notebooklm/_core.py
+++ b/src/notebooklm/_core.py
@@ -410,6 +410,18 @@ class ClientCore:
         # Lazily-created — ``asyncio.Lock()`` needs a running loop in some
         # Python versions, and this object can be constructed outside one.
         self._reqid_lock: asyncio.Lock | None = None
+        # Serializes ``_AuthSnapshot`` reads in :meth:`_snapshot` with the
+        # refresh-side mutation block in :meth:`NotebookLMClient.refresh_auth`
+        # (audit §12 / T7.F2). The lock holds only across the four
+        # ``self.auth.*`` scalar reads / two scalar writes — never across
+        # an ``await`` — so RPC throughput isn't serialized to refresh
+        # latency. Lazy-init mirrors ``_reqid_lock`` because ``asyncio.Lock()``
+        # needs a running loop in some Python versions. Distinct from
+        # ``_refresh_lock`` (which is owned by refresh-task creation and
+        # held across ``await self._refresh_callback()``): mixing the two
+        # would re-introduce the reentrancy ambiguity T7.F2 set out to
+        # avoid.
+        self._auth_snapshot_lock: asyncio.Lock | None = None
         # OrderedDict for FIFO eviction when cache exceeds MAX_CONVERSATION_CACHE_SIZE
         self._conversation_cache: OrderedDict[str, list[dict[str, Any]]] = OrderedDict()
         # Keepalive background task configuration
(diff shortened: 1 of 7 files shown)


teng-lin:
@claude review

teng-lin:
Thanks @claude — addressed all three findings in 952a8b7:

**Finding 1 (IMPORTANT — sequential, not concurrent)**: You're right that the previous structure ran two sequential `gather`s around `await refresh_callback()`, so no RPC was actually in-flight during the refresh. Restructured the test so the 50 RPC coroutines + 1 refresh coroutine are dispatched into a single `asyncio.gather(...)` — they all become ready at the same tick, and the mock transport's per-handler `asyncio.sleep(0)` lets the event loop interleave them. The refresh's lock-acquired write block now contends with concurrent `_snapshot()` reads on in-flight RPCs.

I also tightened the test's docstring to honestly frame its scope: it's the runtime smoke proof that the new design composes correctly under concurrent load. The pre-fix torn read in `_build_url` only materializes if a yield point slips into `_perform_authed_post`'s prologue between snapshot capture and request build — which the existing AST guards (`test_perform_authed_post_has_no_await_before_post_per_iteration`, `test_build_url_does_not_read_self_auth`, `test_snapshot_acquires_auth_snapshot_lock`) lock down statically. Together they form the regression net.

**Finding 2 (NIT — direct field access)**: The synthetic refresh callback now acquires the lock via `core._get_auth_snapshot_lock()` (the production accessor), keeping it in lockstep with how `NotebookLMClient.refresh_auth` acquires the lock. Also exercises the lazy-init path on the refresh side.

**Finding 3 (NIT — `_snapshot_for` duplication)**: Kept the inline staticmethod per-test-class. With 3 callsites (test_client.py, test_core.py, test_env_base_url.py), introducing a shared helper would require either a new module in `tests/` or exposing `_AuthSnapshot` more broadly, and the construct is a 4-line frozen-dataclass init that's robust to drift only if the dataclass shape changes — at which point the test signature breaks loudly anyway. Tradeoff documented; happy to consolidate in a follow-up if you'd like.

CI matrix: macOS/Ubuntu green; the Windows failures are pre-existing on `origin/main` (e.g. run 25926796742) — `tests/integration/cli_vcr/test_settings.py:53` `config_path.exists()` assertion on Windows-specific path handling, unrelated to this PR.

teng-lin:
@claude review

PR #744, opened 2026-05-17
fix(rpc,chat): narrow decode catches; raise on empty streaming with clear contract (I3, I4)

teng-lin:
## Summary

Closes Tier-9 audit findings **I3** (over-broad \`except Exception\` in \`_core_rpc.py\` masks code bugs as shape drift) and **I4** (streaming chat returns empty result on zero parseable chunks instead of raising).

## Exception contract

The fix is about surfacing real decoder bugs that are currently masked as "shape drift":

- **Genuine shape-drift exceptions** (\`JSONDecodeError\`, \`KeyError\`, \`IndexError\`, \`TypeError\` from indexing into a wrong-shape response) → remain wrapped as \`RPCError("Failed to decode response...")\` because these ARE shape drift.
- **Code bugs** (\`AttributeError\` from a typo, \`NameError\`, runtime errors NOT from shape) → **propagate as their native type**, NOT wrapped as RPCError.
- **Streaming chat with zero parseable chunks** → raise new \`ChatResponseParseError\`, NOT return empty result.

### \`_core_rpc.py:187\`

\`\`\`python
# Before
except Exception as exc:

# After
except (json.JSONDecodeError, KeyError, IndexError, TypeError) as exc:
\`\`\`

\`AttributeError\`/\`NameError\`/\`RuntimeError\` subclasses now propagate unmasked.

### \`_chat_protocol.py::parse_streaming_chat_response\`

Tracks \`parseable_chunk_count\` while iterating \`wrb.fr\` chunks:
- **Zero parseable chunks** (API drift) → \`raise ChatResponseParseError("no parseable chunk in streaming response")\`
- **Chunks parsed but empty answer** (legitimate empty response) → return \`StreamingChatParseResult("", final_refs, server_conv_id)\` as today

The conflation we're fixing: previously the function returned a (possibly empty) result for both cases, so a totally broken stream looked indistinguishable from a model that legitimately said nothing.

### \`exceptions.py\` + \`__init__.py\`

New \`ChatResponseParseError(ChatError)\` — inherits via \`ChatError\` (not directly \`NotebookLMError\`) so existing chat-domain catches still cover it. Exported via \`notebooklm.exceptions\` and the top-level \`notebooklm\` package.

## Tests

- \`test_core_rpc.py::test_decode_shape_error_wrapped\` — wrong-shape response → \`RPCError\`
- \`test_core_rpc.py::test_decode_code_bug_propagates\` — decoder raises \`AttributeError\` → \`AttributeError\` propagates (not wrapped)
- \`test_streaming_chat_protocol.py\` — zero-chunks → \`ChatResponseParseError\`; chunks-parsed-empty-answer → empty \`StreamingChatParseResult\`
- \`test_chat.py\` (new) — end-to-end propagation
- \`test_chat_characterization.py\` — stubs updated for new semantics

## Test plan

- [x] \`uv run ruff format .\` clean (349 files)
- [x] \`uv run ruff check .\` clean
- [x] \`uv run mypy src/notebooklm --ignore-missing-imports\` clean (100 source files)
- [x] Targeted pytest on the 4 affected test files: 146 passed in 0.24s
- [ ] Full \`uv run pytest\` (CI will run this)

## Release Notes

* **New Features**
  * Added `ChatResponseParseError` exception to the public API for handling streaming chat response parsing failures.
  * Enhanced response parsing to distinguish between legitimate empty answers and unparseable/malformed responses.

* **Bug Fixes**
  * Improved error handling to raise exceptions for unparseable chat responses instead of silently returning empty results.
  * Refined RPC error reporting to distinguish parsing errors from code-level exceptions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/744?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/notebooklm/__init__.py
+++ b/src/notebooklm/__init__.py
@@ -65,6 +65,7 @@ from .exceptions import (
     AuthExtractionError,
     # Domain: Chat
     ChatError,
+    ChatResponseParseError,
     ClientError,
     # Validation/Config
     ConfigurationError,
@@ -205,6 +206,7 @@ __all__ = [
     "NotebookLimitError",
     # Domain Exceptions: Chat
     "ChatError",
+    "ChatResponseParseError",
     # Domain Exceptions: Sources
     "SourceError",
     "SourceAddError",
--- a/src/notebooklm/_chat_protocol.py
+++ b/src/notebooklm/_chat_protocol.py
@@ -17,7 +17,7 @@ from urllib.parse import quote, urlencode
 
 from ._env import get_default_bl, get_default_language
 from .auth import format_authuser_value
-from .exceptions import ChatError
+from .exceptions import ChatError, ChatResponseParseError
 from .rpc.encoder import nest_source_ids
 from .rpc.types import get_query_url
 from .types import ChatReference
@@ -130,7 +130,21 @@ def build_streaming_chat_request(
 
 
 def parse_streaming_chat_response(response_text: str) -> StreamingChatParseResult:
-    """Parse a streamed-chat response into answer, references, and conversation ID."""
+    """Parse a streamed-chat response into answer, references, and conversation ID.
+
+    Failure contract (see :class:`notebooklm.exceptions.ChatResponseParseError`):
+
+    * **Zero parseable chunks** — no chunk in the response yielded a
+      successfully decoded ``wrb.fr`` envelope. This means either the
+      response body was empty/garbage, or the API's wire format drifted
+      and the parser no longer recognizes the envelope shape. Raises
+      :class:`ChatResponseParseError`.
+    * **Chunks parsed but empty answer** — at least one ``wrb.fr`` chunk
+      decoded, but no chunk yielded answer text (the model legitimately
+      returned an empty response). Returns
+      ``StreamingChatParseResult("", refs, conv_id)`` — empty answer is
+      a valid outcome, not a parse failure.
+    """
     if response_text.startswith(")]}'"):
         response_text = response_text[4:]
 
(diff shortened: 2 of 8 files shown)


teng-lin:
@claude review

teng-lin:
Addressed claude-bot review feedback in 6532c2a:

- **`ValueError` propagation gap** (claude bot's recommended fix): added a bare `ValueError("non-json value error")` case to `test_decode_code_bug_propagates` parametrize, so the documentation-in-tests now explicitly covers non-`JSONDecodeError` `ValueError` propagation through the narrow tuple.
- **Fall-through invariant clarity** (claude bot's suggested inline comment): added a comment at the `_extract_chunk_with_parseable` fall-through point explaining the real-world heartbeat (`"[]"` inner) case — those frames are deliberately still counted as parseable so a heartbeats-only stream surfaces as "empty answer" rather than `ChatResponseParseError`.

Also considered but deferred (with rationale):

- **Codex's stricter parseable definition**: would require the decoded inner_data to match the full chat record shape before being counted as parseable. Rejected because real-world streams send empty `wrb.fr` heartbeats (`"[]"`, `"{}"`, etc.) that the live API uses for keep-alive — tightening this would regress to wrongly raising `ChatResponseParseError` on otherwise healthy connections.
- **Codex's suggestion to strip `TypeError` from the narrow tuple**: PR-D spec explicitly includes `TypeError` because shape-drift `TypeError` (`data["key"]` on a list, `data + str`) is the dominant case in this codebase; wrong-arity `TypeError` from helper calls is rare and the trade-off is documented.
- **Gemini's namedtuple/dataclass for the 5-tuple**: future polish opportunity; the tuple is internal-only (`_extract_chunk_with_parseable` is private) and `extract_answer_and_refs_from_chunk` keeps its 4-tuple public contract.

PR #751, opened 2026-05-17
fix(core): drain poll tasks on close + streamed RPC size cap (I1, I10)

teng-lin:
## Summary

Closes Tier-9 audit findings **I1** (orphaned artifact poll tasks survive \`close()\` and wake against an already-closed HTTP transport) and **I10** (unbounded RPC response body — a misbehaving or malicious endpoint can make the client allocate arbitrary memory).

## BREAKING change

\`NotebookLMClient.__aexit__\` / \`close()\` default \`drain\` flipped from \`False\` → \`True\`. Callers relying on fire-and-forget close semantics will now block briefly while pending artifact poll tasks are cancelled and drained. Pass \`drain=False\` to opt back into the old shape.

### \`_core_polling.py\` — \`PollRegistry.active_tasks()\`

New method returning a snapshot list of pending leader poll tasks (already-completed tasks filtered out). Snapshot before iteration so a finishing leader removing its entry can't race the cancel/gather pair.

### \`_core.py\` — \`close()\` drains pending polls

After cancelling \`_keepalive_task\` and **before** tearing down the HTTP client: iterate \`self.poll_registry.active_tasks()\`, \`cancel()\` each, then \`await asyncio.gather(*tasks, return_exceptions=True)\`. \`return_exceptions=True\` so a single misbehaving task can't block the rest of the close sequence.

### \`client.py\` — \`drain=True\` default

Documented inline as a behavioral change.

### \`_core_transport.py\` — \`_stream_post_with_size_cap\`

RPC POST path now uses \`httpx.AsyncClient.stream("POST", ...)\` with a running size guard. Body read chunk-by-chunk; when total exceeds \`MAX_RPC_RESPONSE_BYTES\` (50 MiB), \`RPCResponseTooLargeError\` is raised and the stream is aborted. Buffered bytes are re-attached to a fresh \`httpx.Response\` (same status, headers, request) so downstream callers can keep using \`response.text\` / \`response.content\` unchanged.

Existing transport semantics preserved verbatim:
- 429 / Retry-After path
- 5xx / network retry path
- Auth-refresh path

\`raise_for_status()\` is invoked inside the streaming context so the HTTPStatusError chain (with \`exc.response.headers\` intact) reaches \`AuthedTransport\` unchanged.

### Import-cycle avoidance

\`MAX_RPC_RESPONSE_BYTES\` lives in \`_core_transport.py\` near \`MAX_RETRY_AFTER_SECONDS\` — NOT in \`_core.py\` — because \`_core.py\` imports \`_core_transport\` and a reverse import would cycle.

### \`exceptions.py\` — \`RPCResponseTooLargeError(RPCError)\`

Inherits from existing \`RPCError\` so existing \`except RPCError\` clauses still cover the size-cap path. Carries \`limit_bytes\` and \`bytes_read\` for diagnostic logging.

## Tests

- \`tests/unit/test_core_close.py\` (new): \`test_core_close_drains_polls\` — parked poll task cancelled within 1s of close() and the \`except asyncio.CancelledError\` branch runs.
- \`test_core_transport.py\`: streaming wrapper coverage incl. size-cap-exceeded, exact-cap-boundary, normal-body, HTTPStatusError-from-raise_for_status paths.
- \`test_client.py\`: drain=True default + per-call override.
- \`tests/unit/conftest.py\`: new \`install_post_as_stream\` helper adapts legacy \`fake_post\`-style mocks to the streaming context manager API.

## Test plan

- [x] \`uv run ruff format .\` clean (349 files)
- [x] \`uv run ruff check .\` clean
- [x] \`uv run mypy src/notebooklm --ignore-missing-imports\` clean (100 source files)
- [x] Full \`uv run pytest tests/unit -q\`: 4147 passed, 8 skipped, 0 failed
- [ ] CI: full suite + e2e

### DIFF
--- a/src/notebooklm/__init__.py
+++ b/src/notebooklm/__init__.py
@@ -83,6 +83,7 @@ from .exceptions import (
     # Domain: Research
     ResearchTaskMismatchError,
     RPCError,
+    RPCResponseTooLargeError,
     RPCTimeoutError,
     ServerError,
     # Domain: Sources
@@ -194,6 +195,7 @@ __all__ = [
     "AuthExtractionError",
     "NetworkError",
     "RPCTimeoutError",
+    "RPCResponseTooLargeError",
     "RateLimitError",
     "ServerError",
     "ClientError",
--- a/src/notebooklm/_core.py
+++ b/src/notebooklm/_core.py
@@ -1090,6 +1090,14 @@ class ClientCore:
         ``self._http_client = None`` runs in an inner ``finally`` so
         the instance is consistently marked closed even if the
         shielded ``aclose`` itself raises.
+
+        Poll-task drain: in-flight artifact poll tasks held by
+        :attr:`poll_registry` are cancelled and awaited before the HTTP
+        client is torn down. Without this, a leader poll waking mid-aclose
+        would issue a request against an already-closed transport and
+        surface as a confusing httpx error in the user's logs. The drain
+        uses ``return_exceptions=True`` so a single misbehaving task can't
+        block the rest of the close sequence.
         """
         try:
             # Stop the keepalive task before tearing down the HTTP client so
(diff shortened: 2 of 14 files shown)


teng-lin:
@claude review

PR #758, opened 2026-05-17
fix(tests): adapt integration mocks to streaming POST (regression from #751)

teng-lin:
## Summary

Main is currently red on 15 integration tests after #751 merged. PR #749 (and presumably every other in-flight PR) inherits the breakage on every CI run.

#751 switched the RPC POST path from \`client.post(...)\` to \`client.stream(\"POST\", ...)\` so the new \`MAX_RPC_RESPONSE_BYTES\` size guard could fire chunk-by-chunk. The companion \`install_post_as_stream\` adapter was wired into \`tests/unit/conftest.py\` only — 15 integration tests under \`tests/integration/\` still mock \`http_client.post\` and surface as two distinct symptoms:

- **11 tests** in \`test_auto_refresh.py\` / \`test_core.py\`: \`Error -3 while decompressing data: incorrect header check\` — the legacy \`http_client.post\` mock never fires, \`client.stream()\` falls through to a default response with a mismatched content-encoding header.
- **4 tests** in \`concurrency/test_rate_limit_default.py\`: \`TypeError: 'async for' requires an object with __aiter__ method, got coroutine\` — \`AsyncMock(spec=httpx.AsyncClient).stream(...)\` returns a coroutine, not an async context manager.

Confirmed identical 15-test failure cluster on main HEAD (\`c5d1ce3c\`, Test run \`25988127281\`) and on PR #749's latest commit. No production code changes — pure test-side adaptation matching the unit-tree pattern #751 already introduced.

## Changes

- New \`tests/integration/_streaming_mock.py\` exposing the same \`install_post_as_stream\` helper as \`tests/unit/conftest.py\`. Lives as a sibling of \`conftest.py\` rather than inside it so \`tests/integration/concurrency/\` (which has its own \`conftest.py\`) can also import it via \`tests/integration/\` already being on \`sys.path\`.
- Each of the 15 failing tests calls \`install_post_as_stream(None, http_client, fake_post)\` immediately after wiring up \`http_client.post\` so the streaming wrapper sees the same mocked behavior.

## Test plan

- [x] Targeted: \`pytest tests/integration/test_core.py tests/integration/test_auto_refresh.py tests/integration/concurrency/test_rate_limit_default.py\` → **58 passed** (was 15 failed before).
- [x] Full suite: \`pytest\` → **4809 passed, 12 skipped, 0 failed** (was 4794 passed, 15 failed).
- [x] \`ruff format\`, \`ruff check\`, \`mypy src/notebooklm --ignore-missing-imports\` all clean.
- [ ] CI matrix should now go fully green; rebase / re-run PR #749 once this merges.

## Summary by CodeRabbit

* **Tests**
  * Enhanced integration test infrastructure to properly support streaming HTTP client mocking, ensuring consistent behavior validation across rate-limiting, token refresh, and RPC call scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/758?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/integration/_streaming_mock.py
+++ b/tests/integration/_streaming_mock.py
@@ -1,0 +1,93 @@
+"""Bridge legacy ``http_client.post`` mocks to the streaming POST API.
+
+The RPC POST path now uses :meth:`httpx.AsyncClient.stream` so the
+running size guard in :data:`notebooklm._core_transport.MAX_RPC_RESPONSE_BYTES`
+can fire. Many integration tests in this tree predate that switch and
+still express intent as ``client._http_client.post = fake_post`` or
+``patch.object(..., "post", side_effect=err)``. This module exposes the
+same ``install_post_as_stream`` helper as ``tests/unit/conftest.py`` so
+both test trees can adapt legacy mocks without rewriting each call site.
+
+Why a sibling module instead of putting this in ``conftest.py``: the
+``tests/integration/concurrency/`` subdirectory has its own ``conftest.py``,
+and ``from conftest import ...`` from a test in that subdir resolves to
+the closer conftest. Pytest adds ``tests/integration/`` to ``sys.path``
+when collecting any of its tests (because ``conftest.py`` lives here),
+so ``from _streaming_mock import install_post_as_stream`` works from
+both ``tests/integration/`` and ``tests/integration/concurrency/`` without
+``sys.path`` hackery.
+"""
+
+from collections.abc import Awaitable, Callable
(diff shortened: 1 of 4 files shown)


teng-lin:
Superseded by a2c7bf4 on #749, which bundles the same streaming-mock adapter fix directly into the PR that triggered the investigation. Both approaches resolve the 15-test regression from #751 — keeping #749's bundled version for less branch-juggling.

PR #761, opened 2026-05-17
test(logging): stabilize fast-path benchmark

teng-lin:
## Summary
- Stabilize the logging fast-path microbenchmark so coverage-instrumented CI still verifies a material speedup without requiring a brittle 5x runner-specific ratio.

## Verification
- uv run --extra dev pytest -n auto tests/unit/test_logging.py::test_fast_path_microbenchmark_speedup_ratio
- uv run --extra dev ruff check tests/unit/test_logging.py src/notebooklm/_types/common.py
- uv run --extra dev ruff format --check tests/unit/test_logging.py src/notebooklm/_types/common.py

## Notes
- The earlier formatting and RPC-health changes are already on main via #762. This PR is now rebased from main and only changes tests/unit/test_logging.py.

## Summary by CodeRabbit

* **Tests**
  * Adjusted logging performance benchmark acceptance criteria.
  * Updated benchmark test documentation to reflect revised performance expectations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/761?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #765, opened 2026-05-17
test(logging): stabilize fast-path benchmark threshold

teng-lin:
## Summary

- The `test_fast_path_microbenchmark_speedup_ratio` test on `main` asserts a 5× speedup ratio between the fast-path-gated `scrub_secrets` call and a forced full-regex sweep on identical innocuous input. CI runners under coverage instrumentation consistently land around 3× (observed: 3.11× on macOS, gated=51ms vs bypassed=160ms over 10k iterations), turning this into a recurring red on otherwise-clean PRs.
- The check exists to guard *that the fast-path is doing meaningful work*, not to publish a specific runtime number. A 2× threshold preserves the regression signal (a busted fast-path would collapse the ratio to ~1×) without coupling the test suite to coverage overhead or Python-version JIT/optimizer behavior.
- Cherry-picks an already-authored commit from a local types-refactor branch where it was a drive-by stabilizer; reposting standalone so the threshold relaxation isn't bundled with unrelated work.

## Test plan

- [x] `uv run pytest tests/unit/test_logging.py::test_fast_path_microbenchmark_speedup_ratio` passes locally (Python 3.11, macOS, 0.09s).
- [ ] CI passes across the matrix (the failure being fixed only appeared on macOS + Linux coverage runs).
- [ ] Sanity: `tests/unit/test_logging.py::test_fast_path_still_redacts_when_token_present` still passes — confirms the gate is still a gate, not bypassed.

### DIFF
--- a/tests/unit/test_logging.py
+++ b/tests/unit/test_logging.py
@@ -601,12 +601,13 @@ def test_fast_path_skips_innocuous_messages_unchanged():
 
 
 def test_fast_path_microbenchmark_speedup_ratio(monkeypatch):
-    """Fast-path must speed up innocuous redactions by at least 5×.
+    """Fast-path must materially speed up innocuous redactions.
 
     NOT a fixed-time assertion (machine-dependent). Instead measures the
     ratio of full-regex-sweep time vs. gated time on identical inputs. The
-    ratio is robust to CPU speed; we run the same workload twice on the
-    same process.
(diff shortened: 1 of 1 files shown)


teng-lin:
Superseded by #761, which already merged the same logging benchmark stabilization to main as 2567688. Closing this duplicate to keep the PR queue clean.

PR #771, opened 2026-05-17
fix(core): strip content-encoding on buffered RPC response (#769)

### LINKED ISSUE #769: RPC calls fail: "Error -3 while decompressing data: incorrect header check" — raw httpx POST to batchexecute decodes fine, only library path fails
## Summary

Every batchexecute RPC (`LIST_NOTEBOOKS`, etc.) fails with:

```
RPC LIST_NOTEBOOKS failed after ~8s: Error -3 while decompressing data: incorrect header check (server-error retries exhausted)
Network error: Request failed calling LIST_NOTEBOOKS: Error -3 while decompressing data: incorrect header check
```

Auth is healthy (`auth check --test --json` → `status: ok`, all checks `true`; `doctor` all green). The failure is **isolated to the library's RPC path**: a manual `httpx` POST to the exact same `batchexecute` endpoint, using the same profile cookies, in the same Python/environment, decodes the response **without error**.

## Environment

- OS: Windows 10/11 (x64), Microsoft Store Python
- Python: 3.13.13
- notebooklm-py: 0.4.1, installed from `git+https://github.com/teng-lin/notebooklm-py.git` — reproduced on **latest `main`** commit `77802dcc4461a6c45296a2b583f7146fe4d542a5` (also failed on `c7479da91d66e2b03658184d9c997a638acfb533`)
- httpx: 0.28.1
- `brotli` and `zstandard` installed (no change in behavior)
- No proxy env vars set

## What I ruled out (with evidence)

1. **TLS interception / AV MITM** — TLS cert for `notebooklm.google.com` is `Issuer: CN=WR2, O=Google Trust Services, C=US` (genuine Google; McAfee + Windows Defender present but **not** intercepting).
2. **Missing decoders** — installed `brotli` + `zstandard`; error unchanged.
3. **General httpx/gzip problem in this environment** — `httpx.get("https://notebooklm.google.com/")` → 200, `Content-Encoding: gzip`, ~962 KB, decodes fine.
4. **batchexecute endpoint itself / cookies** — a manual `httpx` POST to `https://notebooklm.google.com/_/LabsTailwindUi/data/batchexecute` using the cookies from `~/.notebooklm/profiles/default/storage_state.json` returns HTTP 400 (dummy body) with `Content-Encoding: gzip` and **`response.text` decodes cleanly** — tested with `Accept-Encoding` default, `identity`, and `gzip`. No `incorrect header check`.
5. **HTTP/2** — library builds its `httpx.AsyncClient` without `http2=True` (and `h2` is not installed), so this is HTTP/1.1, same as the working manual test.
6. **Not intermittent** — fails identically on every attempt; `notebooklm doctor` / `auth check` stay green.

## Where it fails

`notebooklm/_core_rpc.py:165` → `result = self._decode_response(response.text, ...)`. Accessing `response.text` on the library's `AsyncClient` batchexecute POST raises the `incorrect header check` decompression error, even though an equivalent manual `httpx` POST to the same URL with the same cookies does not.

## Minimal manual repro that *works* (proving env is clean)

```python
import json, httpx
from pathlib import Path
ss = json.loads(Path.home().joinpath(".notebooklm/profiles/default/storage_state.json").read_text("utf-8"))
jar = httpx.Cookies()
for c in ss.get("cookies", []):
    jar.set(c["name"], c["value"], domain=c.get("domain","").lstrip("."), path=c.get("path","/"))
r = httpx.post(
    "https://notebooklm.google.com/_/LabsTailwindUi/data/batchexecute",
    data={"f.req": '[[["wXbhsf","[null,1]",null,"1"]]]'},
    cookies=jar,
    headers={"content-type": "application/x-www-form-urlencoded;charset=UTF-8"},
    timeout=25,
)
print(r.status_code, r.headers.get("content-encoding"), len(r.text))  # 400 gzip 203 — decodes fine
```

Same call **through the library** (`notebooklm list`) → `incorrect header check`.

## Note

There appear to be many automated `RPC Health Check: Non-transient ERROR detected` issues filed on 2026-05-17. This report adds the isolating detail that **raw httpx against the same endpoint/cookies decodes fine**, so the decompression failure seems to originate in the library's `AsyncClient`/RPC transport path (e.g. `follow_redirects=True` + async + retry/decode handling), not in the network, the environment, or httpx's gzip decoder generally.

## Question

Is this a known regression in the RPC transport/decode path on current `main`? Is there a recommended workaround (e.g. a way to force `Accept-Encoding: identity`, disable redirect-following, or a transport override) until a fix lands? Happy to provide more traces.

teng-lin:
## Summary

Closes #769 — every RPC fails with `Error -3 while decompressing data: incorrect header check` on current `main`.

`_stream_post_with_size_cap` (introduced in #751) reads response chunks via `response.aiter_bytes()`, which yields **already-decoded** body bytes, then rebuilds a fresh `httpx.Response` with `headers=response.headers`. The copied headers still carry `Content-Encoding: gzip` (the default for `batchexecute`), so the rebuilt Response's `__init__` calls `self.read()` → `iter_bytes()` → gzip decoder on bytes that have already been decompressed, raising `DecodingError`.

Minimal repro (no network):

```python
import httpx
httpx.Response(
    status_code=200,
    headers={"content-encoding": "gzip"},
    content=b"[[\"hello\"]]",
)
# zlib.error: Error -3 while decompressing data: incorrect header check
```

The reporter's manual `httpx.post` worked because it's a single Response object — content-encoding is consumed exactly once.

## Fix

Strip `content-encoding` and the now-mismatched `content-length` from the headers before constructing the buffered Response. The body is already decoded at that point, so the rebuilt Response should not advertise either.

## Tests

- New: `test_streaming_strips_content_encoding_to_prevent_double_decode` — exercises the gzip-header path directly. Pre-fix this raised `DecodingError` inside `Response.__init__`; post-fix the buffered Response round-trips `.text` cleanly and no `content-encoding` survives onto it.
- Existing transport / size-cap tests still pass (36/36 in `test_core_transport.py`).

## Test plan

- [x] `uv run ruff format` clean
- [x] `uv run ruff check` clean
- [x] `uv run mypy src/notebooklm --ignore-missing-imports` clean
- [x] `uv run pytest tests/unit` clean ignoring the playwright-gated session/windows suites (4211 passed, 5 skipped)
- [ ] CI: full suite

@claude review

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented double-decoding and length mismatches for compressed streaming responses by removing misleading encoding/length headers from rebuilt responses, improving content delivery reliability.
* **Tests**
  * Added a unit test to verify streaming responses decode correctly and that misleading encoding/length headers are stripped.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/771?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/notebooklm/_core_transport.py
+++ b/src/notebooklm/_core_transport.py
@@ -29,6 +29,15 @@ MAX_RETRY_AFTER_SECONDS = 300
 # read loop can read it without creating an import cycle through ``_core``.
 MAX_RPC_RESPONSE_BYTES = 50 * 1024 * 1024
 
+# Headers that must NOT survive onto a Response rebuilt from already-decoded
+# body bytes. ``content-encoding`` would make ``httpx.Response.__init__``
+# re-run the gzip/brotli/zstd decoder on bytes that ``aiter_bytes()`` already
+# decoded once, raising ``DecodingError: Error -3 ... incorrect header check``.
+# ``content-length`` advertises the compressed size from the wire and no
+# longer matches the decoded buffer we hand to the rebuilt Response. Compared
+# against ``key.lower()`` so case variants from the wire all match.
+_STRIP_HEADERS_ON_REBUFFER = frozenset({"content-encoding", "content-length"})
+
 
 def _parse_retry_after(value: str | None) -> int | None:
     """Parse RFC 7231 Retry-After: integer-seconds OR HTTP-date.
@@ -177,14 +186,17 @@ async def _stream_post_with_size_cap(
         # over so log/repr surfaces still point at the originating request.
         #
         # ``response.aiter_bytes()`` above yields already-decoded body chunks,
-        # so the buffered payload is plain bytes. Strip ``content-encoding``
-        # (and the now-mismatched ``content-length``) before handing the
-        # headers to a fresh ``httpx.Response`` — otherwise its ``__init__``
-        # re-runs the gzip/brotli/zstd decoder on already-decoded bytes and
-        # raises ``DecodingError: Error -3 ... incorrect header check``.
-        rebuilt_headers = httpx.Headers(response.headers)
-        rebuilt_headers.pop("content-encoding", None)
-        rebuilt_headers.pop("content-length", None)
+        # so the buffered payload is plain bytes. Filter out
+        # ``content-encoding`` (and the now-mismatched ``content-length``) via
+        # a dict comprehension — ``httpx.Headers`` inherits from
+        # :class:`collections.abc.Mapping`, NOT ``MutableMapping``, so we
+        # avoid relying on ``.pop()`` (which is not part of the documented
+        # contract and could change across the ``>=0.27,<0.29`` httpx pin).
+        # ``httpx.Response(headers=...)`` accepts a plain ``dict`` of
+        # ``str -> str`` so this is the documented input shape.
+        rebuilt_headers = {
+            k: v for k, v in response.headers.items() if k.lower() not in _STRIP_HEADERS_ON_REBUFFER
+        }
         return httpx.Response(
             status_code=response.status_code,
             headers=rebuilt_headers,
--- a/src/notebooklm/_core_transport.py
+++ b/src/notebooklm/_core_transport.py
@@ -175,9 +175,19 @@ async def _stream_post_with_size_cap(
         # (``_core_rpc.py`` decode path) can use ``.text`` / ``.content``
         # without dealing with stream state. The request handle is carried
         # over so log/repr surfaces still point at the originating request.
+        #
+        # ``response.aiter_bytes()`` above yields already-decoded body chunks,
+        # so the buffered payload is plain bytes. Strip ``content-encoding``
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review...

teng-lin:
Thanks @claude — applied the dict-comprehension form in 0eebdbb.

Replaced the ``httpx.Headers.pop()`` calls with a comprehension over ``response.headers.items()`` keyed off a module-level ``frozenset({"content-encoding", "content-length"})``. Same observable behavior, but doesn't rely on ``httpx.Headers`` exposing ``MutableMapping``-style methods that aren't part of the abstract ``Mapping`` contract our ``httpx >=0.27,<0.29`` pin is allowed to depend on. ``httpx.Response(headers=dict[str, str])`` is the documented input shape.

Skipping the ``_FakeResponse.headers`` parametrize (production-type observation) — the test stub flows through the production code path which normalizes whatever it receives, and you flagged it as a nit. Will revisit if it surfaces a real bug later.

PR #777, opened 2026-05-17
test(vcr): gzipped cassette coverage to catch #769-class regressions

### LINKED ISSUE #773: test(vcr): cassettes drop Content-Encoding, blinding VCR to RPC-decode regressions (#769 follow-up)
## Context

Issue #769 was a real production bug — every RPC failed with `Error -3 while decompressing data: incorrect header check` — that **slipped through the entire VCR-based integration suite + all 4351 unit tests**. The fix lives in #771. This issue tracks closing the test-coverage gap that let the regression land.

## Root cause of the blind spot

The bug lived in `_core_transport._stream_post_with_size_cap`: it reads chunks via `response.aiter_bytes()` (which yields **already-decoded** body bytes), then rebuilds a fresh `httpx.Response(content=buffer, headers=response.headers)`. When the upstream advertised `Content-Encoding: gzip` (the default for `batchexecute`), the rebuilt Response re-ran the gzip decoder on already-decoded bytes inside its `__init__` and raised `DecodingError`.

For this to surface in a test, the test fixture must:
1. Serve a response whose headers carry `Content-Encoding: gzip` (or `br` / `zstd`), AND
2. Flow through the streaming + Response-rebuild path that #751 introduced.

Neither condition is satisfied by the current cassette suite or unit-level fakes:

- **89 cassettes in `tests/cassettes/*.yaml`, zero carry a `content-encoding` header.** Grep is empty:
  ```
  grep -l 'content-encoding\|Content-Encoding' tests/cassettes/*.yaml
  # (no output)
  ```
  VCR-py captures bodies *after* httpx's content decoders have already run, so by the time the cassette is serialized the encoding header has been stripped or never made it in. On replay, the rebuilt-Response path sees no `content-encoding`, the double-decode branch never fires, and the test passes.

- **Unit fakes in `tests/unit/test_core_transport.py` use bare or `text/plain` headers** on their `_FakeResponse` stubs (e.g. `test_normal_response_below_cap_works`, `test_streamed_response_size_cap`). None of the size-cap test family advertises `content-encoding`, so the regression was invisible at the unit layer too.

- **Pre-#751 the path didn't exist.** `client.post` returns a single Response that decodes exactly once, so even cassettes that *did* preserve `Content-Encoding` couldn't have flagged this combination — there was no historical cassette where the rebuild step was exercised.

### 1. Re-record (or hand-edit) at least one cassette per RPC family with `content-encoding: gzip`

We don't need all 89 cassettes carrying the header — one representative cassette per RPC family (notebooks list, sources add, chat ask, artifacts generate, research start, settings, sharing) is enough to exercise the rebuild path with realistic headers.

Options:
- Re-record against live `batchexecute` with VCR configured to preserve the `Content-Encoding` response header (default behavior drops it after decode).
- Or: a one-shot script that takes existing cassettes, adds `content-encoding: gzip` to the response headers, and gzip-compresses the recorded body so the cassette is internally consistent.

### 2. Cassette-shape lint: assert non-empty coverage of common encoding headers

PR #556 already added a `tests/vcr_cassette_shape_lint` style guard. Extend it with an assertion like:

> At least one cassette in the suite must carry `Content-Encoding: gzip`. Optionally also assert at least one carries `br` or `zstd` if those decoders are installed.

This is the same idea as the existing shape lint, scoped to encoding headers. It would have failed loudly the moment #751 landed because no cassette covered the new code path.

### 3. Unit-level: encoding-header parametrization on the streaming wrapper

`test_normal_response_below_cap_works` already exists for the no-encoding happy path. #771 added `test_streaming_strips_content_encoding_to_prevent_double_decode` for the gzip case. Worth parametrizing over `gzip` / `br` / `zstd` / `deflate` so we're covered against any future content-decoder behavior change in httpx.

### 4. (Stretch) Property check on the rebuild step itself

The invariant is: *the rebuilt Response must round-trip `.text` / `.content` exactly for any (decoded_body, upstream_headers) pair, regardless of what content-encoding the headers claim*. A Hypothesis property check over `(body, encoding ∈ {gzip, br, zstd, deflate, None})` would catch any future regression where someone copies the wrong subset of headers across.

## Why this matters beyond #769

Three of the recent transport-layer fixes (#751, #744, #549-style streaming) touch the boundary between httpx's content decoders and our own response handling. If cassettes faithfully represented production encoding headers we'd have a much earlier signal on this class of bug. Fixing the cassette shape once is a cheap durable win across that whole code surface.

## Acceptance criteria

- [ ] At least one cassette per major RPC family carries `Content-Encoding: gzip` and is internally consistent on replay.
- [ ] `tests/vcr_cassette_shape_lint` (or its analog) asserts the encoding-header coverage and would fail if all cassettes drop it again.
- [ ] `test_streaming_*` parametrized over the common content-encoding values.
- [ ] A run of `git revert <#771-fix-commit>` against the current test suite **fails** (i.e. the suite would have caught #769).

teng-lin:
## Summary
Closes #773. Adds VCR coverage for the gzip-rebuild path so #769-class double-decode regressions can't silently land again.

- `_stream_post_with_size_cap` rebuilds an `httpx.Response` from already-decoded body bytes. Pre-#771 it carried the upstream `Content-Encoding: gzip` over, and `Response.__init__` re-ran the decoder → `DecodingError`.
- The cassette suite was blind to this because `decode_compressed_response=True` strips the header on record. 89/89 cassettes had no encoding header, so the rebuild branch never fired on replay.
- This PR ports one cassette (`artifacts_revise_slide`) to the production wire shape (gzipped body + `Content-Encoding: gzip`), replays it through `_stream_post_with_size_cap`, and pins a floor on encoding-header coverage in the shape lint.

### Scope note
The issue acceptance criteria call for "one cassette per major RPC family" (7 cassettes). After discussion, scope was reduced to a single cassette: the rebuild path is shared across every RPC family, so one cassette plus the lint floor catches the same regression class with ~85% less artifact churn.

## Files
- `tests/scripts/inject_gzip_into_cassette.py` — idempotent gzip-injection helper. Normalizes the gzip OS byte (offset 9) to `0xff` so the artifact is byte-identical across Python 3.10–3.14. Uses `CSafeLoader`/`SafeLoader` (review feedback).
- `tests/cassettes/gzip_coverage/artifacts_revise_slide_gzipped.yaml` — derived cassette. Generated artifact; regenerate via the helper if the source recording changes.
- `tests/integration/test_gzip_cassette_replay.py` — replays the gzipped cassette through `_stream_post_with_size_cap`. Uses a scoped `vcr.VCR(decode_compressed_response=False)` because vcrpy applies the decode filter at cassette LOAD time, not just on record (see `vcr.cassette.Cassette._load` → `self.append` → `before_record_response` chain).
- `tests/unit/test_core_transport.py` — parametrizes the existing `test_streaming_strips_content_encoding_to_prevent_double_decode` over `gzip` / `br` / `zstd` / `deflate`. `br` / `zstd` use `pytest.importorskip` so they don't silently pass when the optional decoder package isn't installed.
- `tests/unit/test_cassette_shapes.py` — adds two new asserts: ≥1 cassette must advertise `Content-Encoding: gzip`, and every cassette under `gzip_coverage/` must be a fixed point of the helper.

## Acceptance criteria
- [x] ≥1 cassette carries `Content-Encoding: gzip` and is internally consistent on replay.
- [x] Cassette shape lint asserts encoding-header coverage and would fail if all cassettes drop it.
- [x] `test_streaming_*` parametrized over gzip / br / zstd / deflate (br / zstd skip honestly when decoder unavailable).
- [x] `git revert <#771-fix-commit>` against the suite fails. Manually verified by monkey-patching `_stream_post_with_size_cap` to the pre-#771 form and replaying the gzipped cassette — produced `httpx.DecodingError: Error -3 while decompressing data: incorrect header check`, the exact #769 error.

## Test plan
- [x] `uv run ruff format --check .` clean.
- [x] `uv run ruff check .` clean.
- [x] `uv run mypy src/notebooklm --ignore-missing-imports` 0 issues.
- [x] `uv run pytest` — 5022 passed, 14 skipped (the 2 new br/zstd skips reflect dev-env lacking those optional decoders).
- [x] Manually verified revert-of-#771 reproduces #769 error against the new integration test.

## Deferred polish findings (cosmetic only)
- Hard-coded URL / body in the integration test mirror the source cassette. If `artifacts_revise_slide.yaml` is ever re-recorded, regenerate the gzipped sibling via the helper (already documented in the test module docstring).
- `if raw is None: return False` in the helper is defensive — unreachable for real batchexecute responses.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive integration and unit tests ensuring gzipped HTTP responses replay correctly and preventing double-decode regressions; extended transport tests to cover gzip, brotli, zstd and deflate.
  * Added a canned gzipped cassette to exercise replay scenarios and a suite that verifies cassette round-trip stability.
* **Tools**
  * Added a utility script to inject deterministic gzip encoding into test cassettes for coverage and regeneration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/777?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/cassettes/gzip_coverage/artifacts_revise_slide_gzipped.yaml
+++ b/tests/cassettes/gzip_coverage/artifacts_revise_slide_gzipped.yaml
@@ -1,0 +1,92 @@
+interactions:
+- request:
+    body: f.req=%5B%5B%5B%22KmcKPe%22%2C%22%5B%5B2%5D%2C%5C%22e52e81c5-613b-43ec-8f45-66a120644a26%5C%22%2C%5B%5B%5B0%2C%5C%22Move%20the%20title%20up%5C%22%5D%5D%5D%5D%22%2Cnull%2C%22generic%22%5D%5D%5D&at=SCRUBBED_CSRF%3A1778848153629&
+    headers:
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      Content-Length:
+      - '241'
+      Content-Type:
+      - application/x-www-form-urlencoded;charset=UTF-8
+      Cookie:
+      - NID=SCRUBBED; SID=SCRUBBED; __Secure-1PSID=SCRUBBED; __Secure-3PSID=SCRUBBED;
+        HSID=SCRUBBED; SSID=SCRUBBED; APISID=SCRUBBED; SAPISID=SCRUBBED; __Secure-1PAPISID=SCRUBBED;
+        __Secure-3PAPISID=SCRUBBED; __Secure-1PSIDTS=SCRUBBED; __Secure-3PSIDTS=SCRUBBED;
+        SIDCC=SCRUBBED; __Secure-1PSIDCC=SCRUBBED; __Secure-3PSIDCC=SCRUBBED; OSID=SCRUBBED;
+        __Secure-OSID=SCRUBBED; _gcl_au=1.1.1583381276.1778846987; _ga=GA1.1.1567240762.1778846987;
+        _ga_W0LDH41ZCB=GS2.1.s1778846986$o1$g0$t1778846986$j60$l0$h0
+      Host:
+      - notebooklm.google.com
+      User-Agent:
+      - python-httpx/0.28.1
+    method: POST
+    uri: https://notebooklm.google.com/_/LabsTailwindUi/data/batchexecute?rpcids=KmcKPe&source-path=%2Fnotebook%2Fbb00c9e3-656c-4fd2-b890-2b71e1cf3814&f.sid=SCRUBBED&hl=en&rt=c
+  response:
+    body:
+      string: !!binary |
+        H4sIAAAAAAAC/51RS0/DMAy+91dUucCkZMrDCelvGEOIa5pDHs5W0a1T1cGJ/046LgjEhYOfnz/b
+        sjf+465pQMjGOfI+x22ZCSW7U9o9Y3Wc60m0kACNYQkfIgMVLQvJGmZ5siFnVYJUPaE9eZoWjNP0
+        +rhvXzBjGc6Y23u5qaClbm0FxsQOUbEkEjLQWJg1IjGRIZaUO9WF3BPvPRX0fB3HfyjnbqYnZa5z
+        BRX+jzLHa9F+esN2OVYZlhHb6+U2/Relro5aohVJMyPUegVMzBaooQlCcgMQpLmRyU8yOeAZ5yER
+        Tx3JQ11Ka7n6oWyPy3KZT18pSowWwI21nCvggnMjZIW4941U63/qR+BbY9Ar9Al+l1NiwgEAAA==
+    headers:
+      Accept-CH:
+      - Sec-CH-UA-Arch, Sec-CH-UA-Bitness, Sec-CH-UA-Full-Version, Sec-CH-UA-Full-Version-List,
+        Sec-CH-UA-Model, Sec-CH-UA-WoW64, Sec-CH-UA-Form-Factors, Sec-CH-UA-Platform,
+        Sec-CH-UA-Platform-Version
+      Alt-Svc:
+      - h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
+      Cache-Control:
+      - no-cache, no-store, max-age=0, must-revalidate
+      Content-Disposition:
+      - attachment; filename="response.bin"; filename*=UTF-8''response.bin
+      Content-Encoding:
+      - gzip
+      Content-Security-Policy:
+      - require-trusted-types-for 'script';report-uri /_/LabsTailwindUi/cspreport
+      Content-Type:
+      - application/json; charset=utf-8
+      Cross-Origin-Opener-Policy:
+      - same-origin-allow-popups
+      Cross-Origin-Resource-Policy:
+      - same-site
+      Date:
+      - Fri, 15 May 2026 12:29:13 GMT
+      Expires:
+      - Mon, 01 Jan 1990 00:00:00 GMT
+      Permissions-Policy:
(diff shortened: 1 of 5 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the review @claude. All four moderate/minor findings addressed in 1ec6c07, plus a Windows CI failure fix:

**Moderate**
- ✅ `_inject_into_response` docstring (`tests/scripts/inject_gzip_into_cassette.py`) — corrected the Returns clause: `True` covers both first-time and idempotent-rewrite paths; `False` is only when the body has no string to process.
- ✅ `pytestmark = pytest.mark.vcr` (`tests/integration/test_gzip_cassette_replay.py`) — kept it (it's required by the integration tier-enforcement rule in `tests/integration/conftest.py:152-182` since we use a scoped `vcr.VCR(...)` rather than `notebooklm_vcr.use_cassette`), but added an inline comment explaining why so the next reader doesn't think it's copy-paste.

**Minor**
- ✅ `CDumper` → `CSafeDumper` in both files for loader/dumper symmetry. Verified `bytes` still round-trips as `!!binary` under `SafeRepresenter`, and the gzipped cassette regenerates byte-identical.
- ✅ Added a comment above `match_on=...` explaining the single-interaction assumption and what to switch to (`rpcids`) if a second batchexecute interaction is added.

**Bonus (Windows CI)**
- The first push failed on `windows-latest, Python 3.11/3.12` because those Windows runners ship without libyaml — the new "walk every cassette" test was hitting pure-Python `yaml.safe_load`, ~10× slower, and timing out at 60s. Replaced the structural walk with a precompiled regex (`_CONTENT_ENCODING_GZIP_RE`) against the raw cassette text. The test now runs in <200ms across all platforms while preserving the same "≥1 cassette advertises Content-Encoding: gzip" floor.

Full suite: 5022 passed, 14 skipped locally; pushing now to re-run CI.

PR #792, opened 2026-05-17
docs(claude): update Layered Design + Repository Structure for seam modules (D1 partial)

teng-lin:
## Summary

Addresses the **deferred finding** from PR #790 claude bot review: CLAUDE.md's Layered Design diagram + Repository Structure tree still showed single \`_core.py\` and didn't reflect the seam-module split that landed in Phase 1 (PRs #778, #785, #786) + B1 (PR #789).

This is the **claude-cleanup subset of original Phase 4 D1**, dispatched in parallel with the in-flight B2 / D1-conformance-test / D1-audit-full.

## Changes

- **Layered Design diagram**: \`Core Layer (_core.py)\` → \`Core Layer (_core.py + _core_*.py seam modules)\`
- **Core Layer narrative**: per-module bullet list (transport, rpc, auth, cookie persistence, metrics, drain, reqid, cache, polling, capabilities)
- **Repository Structure tree**: add the existing \`_core_*.py\` modules

\`_core_lifecycle.py\` is hedged as \"(Phase 2 in progress)\" since B2 may merge concurrently.

## Test plan

- [x] \`uv run ruff format --check\` — 362 files unchanged
- [x] \`uv run ruff check\` — all checks pass
- [x] \`uv run mypy src/notebooklm\` — 114 files, no issues
- [ ] CI: full pytest matrix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated architecture documentation detailing core module decomposition and repository structure. These updates provide clarity on the current system organization and component hierarchy.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/792?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #790, opened 2026-05-17
docs(core): pre-stage decomposition module map + architecture text (D1 partial)

teng-lin:
## Summary

Pre-stages the predictable docs portion of Phase 4 D1 (the core-decomposition mini-phase's audit + cleanup wave). Adds a comprehensive module-map section to `docs/python-api.md`, upgrades the Layer Responsibilities table in `docs/development.md` with a per-module Core-layer breakdown, and replaces the single `_core.py` row in `CLAUDE.md`'s Key Files table with a per-module list.

## Why now (parallel-dispatch context)

This is part of an aggressive parallel-launch strategy: while B1 (auth coordinator) and C1a (rpc_call body collapse) are in flight, this docs PR lands independently because it touches no `src/` code.

The compat-property audit + dead-shim removal (the audit subset of D1) **defers** to D1-full after Phase 3 merges, since that work needs final post-merge LOC numbers.

## Out of scope

- Compat-property audit + dead-shim removal → D1-full after Phase 3
- Module-level import-surface smoke test → D1-full
- \`docs/configuration.md:307\` stale line numbers → D1-full (LOC will shift again with B2/C1)
- \`docs/auth-keepalive.md:531\` ClientCore→ClientLifecycle reference → D1-full (waits for B2 merged)

## Hedging

\`_core_auth.py\` and \`_core_lifecycle.py\` are marked "Phase 2 in progress" in all three updated docs since these PRs may merge concurrently with this docs PR. If they land first, the "(Phase 2 in progress)" hedge becomes outdated but not factually wrong — a follow-up doc cleanup in D1-full removes the hedge.

## Test plan

- [x] \`uv run ruff format --check\` — 360 files left unchanged
- [x] \`uv run ruff check\` — all checks pass
- [x] \`uv run mypy src/notebooklm\` — 113 files, no issues
- [x] \`uv run pytest tests/unit/test_public_shims.py tests/unit/test_init_order.py\` — 281 pass
- [ ] CI: full pytest matrix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Updated developer and architecture documentation to provide comprehensive details about internal module organization and component responsibilities.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/790?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #794, opened 2026-05-17
test(capabilities): add ClientCoreCapabilities Protocol conformance test (D1 partial)

teng-lin:
## Summary

Adds the **conformance-test subset of original Phase 4 D1**. Structural-conformance tests for the 9 base Protocols in \`_capabilities.py\` (the public capability surface that sub-clients consume).

This is part of the parallel split of D1 into independent shippable subsets: D1-docs (#790, merged), D1-claude-cleanup (#792, in review), D1-conformance-test (this PR), D1-audit-full (deferred until B2 + this + #792 land).

## Test plan

- [x] \`uv run ruff format --check\` — 363 files unchanged
- [x] \`uv run ruff check\` — all checks pass
- [x] \`uv run mypy src/notebooklm\` — 114 files, no issues
- [x] \`uv run pytest tests/unit/test_capabilities_conformance.py\` — 22 tests pass
- [ ] CI: full pytest matrix

## Key invariants pinned

- ClientCore and ClientCoreCapabilities both implement every member of all 9 Protocols
- \`ClientCore.__new__(ClientCore)\` fixture — zero runtime side effects (no \`open()\`, no network)
- \`len(*Provider classes) == 9\` — catches future Protocol drift loudly

### DIFF
--- a/tests/unit/test_capabilities_conformance.py
+++ b/tests/unit/test_capabilities_conformance.py
@@ -1,0 +1,259 @@
+"""Structural conformance tests for the 9 base Protocols in ``_capabilities.py``.
+
+These tests are pure structural checks: they assert that
+``ClientCoreCapabilities`` (the adapter) — and, where applicable,
+``ClientCore`` (the underlying concrete class) — expose every member
+that each base Protocol declares. They never open the client, never
+touch the network, and never instantiate via ``__init__``.
+
+The ``__new__``-only ``ClientCore`` fixture is the explicit contract:
+the test must run with zero runtime side-effects, so we bypass
+``ClientCore.__init__`` entirely.
+
+This test is intentionally noisy when a Protocol is added or removed —
+the 9-Protocol count guard at the top forces future contributors to
+update ``_BASE_PROTOCOLS`` in lockstep.
+"""
+
+from __future__ import annotations
+
+import ast
+import inspect
+from pathlib import Path
+from typing import Protocol
+
+import pytest
+
+from notebooklm._capabilities import (
+    AuthRouteProvider,
+    ChatStreamingProvider,
+    ClientCoreCapabilities,
+    CookieJarProvider,
+    CoreReqIdProvider,
+    CoreRPCProvider,
+    PollRegistryProvider,
+    SourceListProvider,
+    TransportOperationProvider,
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the review! Addressed in commit 7e23577. Two of your findings turned into stronger checks than the original sketch:

**1. Missing `__bases__` == `_BASE_PROTOCOLS` guard (recommended fix)** — FIXED. Added `test_client_core_capabilities_inherits_all_provider_protocols` that asserts `set(ClientCoreCapabilities.__bases__) == set(_BASE_PROTOCOLS)`. Belt-and-suspenders with the AST count guard so a duck-typed-but-not-MRO'd 10th Protocol can't slip through.

**2. Redundant `__annotations__` loops at lines 170–173 / 201–205** — Kept intentionally per the spec recipe but tightened. The loops are now part of a stronger `name in caps_cls.__dict__` check (rather than `hasattr`), so they catch a real failure mode that the surrounding `missing` list does not: the inherited Protocol stub fall-through. (Codex flagged the same `hasattr` weakness as a CRITICAL during /momus triple-review, and the `__dict__` switch was the documented fix.)

**3. `test_protocol_member_enumeration_finds_known_members` error message** — Not changed in this round; the failure renders a list diff which is unambiguous in pytest output. Will revisit if it actually trips someone.

**4. `inspect.getmembers` duplication** — Documented as intentional belt-and-braces; left as-is.

Additional polish hardening on top of your review (from codex CRITICAL/IMPORTANT + ultrathink):

- `_CORE_NATIVE_MEMBERS` is now round-trip-checked via `test_core_native_members_allowlist_is_symmetric` (catches drift in either direction).
- AST scan restricted to `tree.body` top-level classes that directly inherit `Protocol`; AST-discovered names cross-checked against the imported `_BASE_PROTOCOLS` tuple.
- `test_core_fixture_skips_init` strengthened from `"poll_registry" not in vars(core)` to `vars(core) == {}`.
- `_PROTOCOL_BASE_DIR` hoisted to module-level constant (de-duplicates 2 call sites).

Final state: 24 tests, all green; full pytest 5125/5125 passing; ruff/mypy clean.

Issue #474, opened 2026-05-14
source add fails on 0.4.1: "Failed to get SOURCE_ID from registration response" despite RPC   succeeding


DrCarlosCalderon:
## Description

  Since upgrading to `notebooklm-py 0.4.1`, every `notebooklm source add` invocation fails with `Failed to
  get SOURCE_ID from registration response`. The verbose log shows the RPC (`ADD_SOURCE_FILE`, method
  `o4cbdc`) completes successfully with a parsed response, but the SOURCE_ID extraction step fails. This 
  breaks any sync workflow that depends on the CLI to track uploaded sources by ID — uploads either silently
   succeed (and become untracked) or silently fail.

  ## Steps to Reproduce

  1. `pip3 install --upgrade notebooklm-py` (lands on 0.4.1)
  2. `notebooklm doctor` — passes (auth/profile/migration all green)
  3. `notebooklm -vv source add /path/to/any.pdf --notebook <notebook-id>`

  Reproduces consistently across multiple distinct PDFs, multiple folders, and the same target notebook ID.

  ## Expected Behavior

  `source add` returns the newly-created source's ID (or a clean error if the upload was rejected 
  server-side), allowing the CLI to track the upload in any manifest layer downstream.

  ## Actual Behavior

  The CLI exits with:

  ```text
  Error: Failed to get SOURCE_ID from registration response

  Despite the RPC layer reporting success. Verbose log under "Debug Output" below.

  Environment

  - OS: macOS 26.4.1 (Build 25E253, Darwin 25.4.0)
  - Python version: 3.12.10 (miniforge3)
  - notebooklm-py version: 0.4.1
  - Install method: pip3
  - Auth status: notebooklm doctor — all checks pass (SID cookie present, 30 cookies, migration complete,
  default profile)

  Debug Output

  notebooklm -vv source add /path/to/file.pdf --notebook <id>

  Relevant tail:

  13:35:26 DEBUG [notebooklm.cli.helpers] CLI command starting: source_add
  13:35:26 DEBUG [notebooklm.auth] Extracted 30 cookies from domains: .google.com, .notebooklm.google.com,
  .workspace.google.com, accounts.google.com, notebooklm.google.com, ogs.google.com
  13:35:26 DEBUG [notebooklm.auth] SID cookie from domain: .google.com
  13:35:26 DEBUG [notebooklm.auth] Fetching CSRF and session tokens from NotebookLM
[...]

PR #817, opened 2026-05-18
fix(sources): address coderabbit review on PR #813 — baseline-unavailable sentinel + probe-before-fail

teng-lin:
## Summary

Follow-up to PR #813 (b-sources). PR #813 merged at \`01849df\` before the CodeRabbit review fully posted; this PR ships the two CR critical/major findings that landed after the merge.

## Changes

1. **Baseline-unavailable sentinel** (\`_source_upload.py:register_file_source\`). The previous \`baseline_ids = set()\` on baseline-fetch failure made the probe maximally permissive — a single pre-existing same-name source could be silently returned as if it were the just-created upload, directing the subsequent streaming POST onto the wrong source. Replaced with \`baseline_ids: set[str] | None = None\` sentinel; the probe raises \`SourceAddError\` ("baseline snapshot was unavailable") when any same-titled match is found in this state rather than guessing.

2. **Probe-before-fail on source-less success** (\`_source_upload.py:register_file_source._create\`). If the ADD_SOURCE_FILE RPC returns 200 but \`_extract_register_file_source_id\` can't find a SOURCE_ID (e.g. schema drift per #474), the previous code raised \`SourceAddError\` immediately. Now \`_create\` calls \`_probe()\` first — the create may have actually committed server-side and the probe can recover the new source's id. Only raises if the probe also fails.

## Test plan

New test \`test_register_file_source_baseline_unavailable_raises_on_ambiguity\` covers the sentinel path. Existing \`TestRegisterFileSourceError\` tests updated for the extra probe call now made before raising on source-less success responses.

All checks green:
- \`ruff format\` + \`ruff check\` clean
- \`mypy src/notebooklm --ignore-missing-imports\` clean
- \`pytest tests/unit tests/integration\` — 5369 passed / 20 skipped

## Context: why a follow-up?

PR #813 was merged at \`01849df\` before CodeRabbit's review fully posted. The CR review included three findings:
- **Critical (Drive URL boundary)**: already addressed in \`01849df\` via the Claude-bot fix (same issue both bots flagged).
- **Major (baseline-unavailable sentinel)**: shipped here.
- **Major (probe-before-fail)**: shipped here.

The CR review threads on PR #813 have replies pointing at this follow-up's SHA.

## Release Notes

* **Bug Fixes**
  * Enhanced file source registration to handle unavailable baseline snapshots without misidentifying existing sources.
  * Improved error handling for source ID extraction failures with better fallback recovery.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/817?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/notebooklm/_source_upload.py
+++ b/src/notebooklm/_source_upload.py
@@ -348,22 +348,71 @@ class SourceUploadPipeline:
 
         # Capture baseline source IDs before the first create attempt so the
         # probe can distinguish "this upload landed" from "a same-named source
-        # already existed." Mirrors the pattern in NotebooksAPI.create. A
-        # transport failure during the baseline fetch falls back to an empty
-        # set, which makes the probe maximally permissive (any same-named
-        # source is treated as a potential match, including pre-existing
-        # ones). This is a doubly-exceptional scenario — baseline list failure
-        # AND a same-name collision — and is accepted as a known limitation
-        # mirroring the parallel note in NotebooksAPI.create. The baseline
-        # fetch is best-effort and never blocks the primary create path.
+        # already existed." Mirrors the pattern in NotebooksAPI.create.
+        #
+        # ``None`` is the "baseline unavailable" sentinel — used when the
+        # baseline fetch failed (e.g. transient 5xx). The probe treats this
+        # as "we cannot safely distinguish new sources from pre-existing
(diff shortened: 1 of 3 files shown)


teng-lin:
@claude review

Issue #659, opened 2026-05-16
CLI conversations invisible in web UI: client-side uuid4() conversation_id not recognized by server


XanderChen0530:
## Summary

CLI-created conversations do not appear in the NotebookLM web UI. The root cause is that the CLI generates `conversation_id` client-side via `uuid.uuid4()` (in `_chat.py` line ~100), which the NLM server does not register in the web UI's conversation list.

## Steps to Reproduce

1. `notebooklm use <notebook_id>`
2. `notebooklm ask "any question"` (without `-c` flag)
3. Open NotebookLM web UI → conversation does not appear
4. The answer is returned successfully, but the conversation is invisible on the web

## Root Cause Analysis

In `src/notebooklm/_chat.py`:
```python
is_new_conversation = conversation_id is None
if is_new_conversation:
    conversation_id = str(uuid.uuid4())  # Client-side UUID — server doesn't recognize this
```

The NLM web UI assigns conversation_ids server-side. When the CLI generates its own UUID, the server accepts the question but doesn't associate it with the web UI's conversation list.

## Workaround

Obtain the server-side conversation_id first via `hPTbtc` RPC, then use `-c`:

```python
import asyncio
from notebooklm import NotebookLMClient

async def get_server_conv_id(notebook_id):
    async with await NotebookLMClient.from_storage() as client:
        return await client.chat.get_conversation_id(notebook_id)

server_conv_id = asyncio.run(get_server_conv_id("your-notebook-id"))
```

Then:
```bash
notebooklm ask --json -c "$server_conv_id" "your question"
```

This makes the conversation appear in the web UI with working `[N]` hover links.

## Proposed Fix

When `conversation_id is None` (new conversation), instead of generating a client-side UUID, the CLI should either:

1. **Let the server assign the conversation_id** by sending the request without one and extracting the server-assigned ID from the response, or
2. **Create a conversation via a dedicated server RPC** before sending the first question

This would make CLI conversations natively visible in the web UI without the `-c` workaround.

## Environment

- notebooklm-py version: 0.4.1 (also confirmed on 0.3.2)
- Platform: WSL2 Ubuntu, Python 3.12

teng-lin:
## Live API investigation — option 1 isn't quite right; here's the actual contract

[...]

PR #821, opened 2026-05-18
fix(ci): scope nightly retry to tests/e2e + drop broken new-conversation e2e

teng-lin:
## Summary

Fixes two issues uncovered while triaging a nightly E2E job that reported 84 failures: https://github.com/teng-lin/notebooklm-py/actions/runs/26019127504

- **`.github/workflows/nightly.yml` retry step** — was running `pytest --last-failed -s -v --tb=short` with no path. pyproject's `addopts = --ignore=tests/e2e` drops every e2e entry from collection, the cached e2e lastfailed entry can't match, and pytest's default `--last-failed-no-failures=all` then expands to the entire non-e2e suite under `NOTEBOOKLM_AUTH_JSON` — which causes ~80 spurious login unit-test failures (the CLI refuses to run with that env set). Pin to `tests/e2e` and use `=none` so missing/empty cache fails loudly instead of silently broadening.
- **`tests/e2e/test_chat.py::test_ask_new_conversation_flag`** — asserted two consecutive `chat.ask()` calls with no `conversation_id` return *different* conversation_ids, but `_chat.py:204-209` documents the opposite (all such calls extend the most-recent conversation). Per issue #659 the "fresh conversation" RPC has not been reverse-engineered. The test was asserting documented-impossible behavior — deleted.

## Evidence

Nightly run summary:
- First e2e step: `1 failed, 126 passed, 4 skipped, 25 deselected, 1 xpassed, 10 warnings, 1 rerun` (just e2e)
- Retry step: `84 failed, 5297 passed, 20 skipped` ← the whole non-e2e suite ran

Windows was green because it runs `-m "readonly and not variants"`, never failed, so the retry never fired and the scope bug stayed hidden.

## Follow-ups (out of scope for this PR)

- The CLI `--new` flag (`cli/chat.py:86-91`, help text "Start a fresh conversation") is genuinely misleading per the same `_chat.py:204-209` contract — `--new` just skips local-cache resume, the server still extends the most-recent conversation. Pre-existing; should be either a help-text fix or proper implementation pending issue #659.

## Test plan

- [ ] CI green on this branch (unit + integration + verify-package + rpc-health)
- [ ] Next nightly run exercises the new retry path naturally if any e2e fails
- [ ] If no nightly failure occurs, a `workflow_dispatch` with a known-failing `test_filter` will validate the retry scope

## Summary by CodeRabbit

* **Tests**
  * Removed an end-to-end test case related to conversation functionality.

* **Chores**
  * Updated the nightly CI workflow's end-to-end test retry mechanism to improve failure detection accuracy and provide faster feedback when tests encounter issues.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/821?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -217,4 +217,9 @@ jobs:
         fi
         echo "Initial E2E run failed. Sleeping 600s before retrying failed tests."
         sleep 600
-        uv run pytest --last-failed -s -v --tb=short
+        # Pin ``tests/e2e`` so pyproject's ``addopts = --ignore=tests/e2e``
+        # doesn't drop every lastfailed entry; otherwise pytest's default
+        # ``--last-failed-no-failures=all`` would expand to the full non-e2e
+        # suite under NOTEBOOKLM_AUTH_JSON. ``=none`` makes a missing cache
+        # fail fast instead.
+        uv run pytest tests/e2e --last-failed --last-failed-no-failures=none -s -v --tb=short
--- a/tests/e2e/test_chat.py
+++ b/tests/e2e/test_chat.py
@@ -95,26 +95,6 @@ class TestChatE2E:
         assert result2.turn_number > result1.turn_number
 
     @pytest.mark.asyncio
-    async def test_ask_new_conversation_flag(self, client, multi_source_notebook_id):
-        """Test that --new flag starts a fresh conversation."""
-        # Ask first question
-        result1 = await client.chat.ask(
-            multi_source_notebook_id,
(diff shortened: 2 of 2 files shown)


teng-lin:
@claude review

teng-lin:
Follow-up filed: #822 (reverse-engineer delete-conversation RPC to make the deleted test reincarnable and fix the misleading `--new` CLI flag promise).

PR #808, opened 2026-05-18
fix(research,notes): idempotent research start/import + CREATE_NOTE (P0-3)

teng-lin:
## Summary

Tier 9 Wave 2 task `b-research-notes` (P0-3-research + P0-3-notes). Classifies three RPC families in the research + notes domain in `IDEMPOTENCY_REGISTRY` so the transport's inner 5xx retry loop is forced off on commit-lost.

- `START_FAST_RESEARCH`, `START_DEEP_RESEARCH`, `IMPORT_RESEARCH` -> `NON_IDEMPOTENT_NO_RETRY`
- `CREATE_NOTE` -> `NON_IDEMPOTENT_NO_RETRY` (default + explicit `"plain"` and `"saved_from_chat"` operation variants)

## Policy Decision (per RPC)

Per the spec's `policy_decision_process`, the investigation step inspected the param shapes at the cited call sites:

| RPC | Param shape (call site) | Client-token slot? | Probe surface | Decision |
|---|---|---|---|---|
| `START_FAST_RESEARCH` | `[[query, source_type], None, 1, notebook_id]` (`_research.py:316`) | No | `ResearchAPI.poll` by `(notebook_id, query)` is ambiguous when multiple peer tasks exist | **NON_IDEMPOTENT_NO_RETRY** |
| `START_DEEP_RESEARCH` | `[None, [1], [query, source_type], 5, notebook_id]` (`_research.py:319`) | No | Same as fast-research; query is not a unique key | **NON_IDEMPOTENT_NO_RETRY** |
| `IMPORT_RESEARCH` | `[None, [1], task_id, notebook_id, source_array]` (`_research.py:615`) | No (task_id is correlation, not dedupe) | `SourcesAPI.list` post-commit-lost cannot bind URL-matched rows to "this specific import batch" — collides with prior workflows | **NON_IDEMPOTENT_NO_RETRY** |
| `CREATE_NOTE` (plain) | `[notebook_id, "", [1], None, title]` (`_mind_map.py:187`) | No | `GET_NOTES_AND_MIND_MAPS` returns no client-visible `note_id` post-commit-lost (CREATE_NOTE failed before returning the row id); content is empty (set by follow-up `UPDATE_NOTE`) | **NON_IDEMPOTENT_NO_RETRY** |
[...]

PR #823, opened 2026-05-18
ci(workflow): tolerate Playwright cache backend flakes

### LINKED ISSUE #815: ci: CI cache backend failure (Playwright) hard-fails matrix jobs instead of falling through to install
## Problem

The `Cache Playwright browsers` step in `.github/workflows/test.yml` uses `actions/cache@v5`. When the GitHub cache backend itself fails (transient infra flake, not a cache miss), the step exits non-zero, which marks the entire matrix job as failed and **skips every downstream step including the test run**.

We have seen this fail several times across recent Tier-9 PRs (#808 on Windows 3.13, #813 on Windows 3.13 — both within 29s into the job, i.e. before any test ran). The remediation is always "re-run the failed job," but the underlying issue is that we let an optional cache step gate test execution.

### Failure pattern (example from #813 / job 76453070311)

```
✅ Set up job
✅ Run actions/checkout@v6
✅ Set up Python 3.13
✅ Install uv
✅ Install dependencies
✅ Get Playwright version
❌ Cache Playwright browsers     <- transient backend failure
⏭️  Install Playwright browsers
⏭️  Install Playwright system dependencies (Linux)
⏭️  Assert cassettes are sanitized
⏭️  Run tests with coverage
⏭️  Assert per-file coverage floors
```

The cache step itself succeeded materially — Playwright would re-install on a miss — but a backend hiccup hard-failed the job.

## Proposed fix

Add \`continue-on-error: true\` to the \`Cache Playwright browsers\` step in \`.github/workflows/test.yml\` (around line 114):

\`\`\`yaml
- name: Cache Playwright browsers
  uses: actions/cache@v5
  continue-on-error: true
  with:
    path: \${{ github.workspace }}/.playwright-browsers
    key: playwright-\${{ matrix.os }}-\${{ steps.playwright-version.outputs.version }}-ws
\`\`\`

Rationale:
- A cache miss already falls through cleanly (next step always re-installs Playwright).
- A cache **backend** failure should be treated the same as a miss — continue, let the install step do the work.
- Cost: one extra browser download on the affected runner if the cache really was unreachable. Negligible vs. a hard-failed PR.

## Alternative

If the team wants a stricter gate, wrap the cache lookup + install in an explicit failure-tolerant block or pin a different cache action. But the simpler \`continue-on-error\` matches the existing tolerance philosophy (see the Linux system-deps step at line 129, which already swallows timeouts).

## Scope

Single workflow file edit; no source-code change. Ship as \`ci(workflow): tolerate Playwright cache backend flakes\` in its own PR.

## Repro

- Recent occurrences: PR #808 Windows-3.13, PR #813 Windows-3.13.
- 29-second job duration is the tell — real test runs take 5+ min.

teng-lin:
## Summary

- Add `continue-on-error: true` to the `Cache Playwright browsers` step in `.github/workflows/test.yml`.
- A transient `actions/cache@v5` backend failure currently hard-fails the entire matrix job and skips every downstream step — including the test run itself.
- A cache miss already falls through to the unconditional `Install Playwright browsers` step; a backend hiccup is now treated the same.
- Per GitHub Actions docs + `actions/cache@v5` action.yml, step-level `continue-on-error` covers both the restore (action main) and save (post-step) phases, so a flaky save half is also tolerated.

## Why

Two recent matrix jobs hard-failed on Windows × Python 3.13 ~29s into the run, before any test executed:

- PR #808 — Windows-3.13
- PR #813 — Windows-3.13 (job 76453070311)

Failure shape (from #813):

```
✅ Set up job
✅ Install dependencies
✅ Get Playwright version
❌ Cache Playwright browsers     <- transient backend failure (29s)
⏭️  Install Playwright browsers
⏭️  Run tests with coverage
⏭️  Assert per-file coverage floors
```

The cache step succeeded materially (the install step would have re-installed on a miss), but a backend hiccup gated everything downstream.

## Tradeoff

A permanently broken cache backend will go green — every matrix entry will silently re-install Playwright (~10–30s extra) and the only signal is a yellow annotation on the run. This matches the existing tolerance pattern on the Linux system-deps step (`.github/workflows/test.yml` lines 134–141, which emits `::warning::` on `playwright install-deps` timeout). The functional fallback is sound: `uv run playwright install chromium` runs unconditionally and validates the real requirement (browsers present before tests).

## Alternatives considered

- **Split `actions/cache/restore` + `actions/cache/save`** — more granular, but adds two steps and config surface for a transient-flake problem; still needs `continue-on-error` on each half.
- **`restore-keys`** — only helps cache misses, not backend failures.
- **`fail-on-cache-miss: true`** — opposite of the desired behavior.
- **Retry wrappers** — don't work cleanly around a `uses:` step.

The simple knob matches the issue scope.

## Polish notes (deferred, non-blocking)

The polish pipeline (code-simplifier + claude/codex/gemini triple review + ultrathink) returned **0 critical, 0 important**. Three nits were noted and deferred as informational:

- N1 (claude): comment frames the failure recovery as restore-phase only; `continue-on-error` also covers the save post-step (codex confirmed via docs).
- N2 (claude): no automated detection path for permanent cache backend breakage — surfaces only as a workflow annotation. Matches house style (Linux deps step has the same property).
- (codex): comment is incident-specific with "29s"; intentional intent-capture, codex opted to keep.

## Closes

Closes #815.

## Test plan

- [ ] CI green across `ubuntu-latest`, `macos-latest`, `windows-latest` × Python `3.10`, `3.11`, `3.12`, `3.13`, `3.14`.
- [ ] If a cache-backend flake recurs, it surfaces as a workflow annotation on the run and the `Install Playwright browsers` step re-installs cleanly without skipping downstream steps.
- [ ] No new failure modes introduced on the happy path (cache hit).

## Release Notes

* **Chores**
  * Improved CI test reliability by making the browser cache step more resilient to transient backend failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/823?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,6 +122,11 @@ jobs:
 
     - name: Cache Playwright browsers
       uses: actions/cache@v5
+      # Tolerate transient cache-backend failures: a miss already falls
+      # through to the install step below, and a backend hiccup should be
+      # treated the same. Without this, a 29s backend flake on a single
+      # matrix entry hard-fails the job and skips the entire test run.
+      continue-on-error: true
       with:
         path: ${{ github.workspace }}/.playwright-browsers
         # ``-ws`` suffix invalidates archives keyed to the prior dual-path


teng-lin:
@claude review

PR #826, opened 2026-05-18
fix(auth): gate refresh-gen bump on current-attempt inflight (#816)

### LINKED ISSUE #816: auth: spurious _REFRESH_GENERATIONS bump on early-cancel in _coalesced_run_refresh_cmd
## Summary

`_coalesced_run_refresh_cmd` in `src/notebooklm/_auth/refresh.py` (moved from `src/notebooklm/auth.py` in tier-10 PR-B-high #814, but the code is byte-identical to origin/main since tier-9 E #803) can bump `_REFRESH_GENERATIONS[refresh_key]` even when no real refresh happened, in a narrow cancel-before-await race.

## Reproducer (race condition)

1. Caller invokes ``await _coalesced_run_refresh_cmd(refresh_key, path, profile)`` from the cancel-recovery loop in ``_fetch_tokens_with_refresh``.
2. Task is cancelled in the window between ``try:`` and the first ``with _REFRESH_STATE_LOCK:`` block inside ``_coalesced_run_refresh_cmd`` — so NO future is registered in ``_REFRESH_INFLIGHT_BY_LOOP[loop][refresh_key]``.
3. Control jumps to ``except asyncio.CancelledError:`` in the caller.
4. ``registry.get(refresh_key)`` returns ``None``.
5. The inner ``if inflight is None or inflight.done():`` is taken; the ``if inflight is not None:`` block is skipped; ``subprocess_exc`` stays ``None``; ``break``.
6. Outer ``if subprocess_exc is not None:`` is False, so no raise.
7. **Generation is bumped** at the ``with _REFRESH_STATE_LOCK: ... max(existing, refresh_generation + 1)`` line, even though no subprocess ran.
8. Then ``if caller_cancelled: raise asyncio.CancelledError()``.

## Impact

A concurrent waiter on a different event loop sharing the same ``refresh_key`` would observe the phantom generation bump on its lock-re-acquire path, treat the storage as freshly-refreshed, and skip its own ``_run_refresh_cmd`` — proceeding to ``build_httpx_cookies_from_storage`` against STALE on-disk state.

Bounded blast radius: the stale-storage caller will fail again on the next token fetch and trigger its own real refresh. No data corruption. But it's a wasted round-trip.

## Suggested fix (from CodeRabbit on PR #814)

Track whether ``inflight`` was ever non-None during the cancel-recovery loop. Only bump generation if ``subprocess_exc is None AND observed_inflight is True``. Or per the alternative suggestion: raise immediately when ``inflight is None`` at the top of the cancel handler (the subprocess never registered, so we can't be the leader-cancel-before-await case for an actually-running refresh).

## Why not fixed in PR #814

That PR's scope was "extract verbatim from auth.py into _auth/refresh.py + facade-forward". Changing async semantics was explicitly out of scope (``must_not_do``). Filing here for a dedicated follow-up.

## References

- CodeRabbit comment: https://github.com/teng-lin/notebooklm-py/pull/814#discussion_r3256142201
- Original code: ``src/notebooklm/_auth/refresh.py:487-537`` (tier-10), ``src/notebooklm/auth.py:961-1026`` (tier-9 E origin)

teng-lin:
## Summary

Closes #816.

`_fetch_tokens_with_refresh` could phantom-bump `_REFRESH_GENERATIONS` in two narrow cancel races where no current-attempt subprocess ran:

1. **Cold registry** (issue #816 original): cancellation arrived before `_coalesced_run_refresh_cmd` registered its inflight future. The cancel handler saw `inflight is None`; the code fell through to the generation bump anyway.
2. **Warm registry** (surfaced by Codex during polish): `_settle` intentionally leaves done futures in the registry (PR #621 cancel/settle race retention). After a prior successful refresh, the registry still holds that done success future. If cancellation arrives before our `_coalesced_run_refresh_cmd` overwrites the slot, treating the leftover as proof of our attempt re-bumps the generation against an outdated result.

In both cases a concurrent waiter on a sibling event loop would observe the phantom bump on its lock-re-acquire path, treat storage as freshly refreshed, skip its own `_run_refresh_cmd`, and proceed with STALE on-disk state.

## Fix

Capture `prior_inflight` BEFORE the await and `prior_was_active` (was it not-done at capture). In the cancel handler, the registry slot counts as the current refresh attempt only when either:
- it was overwritten during our await (`inflight is not prior_inflight` — a new leader inserted a fresh future), OR
- it already held an actively-running sibling leader at capture (`prior_was_active`).

A stale-done future identical to `prior_inflight` fails both predicates, so the generation bump is correctly skipped.

Also realigned the line-459 comment (the bump happens immediately *before* `build_httpx_cookies_from_storage` reloads, not after) per a Codex Optional review nit.

## Test plan

- [x] `test_cancel_before_subprocess_registers_no_phantom_bump` — cold registry; red-gate verified by stashing the fix (pre-fix bumped 0 → 1)
- [x] `test_cancel_before_register_with_warm_registry_no_phantom_bump` — seeds stale done future + `_REFRESH_GENERATIONS=1`; red-gate verified (pre-fix bumped 1 → 2)
- [x] All 6 refresh-race tests pass
- [x] Full 5383-test unit + integration suite stays green
- [x] mypy clean (124 files), ruff format + lint clean

## Polish receipts (`/polish` pipeline from `/execute-task` Step 4)

- Stage 1 (code-simplifier): comment refinement on the post-cancel propagation block (enumerates both `observed_inflight` branches)
- Stage 2 (triple review: claude + codex + gemini in parallel):
  - **claude** — no critical/important, walk-through verified
  - **gemini** — "Verification Status: Passed"
  - **codex** — found a CRITICAL bug in the v1 fix (warm-registry case) + Important test gap + Optional stale comment. All three fixed before opening this PR.
- Stage 3 (ultrathink): no additional findings; reasoned through 9 edge-case scenarios for the new discriminator and confirmed type narrowing, thread safety, and re-entry correctness.

## Deferred polish findings

None.

## Summary by CodeRabbit

* **Bug Fixes**
  - Improved robustness of authentication token refresh logic to prevent generation counter issues during concurrent refresh attempts
  - Enhanced cancellation handling to properly manage active and stale refresh operations
  - Added regression test coverage for edge cases in refresh coalescing scenarios

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/826?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/notebooklm/_auth/refresh.py
+++ b/src/notebooklm/_auth/refresh.py
@@ -456,11 +456,14 @@ async def _fetch_tokens_with_refresh(
         refresh_token = _REFRESH_ATTEMPTED_CONTEXT.set(True)
         try:
             async with _get_refresh_lock(refresh_storage_path):
-                # Bump generation ONLY after the subprocess succeeds AND
-                # storage is reloaded. An earlier implementation bumped the
-                # generation eagerly BEFORE ``_run_refresh_cmd`` — when the
+                # Bump generation ONLY after the current-attempt subprocess
+                # succeeds — never eagerly. An earlier implementation bumped
+                # the generation BEFORE ``_run_refresh_cmd``; when the
                 # subprocess failed, the phantom bump made concurrent waiters
-                # short-circuit and proceed with stale storage.
+                # short-circuit and proceed with stale storage. The bump
+                # itself happens just below, immediately before
+                # ``build_httpx_cookies_from_storage`` reloads the freshly-
+                # written disk state.
                 #
                 # Re-check under the sync state lock so the read is atomic
                 # ACROSS event loops. The per-loop asyncio lock only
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the reviews — quick triage of claude[bot]'s 4 non-blocking observations:

1. **`prior_was_active` computed outside `_REFRESH_STATE_LOCK`** — claude[bot] itself notes this is benign. Stronger: between `with _REFRESH_STATE_LOCK: prior_inflight = registry.get(...)` (release) and the next line `prior_inflight.done()`, the current coroutine doesn't yield, so no callback can transition the Future's state on this loop. Cross-thread futures on other loops are owned by their own loops' threads and can't mutate ours mid-statement either. Lock-narrowing is intentional. **Not changing.**

2. **Module-level state cleanup in tests** — the existing module-level `@pytest.fixture(autouse=True) _clear_refresh_state` clears `_REFRESH_GENERATIONS` before/after each test (see [test_refresh_cmd_race.py:51-56](https://github.com/teng-lin/notebooklm-py/blob/worktree-issue-816/tests/integration/concurrency/test_refresh_cmd_race.py#L51-L56)), and the per-loop registry dict is GC'd via `WeakKeyDictionary` keyed on the (function-scoped) loop. No cross-test pollution risk. **Not changing.**

3. **`asyncio.gather(..., return_exceptions=True)` for one coroutine** — style consistency: matches the sibling `test_cancel_settle_race_does_not_bump_on_failure` in the same file ([L504-516](https://github.com/teng-lin/notebooklm-py/blob/worktree-issue-816/tests/integration/concurrency/test_refresh_cmd_race.py#L504-L516)). Pivoting only the new two tests to `pytest.raises` would create cross-test inconsistency. **Not changing.**

4. **`prior_inflight`/`prior_was_active` staleness on loop re-entry** — claude[bot] correctly classifies this as "no bug, worth noting". The pre-loop snapshot semantics are intentional: a fresh future inserted by the leader (`inflight is not prior_inflight`) routes to current-attempt, and an unchanged `prior_was_active=True` future stays classified as a followable sibling across iterations. Documented inline at [refresh.py:497-516](https://github.com/teng-lin/notebooklm-py/blob/worktree-issue-816/src/notebooklm/_auth/refresh.py#L497-L516). **Not changing.**

Gemini & CodeRabbit had no actionable findings. Per the issuing `/execute-task --no-merge` invocation, this PR is parked open with the merge gate clean for the maintainer to land.

PR #1082, opened 2026-05-27
fix(artifacts): default infographic generation options

### LINKED ISSUE #1063: Bug: CREATE_ARTIFACT for kind=infographic consistently returns null result (instant 2s fail)
## Summary

In `notebooklm-py 0.5.0`, generating an artifact with `kind=infographic` (RPCMethod kind code `7`) consistently fails ~2 seconds after dispatch with:

```
notebooklm.rpc.UnknownRPCMethodError: safe_index drift at path ()[0]
  [method_id='R7cb6c', path=(), source='_parse_generation_result', data_at_failure='None']
```

`data_at_failure='None'` means the upstream NotebookLM service returns a `None` body for the CREATE_ARTIFACT call — not a wire-shape rotation. The `safe_index drift` label here is a bit of a misnomer because the result is genuinely empty, not just descended-wrong.

## Reproducibility

Tested via [`NotebookLMClient.artifacts.create(kind="infographic", ...)`](https://github.com/teng-lin/notebooklm-py/blob/v0.5.0/src/notebooklm/_artifact_generation.py) against the same notebook that successfully delivers other artifact kinds:

| account  | source                                                                  | infographic | audio | report | study_guide | quiz | flashcards | slide_deck | data_table | mind_map |
|----------|-------------------------------------------------------------------------|-------------|-------|--------|-------------|------|------------|------------|------------|----------|
| Google A | `en.wikipedia.org/wiki/Bayes'_theorem` (prose-heavy)                    | ❌ 2s null  | ✅    | ✅     | ✅          | ✅   | ✅         | ✅         | ✅         | ✅       |
| Google B | `en.wikipedia.org/wiki/List_of_countries_by_GDP_(nominal)` (data tables) | ❌ 2s null  | ✅    | ✅     | ✅          | ✅   | ✅         | ✅         | ✅         | ✅       |

So:
- Two different Google accounts → same failure.
- Two source-content profiles (prose vs structured tables) → same failure.
- Every other artifact kind works on both notebooks/accounts.

The instant ~2s turnaround (not the multi-minute generation time the docstring suggests at `cli/services/artifact_generation.py:33` "infographic: typically 1-3 min") strongly suggests NotebookLM service rejects the request immediately rather than producing then failing.

## Reference code path

The error originates in [`ArtifactsAPI._parse_generation_result`](https://github.com/teng-lin/notebooklm-py/blob/v0.5.0/src/notebooklm/_artifact_generation.py) which calls `safe_index(result, 0, 0, method_id=RPCMethod.CREATE_ARTIFACT.value, ...)` — `result` itself is `None`, so descent fails on the very first index.

```python
# notebooklm/rpc/types.py:139 (v0.5.0)
INFOGRAPHIC = 7

# notebooklm/rpc/types.py:79
CREATE_ARTIFACT = "R7cb6c"
```

The CLI also recognizes this kind: `cli/generate_cmd.py:543` `def generate_infographic`. So the wrapper *thinks* the kind is supported.

## Hypothesis (would appreciate maintainer perspective)

1. **NotebookLM service-side gating for infographic** — perhaps requires NotebookLM Plus / Pro subscription, paid account tier, or a regional A/B test enrollment. If so, the library could detect the empty result and raise a more specific `ArtifactFeatureUnavailableError` rather than the generic `UnknownRPCMethodError`.
2. **Stale params for INFOGRAPHIC kind=7** — the request payload for kind=7 might have drifted vs what the web UI now sends (extra required field, renamed param, etc.). Other kinds may have been independently updated while infographic wasn't.
3. **NotebookLM-side temporary outage** for infographic — possible but unlikely to affect both my accounts for over 12 hours consecutively (initial repro 2026-05-27 02:46 UTC, still failing at 05:07 UTC).

## What would help

- Confirmation whether infographic still works for the maintainer or other users on a non-Plus Google account.
- A capture of the wire request the web UI sends for "Create infographic" vs what `notebooklm-py 0.5.0` sends, to spot any param shape drift.
- If feature-gated, ideally a dedicated exception class so downstream consumers can detect "this artifact kind is unavailable for this account/notebook" without parsing the `data_at_failure='None'` heuristic.

I'm happy to provide additional logs or HAR captures if it helps narrow this down. Thanks for the great library — the recent ArtifactRowAdapter / safe_index work is sharp.

---

**Environment**
- `notebooklm-py 0.5.0` (latest tag as of 2026-05-26)
- Python 3.12
- Linux x86_64
- Auth: `NotebookLMClient.from_storage(storage_state_path)` (Playwright storage_state JSON from Chrome cookie export)

teng-lin:
## Summary
- fixes Python API `generate_infographic()` defaults so it sends the same concrete visual options as the CLI (`LANDSCAPE`, `STANDARD`, `AUTO_SELECT`)
- preserves explicit infographic visual option overrides in the RPC payload
- classifies a null CREATE_ARTIFACT response as `ArtifactFeatureUnavailableError` instead of a schema-drift `UnknownRPCMethodError`

Fixes #1063.

## Root Cause
The CLI already supplied concrete defaults for infographic orientation/detail/style. The Python API accepted `None` defaults and encoded those as null visual option slots. The live NotebookLM API now rejects that null infographic config before task creation, returning a null RPC result.

## Live API Verification
- Before the default-option fix, `client.artifacts.generate_infographic(notebook_id)` reproduced the issue path as a null CREATE_ARTIFACT result.
- Explicit CLI-equivalent defaults succeeded.
- After this fix, the default Python call started task `a10be54f-3b63-4946-82ae-4c843c8f3c79`, poll returned `in_progress`, and cleanup delete returned `true`.

## Local Verification
- `uv run notebooklm doctor`
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run mypy src/notebooklm`
- `uv run pytest tests/unit/test_source_selection.py::TestArtifactsSourceSelection::test_generate_infographic_source_encoding tests/unit/test_source_selection.py::TestArtifactsSourceSelection::test_generate_infographic_visual_option_encoding tests/integration/test_artifacts_integration.py::TestArtifactsAPI::test_generate_infographic_null_result_raises_feature_unavailable -q`
- `uv run pytest` (`6511 passed, 12 skipped`)
- `uv run pre-commit run --all-files`

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -232,12 +232,18 @@ sit at the intersection — they're catchable as **any** of `NotFoundError`
 | `NotebookNotFoundError` | `NotFoundError`, `RPCError`, `NotebookError`, `NotebookLMError` |
 | `SourceNotFoundError` | `NotFoundError`, `RPCError`, `SourceError`, `NotebookLMError` |
 | `ArtifactNotFoundError` | `NotFoundError`, `RPCError`, `ArtifactError`, `NotebookLMError` |
+| `ArtifactFeatureUnavailableError` | `RPCError`, `ArtifactError`, `NotebookLMError` |
 
 Use the table to pick the right level of catch. `client.sources.get(...)`
 returns `None` for a missing source rather than raising; the workflows that
 do raise `SourceNotFoundError` are `client.sources.get_fulltext(...)` and
 `client.sources.wait_until_ready(...)`. Artifact-download workflows raise
 `ArtifactNotFoundError` when a requested artifact ID is not in the listing.
+Artifact generation workflows may raise `ArtifactFeatureUnavailableError`
+when NotebookLM accepts the RPC but returns no generation task for a specific
+artifact feature. For infographic generation, a null `CREATE_ARTIFACT` result
+is reported this way instead of surfacing as schema drift or a failed
+`GenerationStatus`.
 
 ##### Catching any "not found" across domains
 
--- a/docs/stability.md
+++ b/docs/stability.md
@@ -89,7 +89,7 @@ ClientError, ConfigurationError, ValidationError
 # found" types — see docs/python-api.md#error-handling for migration prose.
 SourceError, SourceAddError, SourceProcessingError, SourceTimeoutError, SourceNotFoundError
 NotebookError, NotebookNotFoundError
-ArtifactError, ArtifactDownloadError, ArtifactNotFoundError, ArtifactNotReadyError, ArtifactParseError
+ArtifactError, ArtifactDownloadError, ArtifactFeatureUnavailableError, ArtifactNotFoundError, ArtifactNotReadyError, ArtifactParseError
 ChatError
 
 # Enums
--- a/src/notebooklm/__init__.py
+++ b/src/notebooklm/__init__.py
@@ -57,6 +57,7 @@ from .exceptions import (
     # Domain: Artifacts
     ArtifactDownloadError,
     ArtifactError,
+    ArtifactFeatureUnavailableError,
     ArtifactNotFoundError,
     ArtifactNotReadyError,
     ArtifactParseError,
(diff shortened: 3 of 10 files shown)


teng-lin:
@claude review

PR #1110, opened 2026-05-28
test(cli): expand services boundary test to three-state inventory

teng-lin:
## Summary

- Extends `tests/unit/cli/test_services_boundary.py` so every file under `cli/services/` (recursively, excluding `__init__.py`) is classified into exactly one of three sets: `GUARDED_PATHS`, `TRANSITIONAL_GUARDED_PATHS`, or `WAIVED_PATHS`.
- Adds a Pattern A AST assertion: a function body MAY NOT contain co-occurring `console.print` + `exit_with_code` calls unless the module is in `TRANSITIONAL_GUARDED_PATHS` with a matching `pattern_a_violations` declaration.
- Adds `test_inventory_completeness` so new service modules fail CI until they're classified.
- No production source changes — this PR only extends the test infrastructure.

## What changed

- `GUARDED_PATHS` picks up `cli/services/research.py` (verified clean by the existing import scanner with zero Pattern A pairs).
- `TRANSITIONAL_GUARDED_PATHS` schema migrates from `(path, list[str])` tuples to a richer dict (`path` / `forbidden_imports` / `pattern_a_violations` / optional `pattern_b_violations` / `rationale`) so both checks read from one source of truth.
- `WAIVED_PATHS = {}` (empty initially) with a documented schema for future architecture-exception entries.
- 22 new entries in `TRANSITIONAL_GUARDED_PATHS` covering every service module that currently violates the import boundary OR has Pattern A pairs. Pre-implementation audit verified every `(function_name, line)` citation against the live source via an AST walk.

## How drift is detected

- Adding a new `forbidden_import` or Pattern A pair to a transitional module fails the test.
- Removing one without updating the inventory fails the test.
- Silent line-shifts (e.g. an unrelated edit moving a line) fail the test — the assertion compares declared `(function_name, line)` tuples against the live AST scan, so any drift is surfaced in the same PR.

## Test plan

- [x] `uv run ruff format tests/unit/cli/test_services_boundary.py` — file formatted.
- [x] `uv run ruff check tests/unit/cli/test_services_boundary.py` — all checks passed.
- [x] `uv run mypy src/notebooklm --ignore-missing-imports` — no issues across 174 source files.
- [x] `uv run pytest tests/unit/cli/test_services_boundary.py -v` — 35/35 passed (24 parametrized transitional + 3 parametrized guarded + 2 new inventory checks + 6 helper-sanity tests).
- [x] `uv run pytest tests/unit/cli -q` — 1521 passed, 36 skipped (pre-existing optional skips, no regressions).

## Deferred polish findings

none

### DIFF
--- a/tests/unit/cli/test_services_boundary.py
+++ b/tests/unit/cli/test_services_boundary.py
@@ -29,12 +29,13 @@ service module must appear in exactly one set. New modules added under
 ``cli/services/`` will fail the test until classified.
 
 Pattern A definition: ``console.print`` and ``exit_with_code`` co-occur as
-Pattern A iff both are called from within the SAME ``ast.FunctionDef`` body
-(at any nesting depth inside that function, but NOT crossing into a nested
-``FunctionDef``). The implementation in :func:`_pattern_a_pairs` reports one
-pair per ``exit_with_code`` call site so that line drift after a refactor
-elsewhere is caught — silent shifts (e.g. an unrelated edit moving the
-line) would otherwise mask a real regression.
+Pattern A iff both are called from within the SAME
+``ast.FunctionDef | ast.AsyncFunctionDef`` body (at any nesting depth inside
+that function, but NOT crossing into a nested ``FunctionDef`` /
+``AsyncFunctionDef``). The implementation in :func:`_pattern_a_pairs`
+reports one pair per ``exit_with_code`` call site so that line drift after
+a refactor elsewhere is caught — silent shifts (e.g. an unrelated edit
+moving the line) would otherwise mask a real regression.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the careful review. Responses inline:

**1. Docstring says `ast.FunctionDef` but handles `AsyncFunctionDef`** — applied in 362bf916. Module docstring and `_pattern_a_pairs` docstring now spell out `ast.FunctionDef | ast.AsyncFunctionDef`.

**2. `rookiepy_errors.py` is documentation-only in TRANSITIONAL** — confirmed observation, but keeping it in TRANSITIONAL by design. The task spec explicitly says "Do NOT add waivers beyond cookie_domains.py (the documented Pattern B parser-time exception)", and the dispatched 12-entry list named `rookiepy_errors.py` for TRANSITIONAL. The rationale string captures the reasoning: the module's reach-in uses level-3 `...rendering` which `_boundary_violations` doesn't currently scan; broadening the scanner to level-3 is a deliberate follow-up (one PR, not bundled with the inventory expansion), at which point this entry naturally picks up real `forbidden_imports` content and the documentation-only window closes. WAIVED_PATHS is reserved for indefinite architecture exceptions; this is a planned migration target, not an exception.

**3. Unclassified-module failure message misleading** — applied in 362bf916. New wording leads with the classification requirement and lists pair count as contextual:
```
{name}: unclassified module — classify into GUARDED_PATHS,
TRANSITIONAL_GUARDED_PATHS, or WAIVED_PATHS
(Pattern A pairs found: {sorted(actual)}).
```

**4. `node.lineno` accessed on `ast.AST`-typed variable** — not applying. Mypy passes in the project's current strict-mode config (verified across 174 source files). Both `_is_console_print_call` and `_is_exit_with_code_call` short-circuit on the `isinstance(node, ast.Call)` check internally; at the call site the runtime invariant is guaranteed. Adding a `cast(ast.Call, node).lineno` would document a narrowing that's already done implicitly, at the cost of an extra `cast` import for every reader. Happy to revisit if the project tightens mypy further.

**5. Pattern A bypass via `cli_print` or similar wrappers** — acknowledged, and explicitly documented in the `source_clean.py` entry's rationale already. The `forbidden_imports` check still catches `..rendering`/`..error_handler` reach-in for those modules, so the boundary is enforced at the import level even where Pattern A is silent. Worth noting for future readers, agreed; the per-entry rationale is the right place for that note (rather than the top-level docstring) so it lives next to the affected module.

Re-verified after the doc updates: `uv run pytest tests/unit/cli/test_services_boundary.py -v` → 35/35 passed.

PR #1111, opened 2026-05-28
fix(cli): redact subprocess stderr in playwright_login before printing (audit G4)

teng-lin:
Implements: meta-audit **G4** (`/.sisyphus/plans/cli-audit-2026-05-27-fixes.md` task F16).

## Summary

- Adds `redact_subprocess_output(text, env=None) -> str` to `cli/services/playwright_login.py` — strips ANSI control sequences and replaces every occurrence of each non-trivial environment-variable value with `<redacted>`. Walks JSON-shaped env values (e.g. `NOTEBOOKLM_AUTH_JSON`) so nested leaf strings are also redaction candidates.
- Applies the helper at both subprocess result sites in `ensure_chromium_installed`: the dry-run probe (unexpected stderr → sanitised debug log) and the install-failure path (sanitised diagnostic tail printed alongside the existing "run manually" hint).
- Pins the contract with 18 unit tests using stub `CompletedProcess` fixtures — no Playwright install required.

## Why

`ensure_chromium_installed` invokes the playwright CLI via `subprocess.run(..., capture_output=True)`. The captured stderr/stdout can carry two classes of noise we never want to surface raw:

1. **Environment-variable values** — the playwright CLI inherits the parent environment; if any secret (PSIDTS / cookie material set via `NOTEBOOKLM_AUTH_JSON` / SAPISID / etc.) ever gets interpolated into a traceback or config line by the CLI, it lands verbatim in `result.stderr`. Today nothing prints those streams, but the audit flagged the gap for the "next time we want to surface diagnostics" failure mode — and this PR makes the install-failure path actually emit a (sanitised) diagnostic tail so users have something to act on.
2. **ANSI control sequences** — pip/playwright progress bars and colour codes that dirty log scrapers and console output.

## Implementation notes

- **ANSI is stripped BEFORE redaction.** Otherwise a secret split by an inert reset (`"abc\x1b[0m123"`) wouldn't match the env value `"abc123"`, and the subsequent ANSI strip would reassemble the secret in cleartext.
- **JSON-nested values are expanded.** `NOTEBOOKLM_AUTH_JSON` carries a dict whose leaf strings (cookie tokens) are the actual secrets; a subprocess that re-emits a parsed value would otherwise bypass exact-string matching.
- **Stderr/stdout short-circuit compares on `.strip()`** so a stderr that sanitises down to whitespace (e.g. pure ANSI progress noise) doesn't shadow a stdout line carrying the actionable failure.
- **Candidate values are ansi-stripped too**, in case an env value itself carries embedded escape bytes (rare, but the input `text` has already been stripped above).
- Conservative skip-list: empty / short (`< 3 chars`) / well-known constants (`true`, `false`, `yes`, `no`, …) never trigger redaction, to avoid spamming `<redacted>` across normal path / boolean output.
- `dict(os.environ)` snapshot defends against concurrent env mutation.

## Test plan

`tests/unit/cli/test_playwright_login_stderr.py` (18 tests, all using stub `CompletedProcess` fixtures):

- [x] PSIDTS-style secret in stderr → `<redacted>`.
- [x] Multiple env values in one string → all redacted.
- [x] Trivial values (`""`, `/`, `true`, single-char, < 3-char, …) → never redacted.
- [x] Longer secret wins when one is a substring of another.
- [x] CSI / OSC / two-byte / full C1 Fe (0x40-0x5F) ANSI sequences stripped.
- [x] ANSI-interrupted secret reassembled and redacted (regression).
- [x] Nested JSON leaf values redacted (NOTEBOOKLM_AUTH_JSON case + array case).
- [x] Malformed JSON falls back to exact-match redaction.
- [x] Empty input returns empty.
- [x] Defaults to `os.environ` when `env=None`.
- [x] `ensure_chromium_installed` install-failure path routes captured stderr through helper, with no raw secret or ANSI in output.
- [x] Whitespace-only sanitised stderr falls back to sanitised stdout.
- [x] Dry-run debug log path also sanitised.

Verification:

```
uv run ruff format src/notebooklm/cli/services/playwright_login.py tests/unit/cli/test_playwright_login_stderr.py
uv run ruff check src/notebooklm/cli/services/playwright_login.py tests/unit/cli/test_playwright_login_stderr.py
uv run mypy src/notebooklm --ignore-missing-imports                       # Success: no issues found in 174 source files
uv run pytest tests/unit/cli/test_playwright_login_stderr.py -v           # 18 passed
uv run pytest tests/unit/cli -q                                           # 1547 passed
```

## Deferred polish findings

None.

### DIFF
--- a/src/notebooklm/cli/services/playwright_login.py
+++ b/src/notebooklm/cli/services/playwright_login.py
@@ -146,9 +146,9 @@ CHANNEL_BROWSERS: dict[str, tuple[str, str]] = {
 #
 #   1. Environment-variable VALUES — Playwright forwards the parent process
 #      environment; if any secret (PSIDTS, API tokens, cookie material set
-#      via NOTEBOOKLM_AUTH_JSON / SAPISID / etc.) is interpolated into a
-#      traceback / config line by the CLI we invoke, it lands verbatim in
-#      ``result.stderr``.
+#      via the auth-source env var / SAPISID / etc.) is interpolated into
+#      a traceback / config line by the CLI we invoke, it lands verbatim
+#      in ``result.stderr``.
 #
 #   2. ANSI control sequences — pip/playwright CLIs emit progress bars and
 #      colour codes that mangle log scrapers and dirty test snapshots.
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

PR #1179, opened 2026-05-29
fix(cli): exit non-zero when `doctor` checks fail

### LINKED ISSUE #1160: `notebooklm doctor` exits 0 even when checks FAIL
## Summary
`notebooklm doctor` reports `fail` checks but always exits 0, so it reads as green in CI / monitoring even when auth is broken or migration is incomplete.

## Severity
**High** (operational) — false-green health check. Confidence: high (reproduced).

## Evidence
- `src/notebooklm/cli/doctor_cmd.py:51-171` (`_run_doctor`) builds entries with `status: "fail"` but never exits non-zero; both the JSON branch (`json_output_response(result); return`) and the text branch (`_display_results(...)`) return normally.
- Reproduced: `NOTEBOOKLM_HOME=$(mktemp -d) notebooklm doctor; echo $?` → `0` with an auth check of `{'status': 'fail', 'detail': 'not authenticated'}`.
- Contrast: `auth check` exits 1 on the same condition (`src/notebooklm/cli/session_cmd.py:399-401`).

## Failure mode
A health check in CI, `set -e` scripts, or monitoring reports success while the install is actually broken.

## Suggested fix
Exit non-zero when any check fails, consistent with `auth check` and the CLI exit-code convention.

---
_Identified during a runtime-correctness architecture review (round 2, 2026-05-29) via automated multi-lens audit; the top findings were verified against the code. Please confirm before scheduling._

teng-lin:
## Root cause

`notebooklm doctor` (`src/notebooklm/cli/doctor_cmd.py`) builds check entries with `status: "fail"` (broken auth, incomplete migration, missing profile dir, invalid config) but `_run_doctor` always returns normally. Both the JSON branch (`json_output_response(result); return`) and the text branch (`_display_results(...)`) exited 0 even on failures.

Reproduction from the issue:
```
NOTEBOOKLM_HOME=$(mktemp -d) notebooklm doctor; echo $?   # -> 0, with auth {'status': 'fail'}
```

A health check in CI, a `set -e` script, or a monitoring probe reads the broken install as **green**. By contrast `auth check` already exits 1 on the same condition.

## Fix

Compute overall health from the **final** check states (after any `--fix` repairs have run) and `exit_with_code(1)` when any check still has `status == "fail"`, in **both** the JSON and text branches — consistent with `auth check` and the CLI exit-code convention. The exit happens *after* the payload/table is emitted, so machine-readable output is preserved. Warnings remain non-fatal (exit 0).

## Tests

- Updated the existing failing-check unit/integration cases to assert exit 1 (they previously asserted 0).
- Added regression tests: text-mode doctor exits 1 on failure (the exact issue repro), text-mode exits 0 when all pass, `--fix` that leaves auth failing still exits 1, and a warn-only layout keeps exit 0.

Closes #1160

## Summary by CodeRabbit

* **Bug Fixes**
  * The `doctor` command now properly exits with code 1 when health checks fail and exits with code 0 when all checks pass, in both text and JSON output modes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/1179?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/notebooklm/cli/doctor_cmd.py
+++ b/src/notebooklm/cli/doctor_cmd.py
@@ -17,7 +17,7 @@ from ..paths import (
     get_profile_dir,
     get_storage_path,
 )
-from .error_handler import handle_errors
+from .error_handler import exit_with_code, handle_errors
 from .rendering import console, json_output_response
 
 logger = logging.getLogger(__name__)
@@ -156,6 +156,12 @@ def _run_doctor(fix_issues: bool, *, json_output: bool) -> None:
     if fix_issues:
         fixes_applied = _apply_fixes(checks, home, profile_dir)
 
+    # Determine overall health from the *final* check states (after any fixes
+    # have been applied above). A lingering "fail" means the install is broken,
+    # so exit non-zero — consistent with ``auth check`` and the CLI exit-code
+    # convention — instead of reading as green in CI / ``set -e`` scripts.
+    has_failures = any(c["status"] == "fail" for c in checks.values())
+
     # Output
     if json_output:
         result = {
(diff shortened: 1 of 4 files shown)


teng-lin:
@claude review

Issue #1205, opened 2026-05-29
Refactor: collapse duplicate implementations of one concept (divergence-by-duplication)


teng-lin:
_From the v0.6.0 architecture gap review (2026-05-29)._

## Summary

The same concept is implemented twice (or more) in divergent ways across the codebase. This duplication-with-divergence pattern recurs and is the **top defect generator** in the review: each duplicate drifts from its twin, and the divergence is where bugs live.

## Instances

- **Two source parsers.** `Source.from_api_response` in `src/notebooklm/_types/sources.py:164` vs the parsing in `src/notebooklm/_source_listing.py:162`. The classmethod variant omits `status`, so the two paths produce non-equivalent `Source` objects.
- **Two auth-refresh-and-retry paths.** `src/notebooklm/_middleware_auth_refresh.py:246` vs `src/notebooklm/_rpc_executor.py:266`. They use different guards and raise different exception types for the same logical failure.
- **Two anti-XSSI strippers.** `src/notebooklm/rpc/decoder.py:165` vs `src/notebooklm/_chat_protocol.py:148`.
- **Two UPDATE_SOURCE rename payloads.** Inline construction in `src/notebooklm/_sources.py:656` vs the builder in `src/notebooklm/_source_upload_payloads.py:31`.
- **Re-declared generate_* signatures.** `src/notebooklm/_artifacts.py` re-declares 11 `generate_*` signatures that already exist in `src/notebooklm/_artifact_generation.py`.
- **Parallel verb surfaces.** `SourcesAPI` vs `SourceUploadPipeline` carry parallel `rename` / `wait_until_*` / `get` / `list` verbs.

## Suggested direction

One concept -> one owner. For each pair, pick a single authoritative implementation and have the other call it (or delete it). Prioritize the source-parser and auth-refresh-retry pairs since their divergence is behaviorally significant (missing `status`, mismatched exception types).

teng-lin:
All five consolidations from this issue are merged:
- #1236 — unify the anti-XSSI stripper (chat parser → shared `strip_anti_xssi`)
- #1237 — unify the two Source parsers via `Source.from_row` (also fixed the add/rename-path `status` always-READY bug + the `int`→`SourceStatus` annotation)
- #1238 — Sources rename-payload + pipeline-verb dedup (`SourcesAPI` is now the single owner of the lister/poller)
[...]

PR #1236, opened 2026-05-30
refactor(chat): unify anti-XSSI prefix stripper

teng-lin:
Part of #1205 (2/5: unify anti-XSSI stripper)

## What

The streamed-chat wire parser (`src/notebooklm/_chat_wire.py`) re-implemented the `)]}'` anti-XSSI prefix removal as a blind `response_text[4:]` slice, duplicating the canonical regex-anchored `strip_anti_xssi` in `src/notebooklm/rpc/decoder.py`. Issue #1205 flags this duplicate-with-divergence pattern as the top defect generator.

This collapses the two into one owner: the chat parser now imports and calls the shared `strip_anti_xssi`.

## Why it's safe (happy path byte-for-byte identical)

The shared `strip_anti_xssi` is regex-anchored on `\)]\}'\r?\n` — it consumes the prefix plus its trailing newline. The chat parser still runs `.strip().split("\n")` right after. On the real chat wire format the prefix is always emitted followed by a newline (`)]}'\n...`), so:

- old: `[4:]` removes the 4-char prefix, then `.strip()` removes the leading newline
- new: `strip_anti_xssi` removes prefix + newline, then `.strip()` is a no-op on the boundary

Both produce the identical `lines` list. Verified against all wire fixtures and the `_length_prefixed` test helper, which all use `)]}'\n`.

## Tests

- Added `test_xssi_prefix_strip_matches_shared_helper_on_real_wire_format` — parse result identical with/without the prefix.
- Added `test_chat_parser_uses_shared_strip_anti_xssi` — asserts the shared helper is the one invoked.
- All existing chat-parser / decoder / streaming tests stay green (612 passed).

## Gates

- `ruff check` / `ruff format`: clean
- `mypy src/notebooklm`: clean
- `audit_public_api_compat.py`: no new public break
- `pytest -k "chat or stream or citation or wire or decoder or xssi"`: 612 passed, 2 skipped

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Improved code consistency in chat response handling by consolidating prefix-stripping logic.

* **Tests**
  * Added comprehensive unit tests for anti-XSSI prefix handling in chat response parsing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/1236?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #1237, opened 2026-05-30
fix(sources): unify the two divergent Source parsers

teng-lin:
Part of #1205 (1/5: unify Source parsers)

## Problem

There were **two divergent parsers** that turn a raw RPC source row into a `Source`:

- `Source.from_api_response` (`src/notebooklm/_types/sources.py`) — used by the
  `ADD_SOURCE` / rename paths (`_source_add.py`, `_sources.py`, `_source_upload.py`).
  It built a `Source` **without `status`**, so the field always fell back to the
  `status: int = SourceStatus.READY` default. The annotation was also wrong: the
  field actually holds a `SourceStatus`, not a bare `int`.
- `SourceLister._parse_source` (`src/notebooklm/_source_listing.py`) — used by the
  `client.sources.list()` / `get()` / poll paths. It built `Source(..., status=row.status, ...)`
  from a `SourceRow`, correctly carrying the decoded processing status.

A `Source` returned from an add/rename therefore silently reported `READY` even when
the wire carried a `PROCESSING`/`ERROR`/`PREPARING` status block.

## Change

Both paths now funnel through a **single construction site**, `Source.from_row(SourceRow)`:

- `Source.from_api_response` normalizes the wire shape via `SourceRow.from_unknown_shape`
  and defers to `from_row` (which sets `status=row.status`).
- `SourceLister._parse_source` wraps the entry via `SourceRow.from_entry` and defers to
  the same `from_row`.

The flat shape still yields `_type_code=None` and `SourceStatus.READY` (the `SourceRow`
returns those defaults when no metadata / status block is present), so all prior behavior
is preserved — `list()` / `get()` and the poller produce identical `Source` instances,
and add/rename results now carry the real `status` / `url` / `created_at`.

The `Source.status` field annotation is corrected from `int` to `SourceStatus`
(still an `int`-compatible enum, so equality and the public-API audit are unaffected).

## Tests

Added to `tests/unit/test_types.py::TestSource`:
- `test_from_api_response_carries_status` — a PROCESSING status block is now decoded.
- `test_from_api_response_status_defaults_ready_without_block` — no status block still
  defaults to READY (flat and medium-nested).
- `test_from_api_response_matches_listing_path` — the classmethod and the `GET_NOTEBOOK`
  listing path produce **identical** `Source` instances from the same entry.

## Gates

- `ruff check` / `ruff format --check`: clean
- `mypy src/notebooklm --ignore-missing-imports`: clean
- `scripts/audit_public_api_compat.py`: no new public break (10 pre-existing allowlisted)
- `pytest -k "source or row_adapter or listing or types"`: 1562 passed
- full `tests/unit tests/integration`: 7254 passed, 52 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Source objects now report real processing status (no longer defaulting to READY).
  * Source creation is unified across flows so status, URL, and creation timestamp are consistent.
  * Status typing corrected so status-related properties behave predictably.

* **Tests**
  * Added unit tests covering status parsing and parity between construction paths.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/1237?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 > **Not removed:** awaiting `NotebookLMClient.from_storage(...)` still works —
 > its deprecation targets v1.0, not v0.6.0.
 
+### Fixed
+
+- **`Source.from_api_response` now reports the real processing `status`.** The
+  `ADD_SOURCE` / rename parsing path previously never read the status block and
+  always fell back to `SourceStatus.READY`, while `client.sources.list()` /
+  `get()` and the source poller read the decoded status. Both parsers now
+  funnel through a single `Source.from_row` construction site, so a `Source`
+  produced from an add/rename response carries the same `status` (and `url` /
+  `created_at`) as the listing path. The `Source.status` field annotation was
+  also corrected from `int` to `SourceStatus` (still an `int`-compatible enum).
+
 ## [0.6.0] - 2026-05-29
 
 ### Breaking changes
--- a/src/notebooklm/_source_listing.py
+++ b/src/notebooklm/_source_listing.py
@@ -148,14 +148,10 @@ class SourceLister:
             )
             return None
 
-        return Source(
-            id=row.id,
-            title=row.title,
-            url=row.url,
-            _type_code=row.type_code,
-            created_at=row.created_at,
-            status=row.status,
-        )
+        # Funnel through the single ``Source`` construction site shared
+        # with ``Source.from_api_response`` so the list/get/poll path and
+        # the ADD_SOURCE/rename path produce identical Sources.
+        return Source.from_row(row)
 
 
 __all__ = ["SourceLister"]
--- a/src/notebooklm/_types/sources.py
+++ b/src/notebooklm/_types/sources.py
@@ -6,13 +6,16 @@ import warnings
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import Enum
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from ..rpc.types import SourceStatus
 from .common import (
     UnknownTypeWarning,
     _datetime_from_timestamp,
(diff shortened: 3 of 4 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the reviews 🙏 — addressed the in-scope suggestions and filed follow-ups for the rest.

**Addressed in b654bb7c:**
- **Flat-shape `_type_code=None` invariant made explicit** — expanded the `Source.from_row` docstring to spell out that the flat invariant is now an emergent property of `SourceRow` (no list at `_raw[2]` → `metadata` is `None` → `type_code`/`url`/`created_at` are `None`, `status` is `READY`), rather than the old explicit early-return.
- **Deeply-nested-with-status coverage gap** — added `test_from_api_response_deeply_nested_carries_status` asserting the deeply-nested dispatch path decodes the status block (uses `SourceStatus.ERROR`).

**Filed as follow-ups (pre-existing, out of scope for this 1/5 slice):**
- Unused `notebook_id` parameter on `from_api_response` → #1241
- `method_id` not forwarded to `from_unknown_shape` (drift logs mis-tagged `GET_NOTEBOOK`) → #1242

The integration-through-`SourceLister` parity suggestion is left as-is for now per the review's "fine to leave" note.

Rebased onto latest `main` (#1236). Gates re-run green: ruff, mypy, public-API audit (no new break), targeted pytest (1563 passed).

@claude review

teng-lin:
Addressed the new finding in 74b67edf: removed the stale `docs/improvement.md` §6.2 cross-reference from the `from_api_response` docstring (that doc doesn't exist; the surrounding text already explains the bug + fix). Re-gated green (ruff, mypy, pytest).

(There are two other pre-existing `docs/improvement.md` references in `_runtime_transport.py` / `_runtime_init.py`, untouched here as they're outside this PR's scope.)

@claude review

Issue #1157, opened 2026-05-29
Auth-refresh replay re-issues non-idempotent writes, ignoring `disable_internal_retries`


teng-lin:
## Summary
The idempotency `disable_internal_retries` flag is the only duplicate-write guard for non-idempotent / probe-then-create RPCs, but it gates **only** `RetryMiddleware`'s 429/5xx loop. Two separate auth-refresh replay paths re-issue the same mutating POST without ever consulting it.

## Severity
**High** — correctness (duplicate writes / duplicate side effects). Confidence: high (verified by code read).

## Evidence
- `src/notebooklm/_middleware_auth_refresh.py:215-254` — on a 401/403 it refreshes the token and re-invokes `next_call` once, guarded only by `_refresh_callback_enabled` / `_is_auth_error` / `RPC_CONTEXT_AUTH_REFRESHED`. No check of the effective disable-retry flag.
- `src/notebooklm/_rpc_executor.py:278-294` — an auth-flavored `RPCError` in a *decoded* body re-runs the entire `rpc_call` via `try_refresh_and_retry(...)`; the guard at 280-284 does not gate on the idempotency classification.
- `src/notebooklm/_idempotency.py` — `disable_internal_retries` only feeds the 429/5xx retry loop.

## Failure mode
A non-idempotent create (`CREATE_NOTEBOOK`, `CREATE_ARTIFACT`, `CREATE_NOTE`, `ADD_SOURCE`, `SHARE_NOTEBOOK`, `GENERATE_MIND_MAP`) reaches the server and commits, then the response is a 401/403 (token expiring mid-flight, LB hiccup) or an auth-shaped `RPCError`. Both replay paths re-POST → duplicate resource, duplicate share-invite email, or wasted generation quota. The 429/5xx leg is correctly protected; the auth leg is not.

## Suggested fix
Make both replay paths honor the effective disable-retry classification: on an auth error for a non-idempotent / probe-then-create method, skip the replay, surface the original error, and let the caller's probe-then-create wrapper disambiguate.

---
_Identified during a runtime-correctness architecture review (round 2, 2026-05-29) via automated multi-lens audit; the top findings were verified against the code. Please confirm before scheduling._

PR #1240, opened 2026-05-30
refactor(auth): unify the two auth-refresh-and-retry paths

teng-lin:
Part of #1205 (5/5: unify auth-refresh-and-retry paths)

## Problem

Auth refresh-and-retry was implemented **twice** with divergent guards, exception shapes, and metrics:

- **HTTP-status layer** (`_middleware_auth_refresh.py`): catches a raw 401/403 `httpx.HTTPStatusError` from `Kernel.post`, refreshes, rebuilds the request envelope, retries the chain once. Once-per-call guard: `RPC_CONTEXT_AUTH_REFRESHED` boolean. Refresh failure → `TransportAuthExpired(...) from refresh_error`. Increments `rpc_auth_retries`.
- **Decoded-RPC layer** (`_rpc_executor.py::try_refresh_and_retry`): catches an auth-shaped decoded `RPCError` (HTTP 200 carrying an auth error in the batchexecute payload), refreshes, re-calls `rpc_call(_is_retry=True)`. Once-per-call guard: the `_is_retry` recursion flag. Refresh failure → original `RPCError` `from refresh_error`. Did **not** increment the metric.

Because the two once-per-call guards could not see each other, a `wire-401 → refresh#1 → 200 → decoded-auth-error → refresh#2` sequence drove **two** refreshes on a single logical call — the audit's named double-refresh concern.

## What this PR consolidates

New `src/notebooklm/_auth_refresh_retry.py` owns the common core exactly once:

- **`refresh_and_count(...)`** — the shared refresh body: log "auth error detected" → `await refresh()` → on failure log + raise the caller-shaped exception `from refresh_error` → optional `refresh_retry_delay` sleep → log "refresh successful" → increment `rpc_auth_retries`.
- **`RefreshBudget`** — a single-consume once-per-logical-call token.

Wiring:

- `RpcExecutor` mints **one** `RefreshBudget` per logical `rpc_call` and threads it through (a) the chain context via the new `RPC_CONTEXT_REFRESH_BUDGET` key (so `AuthRefreshMiddleware` consumes it) and (b) the decode-time retry recursion. The HTTP-status layer and the decoded-RPC layer now share **one** refresh allowance — the `wire-401-then-decoded-auth-error` sequence drives a single refresh.
- Both layers call `refresh_and_count`. The decoded layer now also increments `rpc_auth_retries`, fixing the metrics divergence the issue cites.

## Preserved exactly (no regression)

- **Distinct triggers stay distinct** — raw transport 401/403 vs decoded auth `RPCError`.
- **Distinct refresh-failure exception shapes stay distinct** — `TransportAuthExpired` from the chain (so `rpc_call`'s `except TransportAuthExpired` re-raises the raw `httpx.HTTPStatusError`, and chat maps to `ChatError`); the original `RPCError` from the decoder. Both are pinned by existing tests.
- **#1157 non-idempotent-replay guard** — the effective `disable_internal_retries` classification still suppresses replay on both paths.
- **Single-flight `AuthRefreshCoordinator.await_refresh` coalescing** — untouched.
- **Chat path** drives the chain without a budget, so `AuthRefreshMiddleware` falls back to its per-chain `RPC_CONTEXT_AUTH_REFRESHED` boolean — chat behavior is unchanged.

## What was deliberately NOT merged

The two layers' triggers and their externally-observable refresh-failure exception types were left distinct (callers/tests depend on both, and they are RPC-layer-appropriate: a wire HTTP error vs a decoded RPC error). Forcing one type would regress `test_no_retry_on_cookie_expiration` (HTTP-path failure surfaces a raw `httpx.HTTPStatusError`) and the decoded-path identity assertions. Per the issue's "if a full unification risks behavior change, implement only the safe subset," this PR ships the shared helper + shared once-per-call budget + the metrics-divergence fix.

## Gates

- `ruff check .` + `ruff format .`: clean
- `mypy src/notebooklm --ignore-missing-imports`: clean
- `python scripts/audit_public_api_compat.py`: no new breaks
- `pytest -k "auth or refresh or retry or idempot or rpc_executor or middleware"`: 1368 passed
- `pytest tests/integration/concurrency/`: 106 passed
- Full `pytest`: 7364 passed, 52 skipped

New tests: `tests/unit/test_auth_refresh_retry.py` (budget + shared body), executor-level budget-gate + metric tests in `tests/unit/test_rpc_executor.py`, and the headline `wire-401-then-decoded-auth-error refreshes once` + decoded-metric integration tests in `tests/integration/test_auto_refresh.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Consolidated auth-refresh so at-most-one token refresh occurs per logical API call, preventing duplicate retries and improving reliability.
  * Improved error propagation so original auth errors surface consistently after refresh attempts.
  * Ensures decode-time and transport-time refreshes share a single allowance to avoid redundant retries.

* **Documentation**
  * Updated docs to describe the consolidated refresh-and-retry behavior.

* **Tests**
  * Added integration and unit tests validating single-refresh behavior, retry counts, and auth-retry metrics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/1240?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -105,6 +105,7 @@ RPC Layer (rpc/)
 | `_backoff.py` | Shared capped exponential-backoff calculation with deterministic test injection |
 | `_reqid_counter.py` | `ReqidCounter` — monotonic `_reqid` for the chat backend |
 | `_runtime_auth.py` | `AuthRefreshCoordinator` — refresh task + auth-snapshot lock |
+| `_auth_refresh_retry.py` | Shared auth refresh-and-retry core for the two retry layers (HTTP-status `AuthRefreshMiddleware` + decoded-RPC `RpcExecutor`): the once-per-logical-call `RefreshBudget` token and the common `refresh_and_count` body (log/refresh/sleep/`rpc_auth_retries` metric). Unifies the previously-divergent copies per issue #1205; the two layers keep their distinct triggers and refresh-failure exception shapes. |
 | `_runtime_lifecycle.py` | `ClientLifecycle` — loop-affinity guard + keepalive task |
 | `_runtime_transport.py` | `RuntimeTransport` — authed-POST transport wrapper that drives the middleware chain and typed transport response handling |
 | `_rpc_executor.py` | RPC dispatch executor. Takes its `Kernel`, `RuntimeTransport`, `AuthRefreshCoordinator`, and `ClientMetrics` collaborators directly via keyword-only constructor parameters (ADR-014 Rule 5). The `RpcOwner` Protocol that previously re-declared the former `Session` facade's private attribute surface was deleted in Wave 4 of session-decoupling (#1068); only the local `DecodeResponse` Protocol remains. |
@@ -187,6 +188,7 @@ src/notebooklm/
 ├── urls.py                      # Public URL helper facade
 ├── utils.py                     # Public async utility helpers
 ├── _atomic_io.py                # Atomic JSON write/update helpers
+├── _auth_refresh_retry.py       # Shared auth refresh-and-retry core (RefreshBudget + refresh_and_count) for both retry layers
 ├── _backoff.py                  # Shared retry backoff calculation
 ├── _callbacks.py                # Sync/async callback invocation helper
 ├── _client_composed.py          # Client-owned composition holder
--- a/src/notebooklm/_auth_refresh_retry.py
+++ b/src/notebooklm/_auth_refresh_retry.py
@@ -1,0 +1,150 @@
+"""Shared auth refresh-and-retry core for the two retry layers.
+
+NotebookLM recovers from an auth failure at **two** distinct layers, and
+issue #1205 flagged that they were implemented as divergent copies:
+
+* **HTTP-status layer** — :class:`notebooklm._middleware_auth_refresh.AuthRefreshMiddleware`
+  catches a raw ``httpx.HTTPStatusError`` 400/401/403 from ``Kernel.post``,
+  refreshes, rebuilds the request envelope, and re-invokes the chain leaf
+  once.
+* **Decoded-RPC layer** — :meth:`notebooklm._rpc_executor.RpcExecutor.try_refresh_and_retry`
+  catches an auth-shaped decoded ``RPCError`` (HTTP 200 carrying an auth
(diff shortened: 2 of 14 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the review. Addressed the actionable findings in commit f2c71170:

- **Finding 1 (internal-param naming):** Renamed `refresh_budget` → `_refresh_budget` on `RpcExecutor.rpc_call` / `_execute_once` / `try_refresh_and_retry` to match the `_is_retry` leading-underscore convention. The transport-level `perform_authed_post(refresh_budget=...)` keyword stays unprefixed since it's a shared collaborator signature, not an internal-only one.
- **Finding 3 (always-None context key):** `RPC_CONTEXT_REFRESH_BUDGET` is now only inserted into the chain context when a budget is actually supplied, so the chat path leaves the key absent — matching the key's docstring ("Absent for callers that drive the chain without a budget"). The middleware's `.get(...)` → `None` fallback to the per-chain boolean is unchanged.
- **Finding 4 (consume() ordering):** Added a comment pinning that `_refresh_budget.consume()` must remain the last short-circuit operand, since it's side-effecting and reordering would burn the budget on calls that fall through to a plain raise.

Findings 2 (unused `refresh_error` in the middleware wrapper — intentional, `TransportAuthExpired` wraps the original HTTP error and the refresh error becomes `__cause__` via the shared helper) and 5 (spent-budget + `disable_internal_retries` combination — logically covered by the short-circuit ordering) were observational; no code change.

All gates re-run green after the changes: ruff/format/mypy clean, public-API audit no new breaks, full `pytest` 7364 passed.

@claude review

teng-lin:
Applied the CodeRabbit/Claude follow-up in commit c242fec6: `_refresh_budget` is now a **required** keyword-only argument on `try_refresh_and_retry` (no `None` default). The method is only reached from `_execute_once` after its `_refresh_budget.consume()` gate returns True, so the caller always holds the already-consumed shared budget — making it mandatory forecloses the theoretical second-refresh path a contrived direct call could otherwise create. The three direct test callers now pass an explicit `RefreshBudget()`.

Gates re-run green: ruff/format/mypy clean, full `pytest` 7364 passed.

@claude review

PR #1244, opened 2026-05-30
fix(artifacts): don't unwrap an already-flat suggest_reports list (#1243)

### LINKED ISSUE #1243: ArtifactsAPI.suggest_reports: don't unwrap an already-flat suggestions list
Split out of PR #1239 (the ArtifactGenerationService→ArtifactsAPI fold) as a pre-existing, out-of-scope finding from CodeRabbit.

## Problem

In `ArtifactsAPI.suggest_reports` (`src/notebooklm/_artifacts.py`), the response is unwrapped as:

```python
if result and isinstance(result, list) and len(result) > 0:
    items = result[0] if isinstance(result[0], list) else result
    for item in items:
        ...
```

`result[0]` is a list for both wrapped payloads (`[[row1, row2]]`) and already-flat payloads (`[row1, row2]`). In the **flat** case, `items` becomes the first suggestion row (`row1`), so the loop iterates over that row's scalar fields and returns `[]` instead of the parsed suggestions.

This logic was moved verbatim from the now-deleted `ArtifactGenerationService.suggest_reports`, so it predates #1239 — the fold was behavior-preserving and gated on `scripts/audit_public_api_compat.py`, so the heuristic could not be changed there.

## Proposed fix (from CodeRabbit)

Only unwrap when the outer list is a single nested list:

```python
items = result
if (
    len(result) == 1
    and isinstance(result[0], list)
    and (not result[0] or isinstance(result[0][0], list))
):
    items = result[0]
for item in items:
    ...
```

## Acceptance

- Add a regression test covering both the wrapped (`[[row, row]]`) and already-flat (`[row, row]`) GET_SUGGESTED_REPORTS shapes, asserting both parse to the same suggestions.
- Confirm no change to the `client.artifacts.suggest_reports` public signature.

teng-lin:
Fixes #1243

## Problem

`ArtifactsAPI.suggest_reports` (`src/notebooklm/_artifacts.py`) unwrapped the `GET_SUGGESTED_REPORTS` response with `result[0] if isinstance(result[0], list) else result`. `result[0]` is a list for both the wrapped shape (`[[row1, row2]]`) and the already-flat shape (`[row1, row2]`). In the **flat** case, `items` became the first suggestion row (`row1`), so the loop iterated over that row's scalar fields, found nothing matching `len(item) >= 5`, and returned `[]`.

## Verification (this is a real bug)

- The VCR cassette `tests/cassettes/artifacts_suggest_reports.yaml` decodes to the **wrapped** shape `[[row1..row4]]` (rows of 6 fields) — the current code already handles this correctly.
- The golden fixture `tests/fixtures/rpc_golden/GET_SUGGESTED_REPORTS.json` decodes to the **flat** shape `[row1, row2]` — the current code returns `[]` for this, dropping all suggestions.

Both shapes occur in practice, so the heuristic must handle both.

## Fix

Only unwrap when the outer list is a single element whose first inner element is itself a list (a row):

```python
items = result
if (
    len(result) == 1
    and isinstance(result[0], list)
    and (not result[0] or isinstance(result[0][0], list))
):
    items = result[0]
```

The wrapped shape unwraps; the flat shape (including a single flat suggestion) is left intact. The public signature of `suggest_reports` is unchanged, and `scripts/audit_public_api_compat.py` reports no new break.

## Tests

Adds `TestSuggestReportsUnwrap` covering the wrapped, flat, single-flat, and empty shapes, and asserting the wrapped and flat shapes parse to identical suggestions.

## Gates

- `uv run ruff check .` / `uv run ruff format .` — clean
- `uv run mypy src/notebooklm --ignore-missing-imports` — clean
- `uv run python scripts/audit_public_api_compat.py` — no new public break
- `uv run pytest -q -k "suggest or report or artifact"` — 1007 passed, 7 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of suggested-reports responses that can arrive in different nested shapes, preventing incorrect unwrapping and ensuring consistent report suggestions.

* **Tests**
  * Added comprehensive tests covering multiple response shapes, single-row and empty cases, and a regression case to prevent future breakage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/1244?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/notebooklm/_artifacts.py
+++ b/src/notebooklm/_artifacts.py
@@ -1001,7 +1001,19 @@ class ArtifactsAPI:
 
         suggestions = []
         if result and isinstance(result, list) and len(result) > 0:
-            items = result[0] if isinstance(result[0], list) else result
+            # GET_SUGGESTED_REPORTS returns either a wrapped list of rows
+            # (``[[row1, row2, ...]]``) or an already-flat list of rows
+            # (``[row1, row2, ...]``). Only unwrap the wrapped case, detected
+            # by a single outer element whose first inner element is itself a
+            # list (a row). Unwrapping the flat case would mistake the first
+            # row's scalar fields for the suggestion rows and return nothing.
+            items = result
+            if (
+                len(result) == 1
+                and isinstance(result[0], list)
+                and (not result[0] or isinstance(result[0][0], list))
+            ):
+                items = result[0]
             for item in items:
                 if isinstance(item, list) and len(item) >= 5:
                     suggestions.append(
--- a/tests/unit/test_artifacts_coverage.py
+++ b/tests/unit/test_artifacts_coverage.py
@@ -1084,9 +1084,29 @@ class TestSuggestReportsUnwrap:
         assert suggestions[0].prompt == "Write a briefing."
 
     @pytest.mark.asyncio
+    async def test_wrapped_single_suggestion_parses(self, mock_artifacts_api):
+        """A wrapped single row (``[[row]]``) unwraps to one suggestion."""
+        api, mock_core = mock_artifacts_api
+        mock_core.rpc_executor.rpc_call.return_value = [[list(self._ROWS[0])]]
+
+        suggestions = await api.suggest_reports("nb_123")
+
+        assert len(suggestions) == 1
+        assert suggestions[0].title == "Briefing Doc"
+        assert suggestions[0].prompt == "Write a briefing."
+
+    @pytest.mark.asyncio
     async def test_empty_result_returns_empty(self, mock_artifacts_api):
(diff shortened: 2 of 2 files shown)


teng-lin:
@claude review

teng-lin:
Thanks @claude — addressed the non-blocking note in commit 95200a40: added `test_wrapped_single_suggestion_parses` (`[[row]]`) plus a `test_wrapped_empty_returns_empty` (`[[]]`) regression to harden the unwrap heuristic. The gemini/coderabbit/cubic passes had no actionable threads.

@claude review

Issue #1263, opened 2026-05-30
Test infra: can't record VCR cassettes under pytest — autouse NOTEBOOKLM_HOME isolation hides real auth


teng-lin:
## Problem

VCR cassettes cannot be **recorded** through the normal pytest path. The autouse fixture `_isolate_notebooklm_home` (`tests/conftest.py:16`) pins `NOTEBOOKLM_HOME` to a per-test tmp dir for **every** test:

```python
@pytest.fixture(autouse=True)
def _isolate_notebooklm_home(request, tmp_path, monkeypatch):
    monkeypatch.setenv("NOTEBOOKLM_HOME", str(tmp_path / "notebooklm-home"))
```

That isolation is correct for replay/unit runs (deterministic empty storage), but when recording (`NOTEBOOKLM_VCR_RECORD=1`) the client/CLI then looks for auth in the empty tmp home instead of the real `~/.notebooklm/profiles/<profile>/storage_state.json`, so it can't authenticate against the live API. The result: recording a new cassette under pytest fails on auth.

## Current workaround

Recording is done with a **hand-rolled standalone script** outside pytest that sets `NOTEBOOKLM_VCR_RECORD=1`, leaves `NOTEBOOKLM_HOME` at the real location, drives the CLI/client under `notebooklm_vcr.use_cassette(...)`, then relies on the scrub hooks. This is how `tests/cassettes/mind_maps_interactive.yaml` and `tests/cassettes/generate_mind_map_interactive.yaml` were captured.

This works but is undocumented, easy to get wrong, and skips the safety rails (forgetting to verify `is_clean` / run the sanitizer gate before committing is exactly the kind of slip that has leaked credentials before).

## Why it matters

- Contributors can't follow the documented "set `NOTEBOOKLM_VCR_RECORD=1` and run the test" flow in `tests/vcr_config.py` to (re)record cassettes for new RPC flows.
- The standalone-script pattern is reinvented per recording and is error-prone around scrubbing/auth.

## Proposed fix (pick one)

1. **Make the isolation fixture defer to the real home when recording.** When `NOTEBOOKLM_VCR_RECORD` is truthy (and/or the test carries an opt-in marker like `@pytest.mark.vcr_record`), `_isolate_notebooklm_home` should *not* override `NOTEBOOKLM_HOME`, so the live-auth path works. Replay/unit runs are unchanged.
[...]

PR #1267, opened 2026-05-30
test: apply #1265 review nits (casefold + ImportError guards) (#1263)

teng-lin:
Follow-up to #1265, which was squash-merged at its original tip before two addressed-feedback commits landed on the branch. This lands those two reviewer nits on `main`:

- **`.casefold()` over `.lower()`** (Gemini + claude[bot]): record-mode env parsing now uses `.casefold()` for Unicode-aware case folding, per the project's general string-comparison rule. Applied to **both** `tests/conftest.py:_vcr_recording()` and the canonical `tests/vcr_config.py:_is_vcr_record_mode()` so the two stay byte-for-byte identical — the `test_vcr_recording_matches_canonical` parity test (incl. padded `" 1"`/`"1 "`) enforces this. No `.strip()` is added to either side; single-sided stripping is the half-recording divergence #1263 guards against.
- **`raise ImportError` instead of bare `assert`** (claude[bot]): the two module-level importlib spec/loader guards in `tests/unit/test_home_isolation.py` now raise `ImportError` with a message, so the diagnostic survives `python -O` and is clearer if the file-path loading ever breaks.

Test-infra only; no `src/` changes. Full suite green (7782 passed, 52 skipped), ruff + mypy clean.

## Summary by CodeRabbit

* **Tests**
  * Improved test robustness by making environment-variable parsing case-insensitive and more reliable.
  * Replaced assert-based module-loading guards with explicit error handling to fail fast and provide clearer failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/teng-lin/notebooklm-py/pull/1267?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,15 +16,15 @@ _PLAYWRIGHT_INSTALLED = importlib.util.find_spec("playwright") is not None
 # Mirror of ``tests/vcr_config._is_vcr_record_mode`` — duplicated (not imported)
 # so the *root* conftest, loaded for every test, stays free of the heavier
 # ``vcr_config`` (vcrpy) import. Kept byte-for-byte identical to the canonical
-# (no ``.strip()``) so the two never disagree on a padded value and split the
-# config into a half-recording state; ``test_home_isolation`` pins the parity.
-# (#1263)
+# (same ``.casefold()``, no ``.strip()``) so the two never disagree on a padded
+# value and split the config into a half-recording state; ``test_home_isolation``
+# pins the parity. (#1263)
 _VCR_RECORD_ENV = "NOTEBOOKLM_VCR_RECORD"
 
 
 def _vcr_recording() -> bool:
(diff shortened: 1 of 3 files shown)


teng-lin:
@claude review

teng-lin:
Thanks @claude — addressed both out-of-scope observations:

1. **`_load_sibling()` bare `assert`** — folded into this PR (7576f121): same fix-class as the PR and in a file already being edited, so I converted it to `raise ImportError` for a consistent `python -O`-safe diagnostic across `vcr_config.py`.
2. **`get_error_injection_mode()` `.lower()` → `.casefold()`** — that's a separate env var (`NOTEBOOKLM_VCR_RECORD_ERRORS`) on its own code path (with its own legitimate `.strip()`) plus a mirror in `notebooklm._error_injection`, so I kept it out of this tight scope and filed #1268 to track it.

Full suite green locally (7782 passed, 52 skipped), ruff + mypy clean.

Issue #1346, opened 2026-06-01
ADR-0019: library-wide error-and-return contract (v0.8.0)


teng-lin:
Umbrella for the SDK-wide error-and-return contract.

- ADR: `docs/adr/0019-error-and-return-contract.md`
- Executable plan (v2.3): `.sisyphus/plans/2026-06-01-error-contract-migration.md`

**Principle:** a return value encodes only success + genuine async-lifecycle state; *resource* absence, server refusal, and shape-drift **raise**; `None`/`""`/empty-sentinel/`"not_found"`-string/`ValueError` never signal an error. Derived reads do **not** police parent existence (uniform-empty). The additive half shipped in **v0.7.0** — including the `NOTEBOOKLM_FUTURE_ERRORS` opt-in flag (#1406, extended to the behavioral changes by #1405) that runs the v0.8.0 contract early for forward-compat testing, plus the `docs/upgrading-to-0.8.0.md` migration guide. The breaking half converges in **v0.8.0** (staging **decided 2026-06-04: one go** — all breaks ship together; #1344 not split to 0.9.0).

### Status (single source of truth — edit this table, no comments)

| Item | Scope | Status |
|------|-------|--------|
| #1247 | `get()` → raise `*NotFoundError` (sources/notes/artifacts/mind_maps) | 🔲 v0.8.0 — flag-previewable (#1406); unlocks the #1343 teeth |
| #1342 | async-kickoff refusal → raise (+ `research.start` drift, + `with_rate_limit_retry` returned-status removal) | 🔲 v0.8.0 — flag-previewable (#1405) |
| #1362 | mutate-existing fail-loud: `notes.update` + `rename(return_object=False)` | 🔲 v0.8.0 — flag-previewable (#1405) |
| #1290 | `refresh`/`delete_conversation` `bool`-always-`True` → `None` (`clear_cache` kept — its bool is meaningful) | 🔲 v0.8.0 — flag-previewable (#1405) |
| #1344 | derived-read drift → `DecodingError` + lister drift-tightening | 🔲 v0.8.0 — ships with the one-go break-set (**not** split to 0.9.0; decided 2026-06-04) |
| #1251 | remove `MappingCompat` dict-subscript | 🔲 v0.8.0 — flag-previewable (#1406) |
| #1254 | remove `interval` kwarg alias | 🔲 v0.8.0 — flag-previewable (#1406) |
[...]

Issue #1344, opened 2026-06-01
v0.8.0: uniform-empty derived reads + typed not-found sentinels + lister drift-tightening


teng-lin:
Part of ADR-0019 (plan Wave E + A2a).

- [ ] **Derived reads don't police parent existence (uniform-empty):** `get_summary`/`get_description`/`get_guide` return empty, `get_tree` returns `None`, on a missing parent; only shape-drift raises `DecodingError`. Drop `get_guide`'s source-existence preflight.
- [ ] `check_freshness` (`_sources.py:689`): keep the valid stale shape `[[null,false,...]]`→`False`; raise `DecodingError` only on a genuinely unrecognized shape.
- [ ] `research.poll`: add `ResearchStatus.NOT_FOUND` (distinct from `NO_RESEARCH`) for a pinned-but-absent task id; `poll_status` keeps typed `not_found`.
- [ ] **Lister drift-tightening (A2a):** raise `DecodingError` on non-list/malformed payloads — note lister (`_note_service.py:127`) + artifact non-list collapse (`_artifact/listing.py:111`) — so `get()`/`get_or_none` distinguish miss from drift.
- [ ] Decide separately: the artifact `except (RPCError, httpx.HTTPError) → []` resilience swallow (`_artifact/listing.py:127`) — drift-collapse vs deliberate partial-availability.

`get_tree`/`rename`/`delete` interpret `_detect_kind`'s `MindMapNotFoundError` per class (read→None, rename→raise, delete→None) — see #1291.

teng-lin:
Shipped in #1434 (merged as d79d9d82): derived-read + lister drift now raises DecodingError instead of silently collapsing to False/[].

PR #1352, opened 2026-06-02
feat(research): add ResearchStatus.NOT_FOUND sentinel for pinned-but-absent poll

teng-lin:
## What

Adds `ResearchStatus.NOT_FOUND` — a typed lifecycle sentinel for the
**poll-observed absence of a *specific* requested research task**, distinct
from the existing `NO_RESEARCH` ("nothing in flight").

`ResearchAPI.poll(notebook_id, task_id=...)` now returns
`ResearchTask.not_found(task_id)` (status `NOT_FOUND`, carrying the requested
id, empty `tasks`) when a **non-empty** pinned `task_id` matches no in-flight
task. The unfiltered `task_id=None` empty-poll path — and the degenerate
empty-string `task_id=""` — stay `NO_RESEARCH` with the exact legacy dict shape
`{"status": "no_research", "tasks": []}`.

## Why

This is the **additive, strictly non-breaking** portion of the ADR-0019 Rule 4
not-found contract (`docs/adr/0019-error-and-return-contract.md`): a poll loop
cannot treat replication-lag as exceptional, so a poll-observed absent task is a
**typed sentinel, not a raise** — and it is categorically different from
"nothing in flight". The contract calls out a "**new** `ResearchStatus.NOT_FOUND`
member (distinct from the existing `NO_RESEARCH`)".

Implements the `research.poll` checkbox of #1344 (umbrella ADR-0019 work #1346).
The terminal-answer escalation for a task that never appears stays in
`wait_for_completion`, not here — `poll` is the stateless primitive.

## Non-breaking guarantees

- The enum is extended only; the str-enum keeps full back-compat
  (`NOT_FOUND == "not_found"`, `isinstance(..., str)`). The public-API
  compatibility audit passes with no new allowlist entry.
- The `task_id=None` empty-poll path is byte-for-byte unchanged (`NO_RESEARCH`,
  same dict shape) — guarded by tests.
- `task_id=""` (degenerate) also stays `NO_RESEARCH`, so no existing dict shape
  changes for it.
- `wait_for_completion` is untouched: it drives `_select_polled_tasks` directly
  (not `poll`), so `NOT_FOUND` never leaks into the wait loop.
- The `poll` never raises for an absent task.

## Tests

- `poll` with a pinned-but-absent `task_id` (both a populated poll with a
  non-matching task, and an empty envelope) → `NOT_FOUND` carrying the requested id.
- `poll` with no `task_id` on an empty poll → still `NO_RESEARCH`.
- `poll(task_id="")` → still `NO_RESEARCH` (legacy dict shape preserved).
- Enum-level: `NOT_FOUND == "not_found"`, distinct from `NO_RESEARCH`;
  `ResearchTask.not_found()` sentinel shape.

Full suite: 7971 passed, 95 skipped. ruff format/lint, mypy, and the
public-API compat audit all green.

Refs #1344, #1346.

## Summary by CodeRabbit

* **New Features**
  * Added a distinct NOT_FOUND status returned when polling a specific task_id that has no matching in-flight task.

* **Documentation**
  * Clarified API docs to describe the typed NOT_FOUND outcome, its difference from an unfiltered empty poll, and that wait-for-completion never returns NOT_FOUND for a pinned task.

* **Tests**
  * Added unit tests verifying NOT_FOUND behavior and compatibility with existing public output shapes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `except NoteError` / `except MindMapError` call sites. These are the
   prerequisite for the mind-map not-found work (ADR-0019; issues #1291, #1346).
   No method raises them yet — this change only adds the types.
+- `ResearchStatus.NOT_FOUND` — a typed lifecycle sentinel for the
+  poll-observed absence of a *specific* requested research task, distinct from
+  `NO_RESEARCH` ("nothing in flight"). `research.poll(notebook_id, task_id=...)`
+  now returns `ResearchTask.not_found(task_id)` (status `NOT_FOUND`, carrying
+  the requested id) when a non-empty pinned `task_id` matches no in-flight task;
+  the unfiltered `task_id=None` empty poll still returns `NO_RESEARCH`
+  unchanged. Additive and non-breaking — the poll never raises for an absent
+  task (ADR-0019 Rule 4; issues #1344, #1346).
 
 ### Changed
 
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -1467,9 +1467,12 @@ async def poll(notebook_id: str, task_id: str | None = None) -> ResearchTask:
     """
     Returns a ResearchTask for the selected research task. If task_id is None,
     selects the latest research task and emits an ambiguity warning if multiple
-    tasks are in-flight. Attributes:
+    tasks are in-flight. When task_id is supplied but no in-flight task matches,
+    returns ResearchTask.not_found(task_id) — status NOT_FOUND, a typed
+    poll-observed-absence sentinel (does not raise); the unfiltered empty poll
+    stays NO_RESEARCH. Attributes:
       - task_id:   str            — task/report identifier
-      - status:    ResearchStatus — COMPLETED | FAILED | IN_PROGRESS | NO_RESEARCH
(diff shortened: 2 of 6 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the reviews. Addressed the take-or-leave nits from the claude[bot] review in 50d34569:

- **Nit #1 (unreachable ambiguity warning):** `poll()` now gates the warning on `warn_on_ambiguous=task_id is None`, matching `wait_for_completion`. Behavior-identical (the warn branch was already unreachable once a discriminator is pinned), but the intent is now explicit.
- **Nit #2 (wait_for_completion docstring):** added a one-liner stating `wait_for_completion` never returns `NOT_FOUND` — a pinned task temporarily absent from a poll is treated as transient replication-lag and keeps polling.
- **Suggested regression test:** added `test_wait_for_completion_never_returns_not_found_for_pinned_task`, which asserts an early empty poll for a pinned task resolves to `completed` (never `NOT_FOUND`), turning the insulation guarantee from code-inspection into a behavioral guard.

cubic and gemini reported no issues. The earlier CHANGELOG conflict with #1349 was a clean rebase onto latest `main` (both entries kept). Full suite green (ruff/mypy/pytest + public-API compat audit).

teng-lin:
@coderabbitai Addressed in 9136706f: `test_poll_pinned_absent_task_id_empty_response_returns_not_found` now also asserts `result.tasks == ()`, pinning the empty-sibling-tasks half of the sentinel contract on the empty-envelope path (so a regression returning stale/non-empty tasks for an absent pinned task would be caught). Good catch.

Issue #1389, opened 2026-06-03
arch(burndown): drain positional-RPC indexing allowlist behind safe_index / row adapters


teng-lin:
## Burndown: migrate raw positional-RPC decoding behind `_row_adapters/` + `safe_index`

Follow-up to #1377. The durable GATE (`tests/_lint/test_no_raw_positional_rpc_indexing.py`) now forbids raw *chained* positional indexing of RPC payloads (`x[i][j]`) outside `rpc/` and `_row_adapters/`. The current offenders are baselined into that lint's `ALLOWLIST`. This issue tracks draining that allowlist — migrating each file behind `rpc/_safe_index.safe_index()` and/or a typed `_row_adapters/` view so genuine shape drift **raises** `UnknownRPCMethodError` (ADR-011) instead of silently degrading to empty/wrong data.

**The allowlist to drain** (each removal re-arms the gate for that file via the lint's `test_no_stale_allowlist_entries`):

- [ ] `_chat/wire.py` — `parse_citations` / `parse_single_citation` (`first[2][0]`, `cite[0][0]`, `first[4][-1]`). NOTE: every chained descent here is already guarded by an `isinstance`/`len` check where the absent case is *legitimate* (non-answer / non-citation records). Migration must preserve "no citations present ⇒ `[]`" and only raise on descents that are genuinely committed; treat with care.
- [ ] `_source/content.py` — `get_guide` / `get_fulltext` positional descents into `GET_SOURCE_GUIDE` / `GET_SOURCE`.
- [ ] `_research_task_parser.py` — residual `src[1][0]` / `src[1][1]` / `result[0][0]`; this module already uses `safe_index` for the task-id/info descents, so it is the most precedented next target.
- [ ] `_mind_maps_api.py` — `_new_artifact_id` (`create_response[0][0]`).
- [ ] `_note_service.py` — note-id / row-content descents (`item[1][0]`, `result[0][0]`).
- [ ] `_settings.py` — NOT caught by the gate (it walks `result[idx]` in a loop via `_extract_nested_value`, not chained literal subscripts), but listed in #1377 as an ad-hoc positional parser; converge it onto `safe_index` for the same drift-raises guarantee.
- [ ] `_artifacts.py` — `result[0][0]` shape probe + `params[2][2]` (request-side).
- [ ] `_artifact/listing.py` — `result[0][0]` shape probe.
- [ ] `_research.py` — `result[0][0]` / `src_data[0][0]` source descents.
[...]

PR #1398, opened 2026-06-03
refactor(settings): route language parser through safe_index per-site (closes #1395)

### LINKED ISSUE #1395: arch: converge _settings.py positional parser onto safe_index (residual from #1389)
## Context

Follow-up residual from #1389 (positional-RPC decode burndown).

#1389 drained the `tests/_lint/test_no_raw_positional_rpc_indexing.py` `ALLOWLIST` to empty — every file the AST gate flagged (chained `x[i][j]` literal subscripts) is now migrated behind a bound-local single-level read or a `_row_adapters/` view.

`_settings.py` was listed on the #1389 checklist but is **NOT caught by that gate**: its `_extract_nested_value` / `_extract_nested_list` walk a `Sequence[int]` path in a `for idx in path: result = result[idx]` loop, not literal chained subscripts. So it never appeared on the lint allowlist and draining the allowlist did not require touching it.

It was deliberately **left unconverted** in the #1389 PR for a correctness reason (ADR-011):

- `_extract_nested_value(..., _SET_LANGUAGE_PATH)` / `_GET_SETTINGS_PATH` resolve the **optional** current-language slot. A user who never set a language legitimately has a short/absent path, and the parser must return `None` (not raise).
- A blanket conversion to `safe_index` would make that legitimately-absent language path **raise** `UnknownRPCMethodError`, regressing the optional-language contract — exactly the "blanket swallow→raise is unsafe" hazard #1389 warns about.

## Task

Converge `_settings.py` onto `rpc/_safe_index.safe_index()` **per-site**, preserving the legitimate-absent semantics:

- The language-slot reads must still degrade to `None` when the slot is legitimately absent/empty (user has no language set).
- Only a genuine **shape drift** (the path exists but the structure moved) should raise.

This likely needs the parser to distinguish "path terminates early because the optional slot is absent" from "path terminates early because the response shape drifted" — which is not cleanly expressible with the current single path-walk helper. Evaluate whether that distinction is even observable from the wire shape before committing to a raise; if it is not, the right outcome may be to leave `_settings.py` on its current bounded `try/except → None` form and document why (ADR-011 legitimate-absent), rather than force a risky conversion.

## Acceptance

- `_settings.py` either routes through `safe_index` with the optional-language slot still returning `None`, or carries a documented ADR-011 rationale for staying on its current form.
- Unit tests prove BOTH: legitimate-absent language ⇒ `None`; genuine shape drift ⇒ raises (if conversion lands).
- `uv run pytest tests/unit/test_user_settings_api.py` green.

Relates to ADR-011, #1377, #1389.

teng-lin:
## What

Converges `_settings.py`'s output-language parser onto `rpc/_safe_index.safe_index()` **per-site**, splitting the descent into a mandatory envelope prefix (raises on drift) and an optional language tail (degrades to `None`). Replaces the legacy `_extract_nested_value` blanket `for idx in path: result = result[idx]` + `try/except -> None` loop with a regime-aware `_extract_language` helper.

## Why

`_settings.py` was on the #1389 (positional-RPC decode burndown) checklist but the #1389 AST gate never flagged it: its parser walked a `Sequence[int]` path in a loop, not literal chained subscripts. It was deliberately left unconverted because a **blanket** `safe_index` conversion would make the legitimately-absent language slot **raise**, regressing the optional-language contract (a user who never set a language must get `None`).

The issue is an **investigate-then-decide** task. The decision: **convert per-site** (the issue's "if distinguishable" branch), because the wire evidence shows the descent is cleanly splittable.

### Wire-shape evidence (`tests/cassettes/settings_*`)

```
GET_USER_SETTINGS inner: [[null,[..limits..],[true,null,null,true,["fr"]],[[1]],[true,1,3,2]]]
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ flags block      -> language at [0][2][4][0]
SET_USER_SETTINGS inner:  [null,[..limits..],[true,null,null,true,["en"]],[[1]],[true,1,3,2]]
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ flags block      -> language at [2][4][0]
```

- The **settings-flags block** (GET `result[0][2]` / SET `result[2]`) is structurally mandatory in every healthy response — drift there is genuine schema drift and is **observable/distinguishable**.
- The **language slot** (`[4]`, then the `["code"]` unwrap `[0]`) is a **routinely-optional trailing element**. A user with no language has it empty/absent — indistinguishable from drift *at that exact position* — so it must degrade to `None`.

The original author documented exactly this: *"The API can return an empty string for the language code, which we treat as None."*

## How

`_extract_language(data, required_prefix, optional_tail, *, method_id, source)`:
1. `required_prefix` (GET `(0, 2)` / SET `(2,)`) descends the mandatory envelope via `safe_index` -> raises `UnknownRPCMethodError` on genuine drift (ADR-011).
2. `optional_tail` (`(4, 0)`) walks the optional language slot with a plain bounded guard -> degrades to `None` on absence/empty/non-list.

This mirrors the established `_notebooks._extract_suggested_topics` precedent (`safe_index` for mandatory descent + plain guard for a routinely-optional trailing slot).

**Behavior preservation:** the legitimate-absent language path returns `None` exactly as today. The *only* behavior change is that genuine envelope drift now raises instead of silently returning `None` — precisely the ADR-011 intent ("empty payload still produces an empty value; drift now produces an exception").

## Verification

- `uv run pytest tests/unit/test_user_settings_api.py` — green (41 passed); new tests prove **both** contracts: legitimate-absent (5 absence shapes) -> `None`; envelope drift (4 shapes) -> raises, plus end-to-end `get`/`set` cases.
- Integration: the pre-existing `..._returns_none_on_malformed_response` test (fed a flags-block-less envelope, asserted `None`) is split into a legitimate-trailing-absent -> `None` case and an envelope-drift -> raises case in `tests/integration/test_settings_integration.py`.
- `uv run mypy src/notebooklm --ignore-missing-imports` — clean (196 files).
- `uv run ruff format --check` + `ruff check` — clean.
- `tests/_lint/test_module_size_ratchet.py` — `_settings.py` at 309 LOC, well under the 900 budget.
- Full `tests/unit` + `tests/integration` + `tests/_lint` suites green; all 155 language/settings-related tests pass.

Closes #1395.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for extracting the configured output language and for handling missing or truncated language slots.
  * Added integration and unit tests to validate correct behavior when response envelopes are malformed or incomplete.

* **Refactor**
  * More robust handling of output-language retrieval: returns the detected language code, degrades to None when legitimately absent, and surfaces errors on unexpected response shape.

### DIFF
--- a/src/notebooklm/_settings.py
+++ b/src/notebooklm/_settings.py
@@ -89,7 +89,11 @@ def _extract_language(
     block = safe_index(data, *required_prefix, method_id=method_id, source=source)
     result: Any = block
     for idx in optional_tail:
-        if not isinstance(result, list) or idx >= len(result):
+        # Bound-check both ends: ``idx >= len`` guards the trailing-optional
+        # absence; ``idx < 0`` guards against a negative index silently
+        # wrapping to Python's from-the-end semantics for any future caller
+        # (today's tails are hardcoded non-negative tuples).
+        if not isinstance(result, list) or not 0 <= idx < len(result):
             return None
         result = result[idx]
     return result or None
--- a/src/notebooklm/_settings.py
+++ b/src/notebooklm/_settings.py
@@ -5,7 +5,7 @@ from collections.abc import Sequence
 from typing import Any
 
 from ._runtime.contracts import RpcCaller
-from .rpc import RPCMethod
+from .rpc import RPCMethod, safe_index
 from .types import AccountLimits, AccountTier
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the review! Addressed both non-blocking observations in 374b3154:

- **Negative-index guard** in the `optional_tail` loop — now bound-checks both ends (`0 <= idx < len(result)`), so a future caller passing a negative tail index degrades to `None` rather than wrapping to Python's from-the-end semantics. Pinned by `test_extract_language_tail_negative_index_returns_none`.
- **SET-prefix drift coverage** — added `test_extract_language_set_prefix_envelope_drift_raises` (4 shapes) directly exercising the shorter `(2,)` prefix, matching the existing GET-prefix `(0, 2)` parametrized drift test.

Unit tests now 46 passed; mypy + ruff clean.

Issue #1362, opened 2026-06-02
v0.8.0: complete mutate-existing fail-loud — notes.update + rename(return_object=False) raise on a missing target


teng-lin:
ADR-0019 Class 5 (*mutate existing → target missing → raise `*NotFoundError`*). Two sub-items were never carried into a tracked v0.8.0 task — surfaced as orphans during the v0.8.0 breakdown (the "no orphans" discipline / #1330). #1255 covered only the `return_object=True` / fail-loud-on-hydrate half (shipped in 0.7.0); these are the remainder.

### (a) `notes.update` silently no-ops on a missing note
`NotesAPI.update` → `_note_service.update_note` uses `UPDATE_NOTE` with `allow_null=True` and returns nothing inspectable, so updating a non-existent note silently "succeeds." Per Class 5 it should **raise `NoteNotFoundError`** (the class now exists, shipped in #1349). Detect at the **public facade** (`NotesAPI.update` via `await self.get(...)` / `get_or_none`), not in the low-level `_note_service` (layer boundary). **Breaking.**

### (b) `rename(return_object=False)` skips miss-detection (sources + artifacts)
Today `return_object=False` short-circuits without detecting a missing target (silent `None`). The flag should control **hydration/return**, not **miss-detection** — a missing target must **raise `*NotFoundError` in both modes**. Add the lightweight existence preflight on the `False` path: `_get_or_none` for `sources`, the studio-only lookup for `artifacts`. **Breaking.** (`mind_maps.rename` already fails loud in both modes — #1291/#1353 — so this is sources/artifacts only.)

### Ship with the gate (per #1330)
Add a **parity test**: every mutate-existing op (`rename`/`update`/`configure`, both `return_object` modes) raises `*NotFoundError` on a missing target. That makes the Class-5 contract a CI invariant, not a remembered convention.

Part of #1346 (ADR-0019). Related: #1247 (sibling not-found flips), #1342, ADR-0019, #1330. Supersedes the deferred remainder of #1255 (closed).

teng-lin:
Shipped in #1438 (merged as 3a0781d0): mutate-existing fail-loud — notes.update + rename(return_object=False) raise on a miss.

Issue #1342, opened 2026-06-01
v0.8.0: synchronous generation refusal raises (drop status='failed' suppression)


teng-lin:
Part of the ADR-0019 error-and-return contract convergence (umbrella to follow). See `docs/adr/0019-error-and-return-contract.md` and `.sisyphus/plans/2026-06-01-error-contract-migration.md` (Wave C).

Async-kickoff ops currently **swallow** a synchronous `USER_DISPLAYABLE_ERROR` into `GenerationStatus(status="failed")`. Per ADR-0019, a refusal is *couldn't-start* and must **raise** the `RateLimitError`/`RPCError` the decoder already produces.

- [ ] Delete the two suppression branches: `_call_generate` (`_artifacts.py:1150-1158`) + `revise_slide` (`:541-549`) — re-raise.
- [ ] Kill the 2nd `status="failed"` overload: `_parse_generation_result` missing-id branch (`:1251-1260`) → raise `ArtifactFeatureUnavailableError` (null) / `DecodingError` (falsey id).
- [ ] `research.start` empty/non-list payload (`_research.py:391`) → raise `DecodingError`.
- [ ] **Rewrite `with_rate_limit_retry`** (`artifacts.py:64`) to `except RateLimitError` (it inspects a *returned* status today) — else CLI rate-limit retry breaks.
- [ ] `retry_failed` (#1319) ships born-correct under this contract; `RETRY_ARTIFACT` golden fixture + `NON_IDEMPOTENT_NO_RETRY` (keep `CREATE_ARTIFACT`/`GENERATE_MIND_MAP` as `PROBE_THEN_CREATE`).

Related: #1319.

teng-lin:
Shipped in #1438 (merged as 3a0781d0): synchronous generation-kickoff refusal now raises (no more status='failed' suppression; with_rate_limit_retry re-raises RateLimitError).

Issue #1319, opened 2026-05-31
API: expose Retry for failed generated artifacts


alexxety:
## Summary

NotebookLM web UI exposes a `Retry` action when a generated Studio artifact fails, but the Python API does not appear to expose an equivalent failed-artifact retry operation.

This is a problem for automation that keeps one canonical notebook per content pack. When a video/audio/infographic artifact fails, a human can open NotebookLM and click `Retry` on the failed artifact. A Python API user currently has to approximate this by deleting the failed artifact and calling `generate_*()` again, but that is not clearly equivalent to the UI retry behavior and can be brittle.

## Current behavior

The library exposes generation primitives such as `client.artifacts.generate_video(...)` and artifact deletion via `client.artifacts.delete(notebook_id, artifact_id)`.

I could not find a public API method that maps to the UI `Retry` action for an already-created failed artifact, for example something like:

```python
await client.artifacts.retry(notebook_id, artifact_id)
# or
await client.artifacts.retry_video(notebook_id, artifact_id)
```

Related issue #887 covered rate-limit retry while starting generation. This request is different: the generation has already produced a failed artifact in the notebook, and the UI exposes a per-artifact retry button.

## Expected behavior

Please expose a supported API for retrying a failed generated artifact in-place, preserving the same notebook and source context.

A good API would make it possible to:

- detect that an artifact is retryable;
- call the same retry operation the NotebookLM UI uses;
- return a `GenerationStatus` or task id that can be polled/downloaded through the existing artifact flow;
- fail loudly if the artifact is not retryable or if NotebookLM rejects the retry.

## Why delete + generate is not enough

Delete + generate is only a workaround. It may not match UI semantics if NotebookLM treats retry differently from a fresh generation request. It also forces automation authors to guess whether the failed artifact should be deleted first, whether deletion is required, and which generation parameters should be resent.

## Acceptance idea

- Add a public retry method for failed generated artifacts, or document the officially supported equivalent if one already exists internally.
- Add at least one test/example showing retry of a failed video artifact.
- Document the behavior for non-retryable artifacts.

alexxety:
I did a sanitized DevTools capture of the NotebookLM web UI `Retry` button for a failed Studio artifact.

Observed behavior, with credentials/headers/cookies intentionally omitted:

- UI action maps to private RPC id: `Rytqqe`
- source path: `/notebook/{notebook_id}`
- request params shape: `[retry_options, artifact_id]`
[...]

Issue #1247, opened 2026-05-30
v0.8.0: flip sources/artifacts/notes/mind_maps.get() to raise *NotFoundError


teng-lin:
## Context

Tracking the breaking flip that the get()-returns-None deprecation layer (#1206 / its PR) sets up.

As of the deprecation layer (warning introduced in **v0.7.0**), `sources.get()`, `artifacts.get()`, and `notes.get()` still **return `None`** on a missing entity but now emit a `DeprecationWarning` on the miss. `notebooks.get()` already **raises** `NotebookNotFoundError`.

`mind_maps.get()` is the **fifth** `get()` and is folded in here — see "mind_maps (folded in from #1291)" below.

## The flip (v0.8.0 — BREAKING)

Replace the warn-and-return-`None` in each `get()` method with `raise <Resource>NotFoundError`, unifying the not-found contract across all `get()` methods:

- `src/notebooklm/_sources.py` — `SourcesAPI.get` → raise `SourceNotFoundError`
- `src/notebooklm/_artifacts.py` — `ArtifactsAPI.get` → raise `ArtifactNotFoundError`
- `src/notebooklm/_notes.py` — `NotesAPI.get` → raise `NoteNotFoundError` (a `NoteNotFoundError` class may need adding to `exceptions.py`; `notes` currently has no dedicated not-found type)
- `src/notebooklm/_mind_maps_api.py` — `MindMapsAPI.get` → raise `MindMapNotFoundError` (see prerequisites below)

The private `_get_or_none()` helper introduced by the deprecation layer holds the current `None`-returning body and can stay as the internal optional-lookup path. Decide whether to expose an optional-lookup public method (the `get_or_none` public method was rejected for the deprecation PR; revisit here).

## mind_maps (folded in from #1291)

`mind_maps.get()` (`_mind_maps_api.py:179`) currently returns `None` on a miss **without** emitting a `DeprecationWarning` — it never entered the v0.7.0 warn rollout. Folding the flip in here (rather than leaving it for a later minor) avoids shipping v0.8.0 with four `get()` methods raising and `mind_maps` alone returning `None` — i.e. spending a breaking window to manufacture a fresh inconsistency.

Because it's on the freshly-released 0.7.x line, it can still get the same warn→flip runway as its siblings if the warn shim lands in a 0.7.x patch.

[...]

PR #1413, opened 2026-06-04
feat: extend NOTEBOOKLM_FUTURE_ERRORS to behavioral v0.8.0 previews (closes #1405)

### LINKED ISSUE #1290: API: three methods return `bool` that is always `True` (refresh / delete_conversation / clear_cache)
Surfaced by an API-design audit. Low severity; filing for a deliberate decision rather than a rushed breaking change.

### Finding
Three public methods are annotated `-> bool` but the value carries no information:

- `SourcesAPI.refresh` (`src/notebooklm/_sources.py:670`) — `return True` unconditionally; any failure raises first.
- `ChatAPI.delete_conversation` (`src/notebooklm/_chat.py:708`) — `return True` unconditionally.
- `ChatAPI.clear_cache` (`src/notebooklm/_chat.py:750`) — **caveat:** returns `self._cache.clear(conversation_id)`, *not* a literal `True`. Whether this is genuinely informative depends on `_ConversationCache.clear()`'s return; it may already convey "something was cleared." This one needs verification before any change, and it interacts with `history --clear --json` / `cache_size()`.

This mirrors the `delete() -> None` cleanup already shipped in the 0.7.0 breaking line — a `-> bool` that can only be `True` is a misleading signal.

### Why not auto-fixed
Changing `-> bool` to `-> None` is a **breaking** return-type change (callers doing `if await client.sources.refresh(...)` would flip behavior, since `None` is falsy). Given it's cosmetic/low-value and `clear_cache` doesn't cleanly qualify, it deserves a deliberate decision rather than being bundled into an unrelated PR.

### Proposal
- `refresh` and `delete_conversation` → `-> None` (clean, unambiguous), batched into the 0.7.0 breaking surface with a CHANGELOG entry + api-compat-allowlist entries.
- `clear_cache` → verify `_cache.clear()`'s contract first; keep `-> bool` if the value is meaningful, else align with the above.

_Filed from an automated codebase audit (2026-05-31)._

### LINKED ISSUE #1405: Extend NOTEBOOKLM_FUTURE_ERRORS to the behavioral v0.8.0 changes (#1290/#1342/#1362)
## Context

The `NOTEBOOKLM_FUTURE_ERRORS` opt-in preview flag (added in #1406, additive in v0.7.0) lets callers run the v0.8.0 error contract (ADR-0019 / umbrella #1346) early to test forward-compatibility. It currently scopes to the **three flips that already have a clean warn-runway**:

- `get()` → raises `*NotFoundError` on a miss (#1247)
- `MappingCompatMixin` dict-subscript → raises `TypeError` (#1251)
- `deprecated_kwarg` deprecated keyword → raises `TypeError` (#1254)

## Follow-up

Extend the flag to the **purely-behavioral** v0.8.0 changes that were deliberately left out of the initial PR because they lack a clean warn-runway and their target behavior is more involved:

- [ ] **#1290** — `delete()` returning `bool` → `None` (the additive half already shipped; the flag should preview the bool-return removal where one remains).
- [ ] **#1342** — refusal-suppression behavior.
- [ ] **#1362** — fail-loud listing change.

Each needs its v0.8.0 behavior pinned down before `future_errors_enabled()` can gate it. As each one's target behavior is defined, route it through the same gate (`notebooklm._deprecation.future_errors_enabled`) and extend the both-modes conformance coverage in `tests/unit/test_public_api_behavior.py` / `tests/unit/test_future_errors_flag.py`.

This is a tracking issue only — no behavior change is owed until the v0.8.0 flips themselves land.

teng-lin:
## What

Extends the existing `NOTEBOOKLM_FUTURE_ERRORS` opt-in preview flag to the three **purely-behavioral** v0.8.0 changes that lack a deprecation warn-runway: #1290 (bool→None returns), #1342 (synchronous refusal raises), and #1362 (mutate-existing fail-loud). Each is gated `if future_errors_enabled(): <v0.8.0 behavior> else: <v0.7.0 behavior>`.

- **#1290** — `SourcesAPI.refresh` and `ChatAPI.delete_conversation` return `None` under the flag (the `True` they return today is uninformative — any failure raises first). The `-> bool` annotation is preserved; it flips to `-> None` at v0.8.0. `ChatAPI.clear_cache` is **not** gated — its bool is meaningful (`_ConversationCache.clear()` returns `False` when the id is absent).
- **#1342** — a synchronous generation refusal **raises** the decoder's `RateLimitError` / `RPCError` / `DecodingError` / `ArtifactFeatureUnavailableError` instead of being swallowed into `GenerationStatus(status="failed")` / returned `None`. Gated in `_call_generate`, `revise_slide`, `_parse_generation_result` (null id → `ArtifactFeatureUnavailableError`, falsey id → `DecodingError`), and `research.start` (empty/non-list payload **or falsey `task_id`** → `DecodingError`). `with_rate_limit_retry` already retries on both a returned rate-limited `GenerationStatus` (flag-off) and a raised `RateLimitError` (#1319), so the flag-on retry path is preserved in both modes.
- **#1362** — `NotesAPI.update` preflights at the public facade (`get_or_none`) and raises `NoteNotFoundError` on a miss; `sources`/`artifacts` `rename(return_object=False)` detect a missing target and raise `*NotFoundError` under the flag while still returning `None`. The flag gates **miss-detection**, not the return.

New both-modes test classes in `tests/unit/test_future_errors_flag.py` (one per gated behavior, each asserting flag-OFF = v0.7.0 and flag-ON = v0.8.0). `future_errors_enabled()` docstring, the `_deprecation`/`exceptions` notes, `CHANGELOG.md`, and `docs/deprecations.md` updated.

## Why

ADR-0019 (umbrella #1346) converges the public error/return contract in v0.8.0. The preview flag lets callers and CI test forward-compatibility **today**. Previously it covered only the three warn-runway flips (#1247/#1251/#1254); the three behavioral changes were deliberately left out pending their target behavior being pinned down (#1405).

**Preview-only invariant (load-bearing):** these are runtime-only previews. **No public return annotation changes** (which would trip the api-compat-allowlist + conformance gates), no default is flipped, and the default (flag-off) path is **byte-identical** to current v0.7.0. This PR does **not** close #1290/#1342/#1362 — their default flips ship in v0.8.0; it closes the tracking issue #1405.

Verified: `mypy` clean, full `pytest` green flag-off, the api-compat audit + conformance + v0.8.0-coverage gates pass, and the artifact/research/notes/source/chat suites exercised with `NOTEBOOKLM_FUTURE_ERRORS=1` to prove the flag-on paths.

## Summary by CodeRabbit

* **Documentation**
  * Expanded docs to clarify the NOTEBOOKLM_FUTURE_ERRORS preview, enumerating runtime-only v0.8.0 behavior changes.

* **New Features**
  * With the flag enabled: some boolean-returning ops now yield None; generation refusals raise underlying errors; mutate-existing ops raise NotFoundError on missing targets.

* **Tests**
  * Added extensive tests covering the behavioral previews gated by the future-errors flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -180,11 +180,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (a runway raises regardless of quiet). The four `get()` methods are now routed
   through a single `_lookup.resolve_get` bridge, eliminating the hand-duplicated
   warn-on-miss pattern. Helper: `notebooklm._deprecation.future_errors_enabled`.
-  The purely-behavioral v0.8.0 changes that lack a warn-runway (`delete()`
-  returning `None`, refusal-suppression, fail-loud listing) are not gated yet;
-  they will be folded in as their behavior is defined. Does **not** close
-  #1247/#1251/#1254 — the runways remain until the v0.8.0 flip. See
-  `docs/deprecations.md`. Additive (issue #1346).
+  The flag now **also** previews the purely-behavioral v0.8.0 changes that have
+  no warn-runway (#1405): the uninformative `bool` returns of `sources.refresh`
+  and `chat.delete_conversation` become `None` (#1290); a synchronous generation
+  refusal **raises** the decoder's `RateLimitError` / `RPCError` /
+  `DecodingError` / `ArtifactFeatureUnavailableError` instead of being swallowed
+  into `GenerationStatus(status="failed")` / returned `None` — across
+  `_call_generate`, `revise_slide`, `_parse_generation_result`, and
+  `research.start` (#1342); and the mutate-existing ops `notes.update` and
+  `sources`/`artifacts` `rename(return_object=False)` fail loud with a
+  `*NotFoundError` on a missing target (#1362). These previews are runtime-only —
+  **no public return annotation changes** until the v0.8.0 flip — so default-off
+  stays byte-identical. Does **not** close #1247/#1251/#1254/#1290/#1342/#1362 —
+  the runways and current behavior remain until the v0.8.0 flip. See
+  `docs/deprecations.md`. Additive (issues #1346, #1405).
 - `client.artifacts.retry_failed(notebook_id, artifact_id)` — retry a failed
   Studio artifact in place (the web UI "Retry" action), via the new
   `RETRY_ARTIFACT` (`Rytqqe`) RPC. The artifact is not deleted first and the
--- a/docs/deprecations.md
+++ b/docs/deprecations.md
@@ -39,13 +39,18 @@ behavior today**, so you can verify your code is forward-compatible before the
 flip ships. It is **off by default**, and default-off is byte-identical to
 current v0.7.0 behavior.
 
-When the flag is on, these three runways adopt their v0.8.0 behavior:
+When the flag is on, these runways adopt their v0.8.0 behavior. The first three
+are deprecation *warn-runways* (the warning becomes a raise); the last three are
+purely-behavioral previews with no warning today (#1405):
 
 | Runway | v0.7.0 (default / flag off) | With `NOTEBOOKLM_FUTURE_ERRORS=1` | Tracked by |
 |--------|-----------------------------|-----------------------------------|------------|
 | `sources.get()` / `artifacts.get()` / `notes.get()` / `mind_maps.get()` on a miss | Warns, returns `None` | Raises the matching `*NotFoundError` (`SourceNotFoundError` / `ArtifactNotFoundError` / `NoteNotFoundError` / `MindMapNotFoundError`) | [#1247](https://github.com/teng-lin/notebooklm-py/issues/1247) |
 | Dict-subscript `result["key"]` on the typed research / mind-map / source-guide returns (`MappingCompatMixin`) | Warns, returns the legacy dict value | Raises `TypeError: '<Type>' object is not subscriptable` (the same error a plain dataclass raises once the mixin is removed) | [#1251](https://github.com/teng-lin/notebooklm-py/issues/1251) |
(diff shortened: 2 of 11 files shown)


teng-lin:
@claude review

teng-lin:
@coderabbitai Addressed in 445c89b0: the `NoteNotFoundError` docstring now names both v0.7.0 preview raise sites — `notes.get` (via `resolve_get`, #1247) and `notes.update` (#1362) — instead of only `update`. Thanks for catching the understated scope.

teng-lin:
Re: cubic finding on `exceptions.py:1340` — fixed in 445c89b0. The `NoteNotFoundError` docstring now documents both preview raise sites (`notes.get` via `resolve_get` #1247, and `notes.update` #1362), generalizing the wording so the exception's scope isn't understated. (identified by cubic)

teng-lin:
Thanks for the thorough review @claude — addressed the actionable nits in 445c89b0:

1. **`_sources.rename` dense condition** — kept the inline form (the named-local extraction would push `_sources.py` over its ADR-0008 module-size ceiling) but clarified the comment so the two orthogonal short-circuits (truthy echo proves existence; flag-off keeps the v0.7.0 no-detect path) are explicit. Same readability goal, no ceiling cost.
2. **`_parse_generation_result` message** — the `DecodingError` already records `method_id` as a structured field on the exception; left the message text as-is since the field carries it.
3. **`notes.update` instance-patching** — intentional: `NotesAPI` has no ADR-007 forbidden-monkeypatch lint constraint (that lint targets top-level `notebooklm._*` private *module* paths + `monkeypatch.setattr` string targets, not instance-attribute assignment), so the direct `api.get_or_none = AsyncMock(...)` is sanctioned here, matching the `test_get_or_none.py` idiom.
4. **`with_rate_limit_retry` both-modes invariant** — already CI-enforced (not new to this PR): `tests/unit/test_artifact_rate_limit_retry.py` covers both `test_retries_rate_limited_status_then_returns_success` (returned status, flag-off) and `test_retries_raised_rate_limit_error_then_returns_success` (raised `RateLimitError`, #1319), so the flag-on raise path is exercised.

PR #1423, opened 2026-06-04
docs(artifacts): complete the type-4 variant taxonomy — include interactive mind map (variant 4)

teng-lin:
## What

Complete the artifact **type-4 variant taxonomy** in the describing comments / docstrings / docs. Several of these framed artifact type code 4 as the quiz/flashcard family but omitted the **interactive mind map (variant 4)**, which is also type 4. Updated, comments/docstrings/docs only:

- `src/notebooklm/_types/artifacts.py` — `ArtifactType` class docstring, `_map_artifact_kind` `variant` arg doc + the type-4 dispatch comment, and the unknown-variant warning string (`Unknown QUIZ variant {variant}.` → `Unknown type-4 (quiz / flashcards / mind-map) variant {variant}.`).
- `src/notebooklm/_row_adapters/artifacts.py` — the `[9]` options-block field comment, the `variant` property docstring, and the raw-type-match `Note:` docstring.
- `docs/python-api.md` — `_variant` field comment now `(1=flashcards, 2=quiz, 4=interactive mind map)`.
- `docs/rpc-reference.md` — the variant comments in the quiz/flashcard CREATE-payload examples now also list `4=interactive mind map` (each example keeps its own variant value).

The canonical block at `rpc/types.py:144-151` (already correct) is the source of wording and was left untouched.

## Why

Type code 4 is a **three-variant family** multiplexed via the variant at `data[9][1][0]`: `FLASHCARDS_VARIANT=1`, `QUIZ_VARIANT=2`, `INTERACTIVE_MIND_MAP_VARIANT=4`. The code already handles all three (`_map_artifact_kind`, `is_interactive_mind_map`, `is_unclassified_type4`); only the describing comments lagged, framing type 4 as just quiz/flashcard. This is a **comments-only** alignment with the canonical taxonomy — **zero behavior change**. No test asserts the old warning text. The genuinely quiz/flashcard-*specific* download/format paths (which use a separate extraction path for the mind map) were intentionally left as-is.

## Summary by CodeRabbit

* **Documentation**
  * Clarified artifact type documentation across API and SDK references to note that the type-4 family includes multiple variants (quizzes, flashcards, and interactive mind maps).
  * Rephrased variant explanations and inline comments so variant 4 is explicitly described as an interactive mind map option.
  * Updated notes and warnings to generically reference “type-4 (quiz / flashcards / mind-map)” for consistency.

### DIFF
--- a/docs/python-api.md
+++ b/docs/python-api.md
@@ -1890,7 +1890,7 @@ class Artifact:
     status: int                     # 1=processing, 2=pending, 3=completed, 4=failed
     created_at: Optional[datetime]
     url: Optional[str]
-    _variant: int | None = None     # Internal variant for type-4 artifacts (1=flashcards, 2=quiz).
+    _variant: int | None = None     # Internal variant for type-4 artifacts (1=flashcards, 2=quiz, 4=interactive mind map).
 
     @property
     def kind(self) -> ArtifactType:
--- a/docs/rpc-reference.md
+++ b/docs/rpc-reference.md
@@ -717,7 +717,7 @@ params = [
         [
             None,
             [
-                2,                    # Variant: 2=quiz, 1=flashcards
+                2,                    # Variant: 2=quiz, 1=flashcards, 4=interactive mind map
                 None,
                 instructions,
                 None,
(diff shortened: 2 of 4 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the review. Addressed both cosmetic nits in 63b8c6db:
- Dropped the redundant trailing clause in the type-4-family resolve comment (`_types/artifacts.py`).
- Reworded the position-9 table comment to `distinguish among QUIZ, FLASHCARDS, and the interactive mind map` to fix the `from … from` awkwardness (`_row_adapters/artifacts.py`).

Comments-only; zero behavior change.

Issue #1365, opened 2026-06-02
v0.8.0: delete dead deprecation machinery (_deprecation.py) after the removals land


teng-lin:
After the v0.8.0 removals land, the `src/notebooklm/_deprecation.py` helpers go dead — but the deletion is spread across #1247/#1251/#1254 and nobody owns the final sweep, so v0.8.0 risks shipping live-but-unreachable deprecation plumbing.

### Goes dead
- `warn_get_returns_none` + its 4 call sites (`_artifacts.py`, `_notes.py`, `_sources.py`, `_mind_maps_api.py`) — dead after #1247.
- `MappingCompatMixin` — dead after #1251.
- `deprecated_kwarg` — dead after #1254 (verify no other live caller first).

### Fix
- Remove the now-unreachable helpers and the corresponding parts of `tests/unit/test_deprecation_helper.py`, keeping only what still has a live caller.
- Keep `warn_get_returns_none` only if a sanctioned `get_or_none`-warning path still needs it (it does not after #1247).
- Sequence **after** #1247/#1251/#1254 merge.

Part of the v0.8.0 error-contract program (#1346 / ADR-0019).

PR #1436, opened 2026-06-05
feat(api)!: get() raises *NotFoundError on a miss (#1247)

teng-lin:
## What

`client.{sources,artifacts,notes,mind_maps}.get(notebook_id, id)` now **raises**
its matching `*NotFoundError` on a genuine miss instead of returning `None` (with
a `DeprecationWarning`), unifying the not-found contract with `notebooks.get()`,
which already raises (#1247). Return annotations narrow `X | None` → `X`.

```python
# Before: returned None on a miss (deprecated)
src = await client.sources.get(nb, source_id)   # None if missing

# After: raises on a miss
src = await client.sources.get(nb, source_id)   # SourceNotFoundError if missing

# Unchanged escape hatch — the sanctioned None-on-miss lookup (never warns):
src = await client.sources.get_or_none(nb, source_id)
```

## Why

ADR-0019 makes resource absence an exception on `get` and reserves `None`-on-miss
for an explicit `get_or_none`. This lands the v0.8.0 flip the v0.7.0
`DeprecationWarning` runway pointed at.

## How

- **Single-sourced:** `_lookup.resolve_get` now delegates to the existing
  `unwrap_or_raise` and always raises on a miss — the `future_errors_enabled()`
  gate and the `warn_get_returns_none(...)` call are gone (so `_lookup.py` no
  longer imports either). The four namespace `get()` bodies were already routed
  through `resolve_get`; they only needed the annotation narrowing + docstring
  update + dropping the now-unused `resource=` arg.
- **The teeth:** the api-compat comparator is return-annotation-blind, so the
  flip is enforced by the conformance gates — `GET_OPTIONAL_EXEMPTIONS` empties
  (`test_public_api_contract.py`) and the four `get_warns` rows flip to `False`
  (`test_public_api_behavior.py::TestGetMissContract`, which then asserts raising
  under both flag modes). The `#1247` runway row is removed from the v0.8.0
  breaking-changes registry (its runway cited the `warn_get_returns_none` usage
  this removes). No api-compat allowlist entry is needed (a comparator-blind
  value flip).

## Tests

- Inverted `test_get_returns_none_deprecation` (warn+None → raises) and trimmed
  the now-removed flag-gated `resolve_get` cases in `test_future_errors_flag`
  (resolve_get's raise-on-miss is flag-agnostic now; the full miss-contract is in
  `test_public_api_behavior::TestGetMissContract`).
- Full suite green; mypy + ruff clean. No internal caller relied on the `None`
  return (internal code uses `get_or_none` / `_get_or_none`).

Note: `resolve_get` is now a one-line delegate to `unwrap_or_raise`; its outright
deletion (and the dead `warn_get_returns_none` def) is tracked for the v0.8.0
capstone (#1365) where the deprecation machinery is removed wholesale.

## Summary by CodeRabbit

* **Breaking Changes**
  * Public get() calls (sources, artifacts, notes, mind_maps) now raise their matching NotFoundError on missing items instead of returning None — update callers to handle exceptions.
  * A deprecated positional-parser public member has been removed.

* **Tests**
  * Test suite updated to expect exception-on-miss behavior across unit, integration, and end-to-end tests.

### DIFF
--- a/scripts/api-compat-allowlist.json
+++ b/scripts/api-compat-allowlist.json
@@ -65,7 +65,7 @@
     {
       "code": "removed-member",
       "object": "notebooklm.Note.from_api_response",
-      "reason": "Dead positional parser removed (#1389). Note rows are decoded through notebooklm._row_adapters.notes.NoteRow; the classmethod had no production callers (only test_types.py exercised it). Construct Note(...) directly from NoteRow's named properties — see NotesAPI._parse_note."
+      "reason": "Dead positional parser removed (#1389). Note rows are decoded through notebooklm._row_adapters.notes.NoteRow; the classmethod had no production callers (only test_types.py exercised it). Construct Note(...) directly from NoteRow's named properties \u2014 see NotesAPI._parse_note."
     },
     {
       "code": "removed-member",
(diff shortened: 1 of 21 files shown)


teng-lin:
@claude review

teng-lin:
Thanks @claude — addressed all three doc-only findings in 94614b6a: the four `get_or_none` docstrings and the three `_get_or_none` alias comments now read past-tense (`get()` **now raises** `*NotFoundError`, not "slated to raise in v0.8.0"), and `test_get_returns_none_deprecation.py`'s module docstring describes the flipped (raise) contract. The separate `docs/deprecations.md` / `CLAUDE.md` rows are deliberately deferred to the v0.8.0 capstone (#1365), which finalizes all docs + deletes the dead `warn_get_returns_none` / `resolve_get` machinery in one pass.

Issue #1204, opened 2026-05-29
test_add_epub_file flakes nightly E2E when NotebookLM's server-side EPUB processing is down


teng-lin:
## Summary

`tests/e2e/test_file_upload.py::TestFileUpload::test_add_epub_file` fails whenever NotebookLM's backend cannot process an uploaded EPUB, surfacing as `notebooklm.exceptions.SourceProcessingError: Source <id> failed to process` (server sets `SourceStatus.ERROR`). This has been flaking the **Nightly E2E** workflow on `main` (failing 2026-05-25, 05-26, 05-29; passing 05-27, 05-28).

## Root-cause investigation (during v0.6.0 release prep, 2026-05-29)

Live-API probes against the same account:

| Probe | Content-type on wire | Result |
|---|---|---|
| Synthetic minimal EPUB — `release/v0.6.0` | `application/epub+zip` | `SourceStatus.ERROR` |
| Synthetic minimal EPUB — **v0.5.0** (released, known-good) | `application/epub+zip` | `SourceStatus.ERROR` |
| **Real** EPUB (Project Gutenberg #11, 137 KB) — v0.6.0 | `application/epub+zip` | `SourceStatus.ERROR` |
| Real EPUB, **explicit** `mime_type="application/epub+zip"` | `application/epub+zip` | `SourceStatus.ERROR` |
| Plain `.txt` — v0.6.0 | `text/plain` | **SUCCESS** |

Conclusions:
- **Not a client regression** — v0.5.0 fails identically.
- **Not a fixture-quality issue** — a real, well-formed EPUB fails the same way.
- **Not a MIME issue** — correct `application/epub+zip` is sent (both extension-inferred and explicit); server still returns `SourceStatus.ERROR`.
- **Not account/pipeline throttling** — `.txt` uploads and processes fine concurrently.

NotebookLM's server-side EPUB ingestion is intermittently failing (Google-side). The client uploads correctly and faithfully raises on the server's terminal error; the test just has no tolerance for a transient server-side processing outage.

## Proposed fix

Make `test_add_epub_file` resilient to a server-side processing failure outside our control, e.g. `xfail(raises=SourceProcessingError, strict=False)` or skip/xfail specifically when the terminal status is `SourceStatus.ERROR` within `wait_timeout`. Keep the upload/registration + `kind == SourceType.EPUB` assertions strict so a genuine client-side EPUB regression (bad MIME routing, wrong disposition, upload corruption) still fails.

[...]

PR #1451, opened 2026-06-05
test(e2e): extract assert_file_source helper for file-upload tests

teng-lin:
## What
Extracts a shared `assert_file_source(...)` helper and routes all 9 file-upload E2E tests through it, applying the #1204 registration-split (client contract vs server processing) consistently instead of 9 near-identical inline bodies.

## Behaviour by type
- **documents/images** (`pdf`, `text`, `markdown`, `csv`, `docx`, `jpg`, `epub`) — **full round-trip**: `add_file(wait=False)` registration must succeed (a client regression fails here, distinctly ours), then `wait_until_ready` asserts `kind`; a terminal `SourceStatus.ERROR` fails **loudly** with a triage note (chaining the original `SourceProcessingError`) rather than an opaque red.
- **dummy media** (`mp3`, `mp4`) — registration-only via `process=False` (a dummy media file can't transcribe). Unchanged behaviour, just routed through the helper.

## Notes
- **Behaviour-preserving:** `add_file(wait=True)` already calls `wait_until_ready` internally, so the split changes structure/attribution, not what's verified. The EPUB test keeps the loud-split from #1450 (now via the helper). `expected_title` is asserted only where the original did (the `tempfile`-based text/markdown tests don't have a clean name).
- **Media fixtures deferred:** producing real, processable `.mp3`/`.mp4` is a separate problem (no TTS here; ffmpeg isn't guaranteed in the nightly runners; non-speech audio may not transcribe) — left for a follow-up, so those two stay registration-only for now.
- Test-only; no `src` change, 0 api-compat allowlist entries.

## Summary by CodeRabbit

* **Tests**
  * Improved test infrastructure for file upload validation with a centralized helper function, ensuring consistent testing across multiple file formats (PDF, text, Markdown, CSV, MP3, MP4, DOCX, JPG, EPUB).

**Note:** This release contains no end-user visible changes.

### DIFF
--- a/tests/e2e/test_file_upload.py
+++ b/tests/e2e/test_file_upload.py
@@ -502,6 +502,57 @@ def create_minimal_epub(path: Path) -> None:
     path.write_bytes(buffer.getvalue())
 
 
+async def assert_file_source(
+    client,
+    notebook_id,
+    path,
+    *,
+    expected_kind,
+    mime_type=None,
+    expected_title=None,
+    process=True,
+    timeout=120.0,
+):
+    """Upload ``path``, then assert the client contract and (optionally) processing.
+
+    Splits the two concerns so a failure is attributable (see #1204):
+
+    - **Client contract** (always): ``add_file(wait=False)`` must return a source
+      with an id — and ``expected_title`` when given. A client-side regression (bad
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude review

Issue #1452, opened 2026-06-05
Strengthen CLI integration testing with VCR (cli_vcr): depth, error-contract, breadth, re-record resilience


teng-lin:
# #1452 — Strengthen cli_vcr testing (reality-based, post-inventory redesign)

## Inventory reality (the basis for this redesign — all verified)
- **105 cassettes recorded against 15 distinct notebooks.** Input ids are **decoupled** from
  cassettes: the matcher (`tests/vcr_config.py` `_rpcids_matcher` + `_freq_body_matcher`)
  keys on the RPC method id + body *shape*, NOT the notebook/source id. `mock_context`
  injects one id (`c3f6285f`) regardless of which of the 15 notebooks a cassette was
  recorded against. **Proof:** `test_artifacts` loads `artifacts_list_flashcards.yaml`
  (recorded vs `167481cd`) while passing `c3f6285f` and replays fine. "One canonical
  fixture notebook" was never the reality and is not needed.
- **No dead cassettes** (all 105 referenced; `real_api_*` belong to `test_vcr_real_api.py`).
- **List responses are huge/volatile** (`real_api_list_notebooks.yaml` = 596 UUIDs) → exact
  counts/contents are unassertable.
- **Input-echo is safe:** 3 existing asserts compare `output["notebook_id"]` to the *passed*
  placeholder `b8d6f2a1` (in 0 cassettes) — the CLI echoes the input id into its own result,
  so this holds regardless of the cassette and survives any re-record.

## Design principle (yours)
Generating all cassettes from one notebook is unrealistic; we build around the 15-notebook
reality and use what we have. **Every assertion must survive a re-record that uses a
DIFFERENT notebook with different data.** Some breadth gaps genuinely need extra recording —
that's fine; the assertions are notebook-agnostic, so new cassettes drop in without fixture
surgery. A re-record needs an assertion change ONLY when the response *shape* actually changed
(a real signal worth catching).

## Re-record-safe assertion tiers (the allowed vocabulary)
1. **Schema** — `--json` envelope field names + types.
2. **Invariants** — each id UUID-shaped, each title non-empty, **`count > 0`** (never `== N`).
3. **Cross-render** — text vs `--json` of the SAME cassette agree (row count / id set).
4. **Filter correctness** — `artifact list --type audio` ⇒ every item is audio (decoder logic).
5. **Input-echo** — a mutation's output id == the placeholder the test passed (CLI threads it).
[...]

PR #1464, opened 2026-06-06
test(rpc): pinned + mutation golden-payload tests for adapter field-position (#1452)

teng-lin:
## What & why

Final layer of the #1452 deep-assert work. The cli_vcr cassette-derived deep asserts (PR-B, merged in #1462) catch fabrication, drop, and miscount — but they **cannot catch field-position errors**. A `title`↔`url` swap, a wrong nesting depth, or a `type`↔`status` confusion all keep the right *count* of values, so a per-field equality on a recorded cassette would still pass while the wrong slot was read.

Closing that residual requires **PINNED synthetic-row tests at the adapter unit level**, where the exact input is controlled. Wrong nesting / wrong field-index is this repo's #1 historical breakage class (source id `[id]` vs `[[[[id]]]]`, url precedence, etc. — see CLAUDE.md "Common Pitfalls").

These are unit tests over synthetic arrays (not cli_vcr cassettes), so the re-record-safety lint does not apply and pinned literals are correct here. **Test-only — no `src/` changes.**

## Deliverable 1 — PINNED field-position ground truth

Every slot carries a distinct, self-identifying literal (`id="ID_AT_PLAIN_0"`, `url="...AT_7_0"`, ...) so a wrong-slot read produces an obviously-wrong value:

- **Source id nesting**: plain `[id]`, bare-string (flat shape), drive-backed `[None, True, [id]]` at `[2][0]`, plus medium- and deeply-nested dispatch (the `[[[[id]]]]` fallback).
- **Source URL precedence + YouTube fallback**: canonical `metadata[7]` > youtube `metadata[5]` > bare `metadata[0]` (deeply-nested only); youtube non-string first element rejected.
- **Source kind + status**: type code at `metadata[4]` → `SourceType`; status block at `raw[3][1]` → `SourceStatus` (with the `[3][0]` decoy ignored).
- **Artifact variant + media-url positions**: variant at `[9][1][0]` (flashcards/quiz/interactive-mind-map); audio `[6][5]` mp4-preference; slide-deck pdf `[16][3]` / pptx `[16][4]`; infographic content scan.
- **Note current vs deleted vs legacy shapes** + `Artifact.from_mind_map` current/deleted.

## Deliverable 2 — MUTATION teeth-proof

Each takes a correct row, moves one value (or shifts a nesting depth), and asserts the adapter reacts — "if the decoder regressed to read the now-mutated wrong slot, THIS test would fail":

- **Source title↔url swap** → the title planted in the url slot must NOT parse as a URL.
- **Source id-nesting shift** → too-deep `[[id]]` / drive-shifted id is not silently mis-extracted (stringifies the wrong envelope / resolves to `""`).
- **Source + Artifact type↔status swap** (parametrized correct-vs-mutated pairs) → decoded enums/codes flip accordingly.
- **Artifact variant-nesting shift** → strict-mode `UnknownRPCMethodError` (a regression reading `[9][1]` directly would silently return the value instead of raising).
- **Slide-deck pdf↔pptx slot swap** → the wrong-slot read is visible.
- **Note deleted-sentinel shift** + **inner content↔title swap**.

## Already covered (intentionally not duplicated)

Position-constant pins (`_URL_POS == 7`, ...) and the soft-degrade / strict-mode edges live in `tests/unit/test_row_adapters.py` (`TestPositionContract`, `TestNoteRow*`, ...). This PR adds the **planted-distinct-value + mutation-pair** angle those constant pins do not: a constant pin proves the index value, but not that a real title and a real url cannot be confused at decode time.

## Verification

- `uv run pytest tests/unit/test_rpc_golden_payloads.py -q` → 251 passed, 41 skipped (was 211 / 41 — 40 new tests)
- Full `uv run pytest` → 8361 passed, 131 skipped, 1 xfailed
- `ruff format --check`, `ruff check`, `mypy src/notebooklm` + the changed test file → all clean
- `scripts/audit_public_api_compat.py` → exit 0, 0 new allowlist entries
- No real decoder bug found while writing these.

## Summary by CodeRabbit

* **Tests**
  * Added extensive unit tests to validate robust handling and extraction of source, artifact, and note data: source ID nesting, URL-resolution precedence, source type→kind and status decoding, artifact variant and media-url selection, and note shape across legacy/current/deleted variants. Also added mutation tests that intentionally swap or shift fields to ensure adapters fail loudly or reflect changed values as expected.

### DIFF
--- a/tests/unit/test_rpc_golden_payloads.py
+++ b/tests/unit/test_rpc_golden_payloads.py
@@ -1293,7 +1293,7 @@ def _make_artifact_row(
     type_code: int = ArtifactTypeCode.AUDIO.value,
     status: int = ArtifactStatus.COMPLETED.value,
     variant: int | None = None,
-    audio_media: list[Any] | None = None,
+    audio_media_list: list[Any] | None = None,
     slide_pdf: Any = None,
     slide_pptx: Any = None,
 ) -> list[Any]:
@@ -1302,6 +1302,10 @@ def _make_artifact_row(
     Slots filled at their canonical ``ArtifactRow`` positions: id ``[0]``,
     title ``[1]``, type ``[2]``, status ``[4]``, variant ``[9][1][0]``, audio
     media list ``[6][5]``, slide-deck pdf ``[16][3]`` / pptx ``[16][4]``.
+
+    ``audio_media_list`` is the media list that lands at ``[6][5]`` (the inner
+    list of ``[url, kind, mime]`` entries), NOT the outer ``[6]`` audio-metadata
+    block — the factory wraps it under a fresh ``[6]`` envelope.
     """
     row: list[Any] = [None] * 19
     row[0] = artifact_id
@@ -1310,9 +1314,9 @@ def _make_artifact_row(
     row[4] = status
     if variant is not None:
         row[9] = [None, [variant]]
-    if audio_media is not None:
+    if audio_media_list is not None:
         audio_block: list[Any] = [None] * 6
-        audio_block[5] = audio_media
+        audio_block[5] = audio_media_list
         row[6] = audio_block
     if slide_pdf is not None or slide_pptx is not None:
         row[16] = [None, None, None, slide_pdf, slide_pptx]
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the review. Addressed the non-blocking notes in 799f5896 (test-only, no behavior change):

- **Note 2 (`audio_media` naming):** renamed the `_make_artifact_row` param to `audio_media_list` and documented that it lands at `[6][5]` (the inner media list), not the outer `[6]` block.
- **Note 1 (`startswith("http")` clarity):** added a comment that the assertion is a derived property of the planted title literal, not a probe of adapter URL-validation — the real teeth are the `mutated.url == title` line above it.
- **Note 3 (no infographic mutation test):** added an explanatory comment on the ground-truth test noting the accessor scans for the first url-bearing item rather than reading a single pinned slot, so there is no fixed index to shift for a slot-confusion mutation.

Note 4 (the pre-existing `cast` import) needs no change — it predates this PR and is still used at L117.

Issue #1488, opened 2026-06-09
Download cli_vcr tests give false coverage: download exits 1 (double LIST_ARTIFACTS), masked by assert_command_success accepting exit-1


teng-lin:
## Summary

The `download` `cli_vcr` integration tests pass while the download command **actually fails** (exit 1, no file written). They provide effectively **zero real coverage** — a complete download regression would not be caught. Two compounding causes: the download path issues `LIST_ARTIFACTS` **twice** (the cassettes record one), and `assert_command_success` accepts exit code 1 by default, swallowing the failure.

## Evidence

A strict probe (assert `exit_code == 0` **and** the output file exists) over the recorded cassettes, for `download quiz`/`report`/`data-table`:

```
command=download data-table   EXIT=1   FILE_WRITTEN=False   EXC=SystemExit(1)
Error: Can't overwrite existing cassette
('tests/cassettes/artifacts_download_data_table.yaml') in your current record mode ('none').
No match for the request (<Request (POST) ...rpcids=gArtLc...>)
rpc failed: RPC LIST_ARTIFACTS (CannotOverwriteExistingCassetteException)
```

All three exit 1 and write no file. The existing test (`tests/integration/cli_vcr/test_downloads.py::TestDownloadCommands::test_download`) calls `assert_command_success(result)` and passes regardless.

## Root cause

**1. The download path lists artifacts twice.**
`execute_download` lists artifacts to select the target:
- `src/notebooklm/_app/download.py:514` — `all_artifacts = await facade.artifacts.list(notebook_id)`

…then the per-type download method lists **again** to find the artifact:
- `src/notebooklm/_artifact/downloads.py:416` — `artifacts = await self._list_artifacts(notebook_id, list_type)` (→ a second `LIST_ARTIFACTS` / `gArtLc`)

The `artifacts_download_*.yaml` cassettes were recorded with **one** `gArtLc` interaction, so on replay the second call has no match → `CannotOverwriteExistingCassetteException` → the command aborts with exit 1. (In production against the real API the second call would succeed, so download *works* live — but it's a wasteful extra `LIST_ARTIFACTS` RPC per download, adding latency and quota.)

**2. The assertion helper masks the failure.**
`tests/integration/cli_vcr/conftest.py:146`:
```python
def assert_command_success(result, *, allow_no_context: bool = True) -> None:
    acceptable_codes = (0, 1) if allow_no_context else (0,)
[...]

Issue #1485, opened 2026-06-08
Notebook summary fails with safe_index drift on a summary-less (empty/new) notebook


teng-lin:
## Summary

Getting a notebook's summary fails with a `safe_index drift` RPC error when the notebook has **no summary yet** (e.g. a brand-new notebook with no sources). A legitimately-empty summary state is being mis-classified as wire-schema drift and raised, instead of being treated as "no description available."

## Symptom

```
RPC: safe_index drift at path ()[0] [method_id='VfAZjd', path=(), source='_notebooks._extract_summary', data_at_failure='None'] (retriable=false)
```

## Affected surfaces (shared layer)

All three go through `notebooklm._notebooks._extract_summary`:
- **CLI**: `notebooklm summary <empty-notebook>` (`src/notebooklm/cli/notebook_cmd.py:267`)
- **Python API**: `client.notebooks.get_summary(notebook_id)` (`src/notebooklm/_notebooks.py:572`)
- **MCP**: the `notebook_describe` tool

## Root cause

`_extract_summary` (`src/notebooklm/_notebooks.py:39-61`) *intends* to return `""` for a missing summary — its docstring says so, and it has an `if summary_val is None: return ""` guard. But that guard only catches a missing **inner** slot. For a summary-less notebook the SUMMARIZE `result[0]` payload (`outer`) is `None`, so the **outer** `[0]` descent inside

```python
summary_val = safe_index(
    outer, 0, 0,
    method_id=RPCMethod.SUMMARIZE.value,
    source="_notebooks._extract_summary",
)
```

treats indexing `None` at the root as schema **drift** and raises before the `None` guard is ever reached (`data_at_failure='None'` at `path=()[0]`).

An empty/absent summary is a routine, legitimate state — not a wire-shape change — so it should not surface as `safe_index drift`.

## Reproduction

```python
async with NotebookLMClient.from_storage() as client:
    nb = await client.notebooks.create("empty")     # no sources
    await client.notebooks.get_summary(nb.id)        # -> RPC safe_index drift
```
or via the CLI: create a fresh notebook, then `notebooklm summary <id>`.

Confirmed the inverse: `get_summary` / `notebook_describe` on **populated** notebooks returns the real AI summary correctly — so this is specific to the no-summary case, not broad drift.

## Proposed fix

[...]

Issue #1466, opened 2026-06-06
chat.ask TransportServerError timeout with byte-count mismatch (declared 151, actual 149) — v0.7.0 and v0.8.0


N4IRS:
Description:
    
    notebooklm ask consistently fails with a 30-second timeout on chat.ask. Earlier RPCs (GET_LAST_CONVERSATION_ID, GET_NOTEBOOK) succeed normally. Reproduced on two machines, two Python versions.
    
    Environment:
    - notebooklm-py v0.7.0 (PyPI) and v0.8.0 (git HEAD, 87d74c1)
    - Machine 1: Debian, system Python 3.13.5, venv Python 3.11.15
    - Machine 2: Python 3.12.3
    - Single Google account (no multi-account routing issue)
    - Shared notebook
    
    Steps to reproduce:
    bash
    notebooklm -vv ask --new --yes --notebook <id> "hello"
    
    
    Observed behavior:
    
    WARNING Byte-count mismatch (declared 151, actual 149 bytes); tolerated
    ...
    WARNING rpc failed: chat.ask (TransportServerError)
    WARNING chat.ask retry timed out after 30.0s
    Network error: chat.ask timed out after retries
    
    
    RPC sequence from -vv:
    
    GET_LAST_CONVERSATION_ID → 200 OK (0.27s)
    GET_NOTEBOOK → 200 OK (0.26s)
    chat.ask → TransportServerError after 30.0s
    
    
    The first chunk is consistently 2 bytes short (151 declared, 149 actual). notebooklm list and notebooklm auth check --test both pass.
    
    Expected behavior: chat.ask completes and returns an answer.
    
    Additional context: The batchexecute framing desync suggests Google may have changed method IDs or response formatting since the latest adapter update. The issue started today — was working previously.

teng-lin:
Thanks for the detailed report — the two-machine/two-Python repro and the RPC sequence were exactly what made this diagnosable. Root-caused and fixed in #1468.

## Root cause

It's **not** a batchexecute framing/method-ID change, and the byte-count mismatch is a red herring (more below). The real cause is a **timeout mismatch on the chat stream**:

- `chat.ask` hits the streaming endpoint `GenerateFreeFormStreamed`. On **shared notebooks**, the server spends tens of seconds on *time-to-first-byte* (resolving access/source context) before it starts streaming the answer.
[...]

PR #1495, opened 2026-06-09
Backport post-0.7.1 bug fixes to release/v0.7.1 (v0.7.2)

teng-lin:
Backports the two user-facing bug fixes that landed on `main` after the v0.7.1 release onto the `release/v0.7.1` line, and cuts version `0.7.2`.

### #1489 — empty notebook summary no longer raises `UnknownRPCMethodError` (#1485)
Clean cherry-pick of `5374411d`. A brand-new, source-less notebook has no summary yet, so `SUMMARIZE` returns an absent/`None` payload. `notebooks.get_summary()` / `get_description()` now treat that routine "no summary yet" state as `""` instead of mis-classifying it as wire-schema drift; genuinely-malformed payloads still raise. `safe_index` also now rejects a `str`/`bytes` value at an intermediate descent hop.

### #1490 — `download <type>` no longer exits 1 with no file written (#1488)
Manual port of `64ec8621`. The download executor listed artifacts twice — once to select the target, then each per-type `download_<x>` re-listed to re-find it by id — so the second `LIST_ARTIFACTS` could not replay against a single-interaction VCR cassette and aborted the download.

This bug exists on 0.7.x too; #1479 (not on this branch) only *relocated* the executor from `cli/services/download.py` to `_app/download.py`. The port therefore applies the **identical runtime-layer changes** as `main` and adapts only the executor to its 0.7.x home:

- `ArtifactListingService.list_artifacts_with_raw` returns `(typed, raw_studio_rows, mind_map_rows)` in one pass; `ArtifactsAPI` gains the `_list_for_download` seam.
- The executor lists once and threads the already-fetched rows into the bound `download_<x>` (which self-fetches only when the kwarg is `None` — standalone path unchanged). Studio downloads pass `spec.kind`, skipping the needless mind-map sub-fetch; mind-map partial-availability is preserved via a `None` sentinel.
- `cli_vcr` `assert_command_success` now defaults strict (exit 0); the download tests assert exit 0 **and** file written so a regression fails loud.

## Also

- Bumps version to `0.7.2` (pyproject + uv.lock).
- Records both fixes under a new `[0.7.2]` CHANGELOG section.
- Bumps the module-size ratchet to the measured post-fix 0.7.x LOC.

## Verification

- `mypy src/notebooklm --ignore-missing-imports`: clean (198 files)
- `ruff check` + `ruff format --check`: clean
- Full suite: **8354 passed, 130 skipped, 1 xfailed**

## Notes

Already-backported fixes (#1466 chat timeout, #1469 decoder warning) are present in the `v0.7.1` tag and excluded. The main-only `fix(audit)` (#1433) is CI infra, not user-facing, and excluded.

https://claude.ai/code/session_01TxizgdMxj575e4t6sSsWVd

---
_Generated by [Claude Code](https://claude.ai/code/session_01TxizgdMxj575e4t6sSsWVd)_

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,10 @@ name: Test
 
 on:
   push:
-    branches: [main]
+    branches: [main, "release/**"]
   pull_request:
-    branches: [main]
+    branches: [main, "release/**"]
+  workflow_dispatch: {}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.7.2] - 2026-06-09
+
+Maintenance patch on the 0.7.x line. Backports fixes from `main`
+(cherry-picked ahead of the v0.8.0 breaking release).
+
+### Fixed
+
+- **Empty notebook summary no longer raises `UnknownRPCMethodError`** (#1485).
+  A brand-new, source-less notebook has no summary yet, so the `SUMMARIZE` RPC
+  returns an absent/`None` payload. `notebooks.get_summary()` and
+  `notebooks.get_description()` now treat that routine "no summary yet" state as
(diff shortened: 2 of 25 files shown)


teng-lin:
@claude review

teng-lin:
@claude review

PR #1528, opened 2026-06-10
fix(types): decode created_at as tz-aware UTC, not naive host-local time

### LINKED ISSUE #1519: API timestamps decoded as naive local-time datetimes — created_at is host-TZ-dependent
**Severity:** Medium (correctness)

**Location:** `src/notebooklm/_types/common.py:150` — `return datetime.fromtimestamp(value)` (no `tz`).

**Bug:** Produces a naive datetime in the host's local zone. The public `created_at: datetime | None` contract therefore mis-states the absolute instant and renders host-TZ-dependently: a notebook created at 13:40:05 UTC serializes as `2026-01-13T08:40:05` (offset-less) under `America/New_York`. Reproduced empirically. No UTC normalization exists anywhere between decode and render.

**Triggers when:** any `notebooks.list/get`, `sources.list`, or `artifacts.list` on a non-UTC host, observable whenever `created_at` is serialized — CLI `--json` (`.isoformat()`), `_app`/MCP `serialize.py`, or human `.strftime`.

**Fix:** `return datetime.fromtimestamp(value, tz=timezone.utc)`. `.timestamp()` round-trips unchanged, so internal sort/dedup/download are provably unaffected; only the rendered string changes (becomes host-independent). Update the unit tests asserting naive expectations (e.g. `test_types.py` asserts `fromtimestamp` called without `tz`).

**Note:** This is the **production sibling** of the timezone bug fixed in the #1511 golden test — the test was made TZ-invariant by pinning epoch, but the production decoder still emits host-dependent strings.

---
*Filed from an adversarially-verified multi-agent code audit (each finding was confirmed against the source by an independent verifier). Severity is post-verification.*

teng-lin:
## Summary

Fixes #1519 (from the adversarially-verified audit). The shared decoder `_datetime_from_timestamp` (`src/notebooklm/_types/common.py`, behind `Notebook`/`Source`/`Artifact`/`MindMap.created_at`) called `datetime.fromtimestamp(value)` with **no `tz`** — producing a **naive** datetime in the host's local zone. The public `created_at: datetime | None` contract therefore mis-stated the absolute instant and rendered host-TZ-dependently: a notebook created `13:40:05` UTC serialized as the offset-less `08:40:05` under `America/New_York`, and `--json`/`isoformat`/`strftime` output varied per machine.

### Fix
`return datetime.fromtimestamp(value, tz=timezone.utc)` — tz-aware UTC, host-independent. `.timestamp()` round-trips identically, so internal sort/dedup/download ordering (which key on `.timestamp()` or the raw int) is provably unaffected; only the rendered string changes (now offset-aware, identical everywhere).

This is the **production sibling** of the timezone slip that was pinned out of the #1511 golden VCR test (that test was made TZ-invariant, but this production decoder still emitted host-dependent strings).

### Tests (red-first)
- tz-aware + host-independence unit tests (TZ-parametrized via `$TZ`/`time.tzset()`), proving the old naive behavior fails and the decoded instant is identical across zones.
- Updated the existing `fromtimestamp`-call-args assertion to expect `tz=utc`; hardened the #1511 golden VCR test with a `tzinfo` assertion.

### Verification
- Full `uv run pytest -m "not e2e"`: 9618+ passed (no failures)
- mypy clean · ruff clean · API-compat audit exit 0 (only rendered string changes; `.timestamp()` unchanged) · module-size ratchet green

Polish: claude implementer + codex independent review — codex SHIP. It independently confirmed this is the **only** production `fromtimestamp` site, found no naive/aware `created_at` comparison that would now raise, verified CLI/`_app.serialize`/MCP paths pin no offset-less strings, and that None/0/negative/huge timestamps keep their behavior. One Minor (CHANGELOG over-listed `Note.created_at`, which the decoder doesn't currently feed) fixed in the final commit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `created_at` timestamp handling to return timezone-aware UTC datetimes instead of host-local times. This ensures consistent temporal representation across different environments and machines, improving accuracy for serialization and formatting operations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Decoded `created_at` timestamps are now tz-aware UTC instead of naive
+  host-local time** (#1519). The shared decoder `_datetime_from_timestamp`
+  (backing `Notebook`/`Source`/`Artifact`/`MindMap.created_at`) called
+  `datetime.fromtimestamp(value)` with no `tz`, producing a **naive** datetime in
+  the host's local zone — so the same epoch surfaced as a different wall-time
+  string per machine, and `created_at.isoformat()` / `--json` output, the
+  `strftime` table cells, etc. mis-stated the absolute instant (a notebook
+  created `13:40:05` UTC rendered as the offset-less `08:40:05` under
+  `America/New_York`). It now returns `datetime.fromtimestamp(value, tz=utc)`:
+  the value is tz-aware UTC and host-independent. `.timestamp()` round-trips
+  unchanged, so internal sort/dedup/download ordering is provably unaffected —
+  only the rendered string changes (now offset-aware, identical everywhere).
+  This is the production sibling of the timezone slip pinned out of the #1511
+  golden VCR test.
+
 - **Playwright login: closing the browser during the final storage-state
   capture now shows the browser-closed help instead of a bug-report prompt**
   (#1514, deferred from the #1512 review). Every in-flow Playwright call in
--- a/src/notebooklm/_types/common.py
+++ b/src/notebooklm/_types/common.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from datetime import datetime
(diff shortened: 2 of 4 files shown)


teng-lin:
@claude review

PR #1553, opened 2026-06-11
ci: retry `uv sync` to absorb transient dep-resolution flakes

teng-lin:
## What

Wrap every regularly-running CI dependency-install step in a bounded **3-attempt backoff** (15s, 30s) around `uv sync`.

## Why

The install step had **no retry**, so a single transient PyPI/GitHub network blip during `uv sync` hard-failed the whole job. Surveying ~200 `Test` runs, this was the dominant infra-flake class on `main`:

- 2× `Install dependencies` failing on a **single matrix entry** (windows, macos) while the other 14 passed — classic network blip.
- 1× windows **MCP suite** `Install uv` / dep-resolution network hop (fetch from `raw.githubusercontent.com`).

(The remaining `main` failures in that window were *real* — API-compat gate, deterministic all-platform test failures — and are out of scope here. The many `cancelled` runs are expected `concurrency: cancel-in-progress` behavior, not flakes.)

## How it's safe

The loop **fails closed**: a real lockfile/resolution problem fails all three attempts identically, so this only absorbs transient flakes and never masks a genuine error. Each retry emits a `::warning::`; exhaustion emits `::error::` and exits non-zero.

The canonical `uv sync --frozen --extra browser --extra dev --extra markdown` string is preserved **verbatim**, so `scripts/check_ci_install_parity.py` still passes.

## Scope

| Workflow | Install sites hardened |
|---|---|
| `test.yml` | quality + MCP suite + REST server + test matrix |
| `nightly.yml` | E2E install |
| `dependency-audit.yml` | locked-env sync |
| `rpc-health.yml` | core-runtime sync |

Release-path workflows (`verify-package.yml`, `verify-artifacts.yml`) are **intentionally untouched** — they run rarely and a manual re-run is cheap; keeping the publish pipeline change-free is the safer trade.

## Verification

- All Code-Quality workflow gates pass locally: install-parity, workflow-permissions, secret-gates, action-pinning.
- YAML parses for all four files; `shell: bash` added where the matrix runs on windows.
- Retry logic unit-checked: fast on success, retries+succeeds on mid-failure, exits 1 after 3 real failures.
- 75 workflow-related unit tests pass.

## Summary by CodeRabbit

* **Chores**
  * Improved CI reliability by adding retry logic with progressive backoff to dependency installation steps across multiple workflows. Install steps now retry up to three times with increasing delays and report a failure after the final attempt, reducing flakiness from transient network or registry issues during builds and tests.

### DIFF
--- a/.github/workflows/dependency-audit.yml
+++ b/.github/workflows/dependency-audit.yml
@@ -37,10 +37,10 @@ jobs:
         # absorbs flakes — it never masks a genuine resolution error.
         shell: bash
         run: |
-          attempt=1
+          attempt=1; max=3
           until uv sync --frozen --extra browser --extra dev --extra markdown; do
-            if [ "$attempt" -ge 3 ]; then
-              echo "::error::uv sync failed after 3 attempts" >&2
+            if [ "$attempt" -ge "$max" ]; then
+              echo "::error::uv sync failed after $max attempts" >&2
               exit 1
             fi
             echo "::warning::uv sync attempt $attempt failed; retrying in $((attempt * 15))s" >&2
--- a/.github/workflows/dependency-audit.yml
+++ b/.github/workflows/dependency-audit.yml
@@ -32,7 +32,21 @@ jobs:
           enable-cache: true
 
       - name: Sync locked env (browser + dev + markdown)
-        run: uv sync --frozen --extra browser --extra dev --extra markdown
+        # Retried to ride out transient registry/network blips: a real
+        # lockfile problem fails all 3 attempts identically, so this only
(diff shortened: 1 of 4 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the review. Addressed the hardcoded-count observation in 5f076866 — each retry block now defines `max=3` once and references it in both the threshold and the error message. Left the composite-action consolidation as a follow-up (acceptable for a 4-file change, as noted), and the Gemini "unsupported file types" note is expected for YAML-only workflow changes.

PR #1559, opened 2026-06-12
fix(research): disambiguate import source rows

### LINKED ISSUE #1558: IMPORT_RESEARCH envelope probe is ambiguous for a flat single-source result
Spun out of PR #1557 review (CodeRabbit). **Pre-existing** behavior — not introduced by #1557, which faithfully preserves it.

`ResearchAPI.import_sources` (and the extracted `_row_adapters.research.unwrap_import_rows`) probe for a wrapped `[[row1, …]]` envelope with:

```python
first = result[0]
if isinstance(first, list) and first and isinstance(first[0], list):
    result = first  # treat as wrapped
```

This collides with a **flat single-row** result whose row is `[[id], title]`: e.g. `[[["id"], "t"]]`. Here `first = [["id"], "t"]` and `first[0] = ["id"]` is a list, so the probe mis-classifies the single row as a wrapper, rebinds `result` to the row, then iterates the row's *elements* (`["id"]`, `"t"`) as if they were rows — both fail the `len >= 2` row guard and are skipped, yielding `[]` (the one imported source is dropped).

The same logic exists verbatim on `main` (pre-#1557), so it's a latent ambiguity, not a regression. Whether it's reachable depends on whether the live IMPORT_RESEARCH wire ever returns the flat single-row shape (rows in practice carry >2 elements, which would not trip this) — needs a captured cassette to confirm.

**Fix idea:** disambiguate wrapped-vs-flat by checking the *inner* shape (a wrapper's `first[0]` is itself a row `[[id], title, …]` i.e. `first[0][0]` is a list), or thread the real envelope depth from a golden capture.

teng-lin:
## Summary
- fix IMPORT_RESEARCH envelope detection so flat single-row results like `[[["id"], "title"]]` are parsed as one row instead of mistaken for a wrapper
- preserve the cassette-backed wrapped shape from `research_import_sources_direct.yaml` / `research_import_sources_populated.yaml`
- update import-source edge tests and trim stale comments

Fixes #1558

## Verification
- `uv run --extra dev pytest tests/unit/test_research_api.py -k 'import_sources'`
- `uv run --extra dev pytest tests/unit/test_research.py::TestResearch::test_import_research tests/integration/test_rpc_gap_backfill_vcr.py::test_import_research_rpc_has_cassette_coverage`
- `uv run --extra dev pytest tests/_guardrails/test_module_size_ratchet.py tests/_guardrails/test_no_raw_positional_rpc_indexing.py`
- `uv run --extra dev pytest -n auto --dist=worksteal tests/unit/test_research_api.py tests/unit/test_research.py tests/integration/test_rpc_gap_backfill_vcr.py tests/_guardrails/test_module_size_ratchet.py tests/_guardrails/test_no_raw_positional_rpc_indexing.py`
- `uv run --extra dev mypy src/notebooklm/_research.py`
- `uv run --extra dev ruff check src/notebooklm/_research.py tests/unit/test_research_api.py`
- `uv run --extra dev ruff format --check src/notebooklm/_research.py tests/unit/test_research_api.py`

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1559?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust source-import handling: avoids misclassifying already-flat responses as wrapped and preserves envelope structure when appropriate.

* **Tests**
  * Added/updated edge-case tests to validate flat vs. wrapped import responses, single-row shapes, absent id envelopes, and regression scenarios to prevent double-unwrapping.

### DIFF
--- a/src/notebooklm/_research.py
+++ b/src/notebooklm/_research.py
@@ -676,14 +676,11 @@ class ResearchAPI:
             self._build_web_import_entry(src.url, src.title) for src in valid_sources
         )
 
-        params = [None, [1], effective_task_id, notebook_id, source_array]
-
         result = await self._rpc.rpc_call(
             RPCMethod.IMPORT_RESEARCH,
-            params,
+            [None, [1], effective_task_id, notebook_id, source_array],
             source_path=f"/notebook/{notebook_id}",
         )
-
         imported = []
         # ``unwrap_import_rows`` centralises the ``[[src1, ...]]`` envelope probe
         # (the former ``result[0]`` / ``first[0]`` reads) behind the research row
--- a/src/notebooklm/_row_adapters/research.py
+++ b/src/notebooklm/_row_adapters/research.py
@@ -360,6 +360,17 @@ class ResearchStartRow:
 # envelope probe + per-row reads ``import_sources`` previously open-coded.
 _IMPORT_ENVELOPE_OUTER_POS = 0
 _IMPORT_ENVELOPE_PROBE_POS = 0
+_IMPORT_ROW_ID_ENVELOPE_POS = 0
+_IMPORT_ROW_TITLE_POS = 1
+_IMPORT_ROW_MIN_LEN = 2
+
+
+def _looks_like_import_row(value: Any) -> bool:
+    """Whether ``value`` has the leading slots of an ``IMPORT_RESEARCH`` row."""
+    if not isinstance(value, list) or len(value) < _IMPORT_ROW_MIN_LEN:
+        return False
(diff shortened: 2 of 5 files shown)


teng-lin:
@claude review

PR #1582, opened 2026-06-15
fix(rpc): migrate GET_NOTEBOOK read path to nested template block (#1549)

### LINKED ISSUE #1549: Track: GET_NOTEBOOK (rLM1Ne) tail also migrated to nested [2,…,[1]] shape (read path — not yet failing)
## Summary

During the #1546 investigation, the reporter captured that the **read-path** RPC `GET_NOTEBOOK` (`rLM1Ne`) also migrated its trailing "template" block to the same nested shape as the write path (gradual Gemini-3.5 rollout):

```
# old: [notebook_id, null, [2],                           null, 1, [[null, null, []]]]
# new: [notebook_id, null, [2, null, null, [1, …, [1]]],  null, 1, [[null, null, []]]]
```

The library currently sends `[notebook_id, None, [2], None, 0]` (see `src/notebooklm/_notebooks.py` `get()` / the raw `get` helper).

## Why this is tracked, not fixed now

Unlike create / add-source, **reads are not failing on migrated cohorts**. The #1546 reporter's original table shows `list`, `source list`, and `ask` all returning OK on the exact cohort that rejects the old *write* shapes. That means the old read-path `[2]` is still accepted by the migrated backend, so there is **no user-facing breakage to fix today**.

Migrating the read path now would be churn with no bug attached, and the captured "new" read shape (`[nb_id, null, [2,…], null, 1, [[null,null,[]]]]`) doesn't match our current params (`[nb_id, None, [2], None, 0]`) — so adopting it blindly would be a larger, unverified change to a working path. Read-shape drift deserves its own capture-confirmed, live-verified evidence before we touch it (same rigor we applied to the write path in #1548).

## Trigger to act

Re-open/act on this if **any** read RPC starts returning `status=3`/`5` (Invalid argument / not-found) against the live API — e.g. the nightly RPC-health canary flips, or a user reports `list`/`get`/`ask` failing while writes succeed. At that point:

1. Capture the live web-UI `rLM1Ne` payload (and sibling reads — `wXbhsf` LIST_NOTEBOOKS, etc.) via `scripts/scrub_rpc_har.py`.
2. Probe the new shape against an un-migrated account for forward-compatibility (as done for the write path).
3. Migrate the read builders + update goldens/cassettes.

## Refs

- Parent investigation: #1546
- Write-path migration (create + add-source): #1548
- Capture tool: `scripts/scrub_rpc_har.py`

_Reported via #1546; read path explicitly scoped out of the #1548 write-path migration._

teng-lin:
Closes #1549.

## What this does

Migrates the **read-path** `GET_NOTEBOOK` (`rLM1Ne`) trailing template block from the flat `[2]` to the same nested `build_template_block()` wrapper the write path adopted in #1548 (the Gemini-3.5 rollout). Reads weren't failing yet — migrated backends still accept the old `[2]` — but this proactively aligns the read path so it stays valid as cohorts migrate, closing the long-tracked #1549.

```
old: [notebook_id, None, [2],                                  None, 0]
new: [notebook_id, None, [2, None, None, [1, None×9, [1]]],    None, 0]
                          └─ build_template_block() ─┘
```

## Live verification (our account)

Re-verified forward-compatibility against the live API on a source-bearing notebook (8 sources), using a **control-repeat baseline** + **decoded-object compare** (GET_NOTEBOOK has read-to-read volatile fields, so raw-byte equality false-alarms):

- control: two OLD-shape reads → identical stable fingerprint ✓
- **NEW shape accepted** (HTTP 200, no `status=3`/`5`) ✓
- **NEW decode == OLD decode** — notebook id/title and all 8 `SourceRow` decodes (id, title, type, status) byte-identical ✓

## Change surface

- **All 3 GET_NOTEBOOK call sites** migrated: `notebooks.get()` + `notebooks.get_raw()` via a new `build_get_notebook_params()` helper (next to `build_create_notebook_params`), and the source-listing path (`_source/listing.py`) inlined with `build_template_block()` — importing `_notebooks` there would cycle via `_source/__init__`, documented in a comment.
- **31 cassettes**: the recorded `GET_NOTEBOOK` request bodies are updated to the new shape (the `freq` VCR matcher compares request-body *shape*; the recorded **response is unchanged** — verified identical live, so this is a request-shape refresh, not a re-record).
- **Tests**: `build_get_notebook_params` live-payload pin (mirrors the create-params test) + updated request-shape assertions in the metadata/source-listing service tests.

## Test plan

- `uv run pytest` green (unit + integration; the 823 VCR integration tests replay against the migrated cassettes).
- `mypy` + `ruff` (check + format) clean.
- No user-facing behavior change — internal wire-format hardening (same posture as the #1548 write-path migration, which likewise wasn't changelogged).

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **Bug Fixes**
  * Fixed internal API request formatting for notebook operations to ensure compatibility with all account types using the updated parameter format.

* **Tests**
  * Updated test fixtures and unit tests to reflect the new request format for notebook retrieval operations.

### DIFF
--- a/src/notebooklm/_notebooks.py
+++ b/src/notebooklm/_notebooks.py
@@ -43,6 +43,20 @@ def build_create_notebook_params(title: str) -> list[Any]:
     return [title, None, None, build_template_block()]
 
 
+def build_get_notebook_params(notebook_id: str) -> list[Any]:
+    """Return the canonical GET_NOTEBOOK (``rLM1Ne``) RPC payload.
+
+    The Gemini-3.5 rollout migrated the read path's trailing template block from
+    the flat ``[2]`` to the same nested :func:`build_template_block` wrapper the
+    write path adopted in #1548 (issue #1549). Live-verified forward-compatible:
+    the nested shape returns a byte-identical decoded notebook (notebook id /
+    title and every ``SourceRow``) as the flat ``[2]`` on an un-migrated account,
+    so it is safe across cohorts. The trailing ``None, 0`` is unchanged — only
+    the template block at position 2 is migrated (the narrow scope #1549 tracks).
+    """
+    return [notebook_id, None, build_template_block(), None, 0]
+
+
 def _extract_summary(outer: Any) -> str:
     """Extract the summary string from a SUMMARIZE ``result[0]`` payload.
 
@@ -571,7 +585,7 @@ class NotebooksAPI:
                 ``title``) for unknown IDs rather than a proper RPC error, so
                 this method post-validates the parsed response.
         """
-        params = [notebook_id, None, [2], None, 0]
+        params = build_get_notebook_params(notebook_id)
         result = await self._rpc.rpc_call(
             RPCMethod.GET_NOTEBOOK,
             params,
@@ -774,7 +788,7 @@ class NotebooksAPI:
         Returns:
             Raw API response data.
         """
-        params = [notebook_id, None, [2], None, 0]
+        params = build_get_notebook_params(notebook_id)
         return await self._rpc.rpc_call(
             RPCMethod.GET_NOTEBOOK,
             params,
--- a/src/notebooklm/_source/listing.py
+++ b/src/notebooklm/_source/listing.py
@@ -11,6 +11,7 @@ from .._row_adapters.sources import SourceRow
 from .._runtime.contracts import RpcCaller
 from ..rpc import RPCError, RPCMethod, safe_index
 from ..types import Source
+from .upload_payloads import build_template_block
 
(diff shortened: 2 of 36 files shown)


teng-lin:
@claude review

PR #1471, opened 2026-06-07
ci(test): add workflow_dispatch with custom_branch to gate release branches

teng-lin:
## Summary

Adds a `workflow_dispatch` trigger (with a `custom_branch` input) to the **Test**
workflow so the full `quality` + cross-platform matrix gate can be run **on
demand against an arbitrary branch**.

### Why
`test.yml` only triggers on `push`/`pull_request` to `main`. A release backport
branch (e.g. `release/v0.7.1`) is intentionally based on the old `0.7.x` line and
**conflicts with `main`**, so it can't run CI via a PR — GitHub can't build the
test-merge commit (`refs/pull/N/merge`) for a `DIRTY` PR, and `pull_request` runs
never start. There was no way to run the gate on such a branch short of polluting
it by merging `main` in.

### What
- New `workflow_dispatch.inputs.custom_branch` (mirrors the existing
  `nightly.yml` pattern).
- Both jobs check out the target via `ref: ${{ inputs.custom_branch || github.ref }}`.
- `custom_branch` folded into the `concurrency.group` so a manual release-branch
  run doesn't cancel `main`'s push/PR run.

### No behavior change for existing triggers
For `push`/`pull_request` the input is empty, so checkout falls back to
`github.ref` (identical to today) and the concurrency group just gains a trailing
`-`.

### Usage (after merge)
`workflow_dispatch` only appears once the workflow is on the default branch, so
this must land on `main` first. Then:

```bash
gh workflow run test.yml --ref main -f custom_branch=release/v0.7.1
```

## Test plan
- [x] `yaml.safe_load` parses; `on:` = push / pull_request / workflow_dispatch
- [x] `scripts/check_workflow_permissions.py` passes
- [x] `scripts/check_workflow_secret_gates.py` passes
- [x] `scripts/check_action_pinning.py` passes
- [ ] CI green on this PR
- [ ] After merge: dispatch against `release/v0.7.1` and confirm it runs the gate

## Summary by CodeRabbit

* **Chores**
  * Enhanced test workflow to support manual execution with custom branch selection. Improved concurrency handling to prevent conflicts between automated and manually-triggered test runs on different branches.

PR #1621, opened 2026-06-21
ci(security): set persist-credentials:false on checkout steps (#1471)

teng-lin:
## Summary
- Adds `persist-credentials: false` to all **13** `actions/checkout@v6` steps across the 10 workflow files.
- Closes the zizmor **`artipacked`** vector: by default `actions/checkout` writes the `GITHUB_TOKEN` into `.git/config`, where any subsequent step or compromised third-party action in the job can read and exfiltrate it. CodeRabbit flagged this on #1471 (🟠 Major) and it was never addressed.

## Why it's safe
- No workflow does `git push` / `git commit` / `git tag` / auto-commit, so nothing writes back using the persisted credential.
- `claude.yml` authenticates via its own `claude_code_oauth_token` and runs with `contents: read` — it doesn't rely on the checkout-persisted token.
- Token-using steps (`gh` calls, ref resolution during checkout itself) are unaffected: `persist-credentials: false` only stops the token from being *persisted to .git/config after* checkout; the token still works *during* checkout and via explicit `GH_TOKEN`/`secrets`.

## Scope
- Pure, behavior-preserving CI hardening. YAML validated for all 10 files; the `claude.yml` step correctly merges the new key into its existing `with:` (`fetch-depth: 1`).
- **Out of scope:** SHA-pinning of actions (zizmor `unpinned-uses`) — a separate, higher-churn concern.

## Test plan
- [x] All 10 workflow YAML files parse cleanly
- [x] 13/13 checkout steps covered; no duplicate `with:` keys
- [x] Polished (code-simplifier + claude+codex review + ultrathink — all clean)

https://claude.ai/code/session_0158QsP9LWmJdfob2491LW3Z

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Chores**
  * Enhanced security configurations in CI/CD workflows to improve credential handling across automated build, test, and publish pipelines.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -41,6 +41,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           fetch-depth: 1
+          persist-credentials: false
 
       - name: Run Claude Code
         id: claude
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -19,6 +19,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - name: Initialize CodeQL
         uses: github/codeql-action/init@v4
--- a/.github/workflows/dependency-audit.yml
+++ b/.github/workflows/dependency-audit.yml
@@ -26,6 +26,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
 
       - uses: astral-sh/setup-uv@v7
         with:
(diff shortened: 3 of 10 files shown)


teng-lin:
@claude review

PR #1628, opened 2026-06-24
fix(sources): pin .md content-type for uploads on Python 3.10

Leslielu:
-

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1628?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upload content-type resolution so `.md`/`.markdown` files reliably resolve to `text/markdown`, even when system MIME guessing is incomplete.
  * Kept safe fallback behavior for unknown extensions (`application/octet-stream`), and ensured precedence when an explicit MIME type is provided.

* **Tests**
  * Added unit coverage for Markdown detection when MIME guessing fails (including uppercase suffixes).
  * Added unit coverage for unknown-extension fallback and for correct precedence between explicit MIME and guessing/overrides.

### DIFF
--- a/src/notebooklm/_source/_upload_decode.py
+++ b/src/notebooklm/_source/_upload_decode.py
@@ -307,10 +307,8 @@ def _resolve_upload_content_type(file_path: Path, mime_type: str | None) -> str:
     guessed, _encoding = mimetypes.guess_type(file_path.name)
     if guessed:
         return guessed
-    # ``mimetypes`` has no entry for some text suffixes (e.g. ``.md``) on
-    # Python 3.10 / hosts without a system MIME table — fall back to the
-    # pinned mapping before the opaque ``application/octet-stream`` default,
-    # which NotebookLM cannot infer a parser for (processing fails, #1627).
+    # Fall back to the pinned overrides before the opaque octet-stream default
+    # (see ``_EXTENSION_CONTENT_TYPES`` for why; #1627).
     return _EXTENSION_CONTENT_TYPES.get(file_path.suffix.lower(), "application/octet-stream")
 
 
--- a/src/notebooklm/_source/_upload_decode.py
+++ b/src/notebooklm/_source/_upload_decode.py
@@ -40,6 +40,16 @@ _MEDIA_TRANSIENT_ERROR_TYPES: tuple[int | None, ...] = (10, 0, None)
 _STRICT_TRANSIENT_ERROR_TYPES: tuple[int | None, ...] = ()
 _HTML_UPLOAD_SUFFIXES = frozenset({".html", ".htm", ".xhtml", ".xht"})
 _HTML_UPLOAD_CONTENT_TYPES = frozenset({"text/html", "application/xhtml+xml"})
+# ``mimetypes.guess_type`` returns ``None`` for some text suffixes on Python 3.10
+# and on hosts without a populated ``/etc/mime.types`` (notably ``.md``). Without
+# an override the upload falls back to ``application/octet-stream`` (see
+# ``_resolve_upload_content_type``), NotebookLM cannot infer how to parse the
+# file, and processing fails with status=ERROR. Pin the types we know NotebookLM
+# accepts so upload works regardless of the host's MIME table.
+_EXTENSION_CONTENT_TYPES: dict[str, str] = {
+    ".md": "text/markdown",
+    ".markdown": "text/markdown",
+}
 
 
 def _normalize_upload_path(path: str) -> str:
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

src/notebooklm/_source/_upload_decode.py
```diff
@@ -295,7 +305,13 @@ def _resolve_upload_content_type(file_path: Path, mime_type: str | None) -> str:
         return content_type
 
     guessed, _encoding = mimetypes.guess_type(file_path.name)
-    return guessed or "application/octet-stream"
+    if guessed:
+        return guessed
+    # ``mimetypes`` has no entry for some text suffixes (e.g. ``.md``) on
+    # Python 3.10 / hosts without a system MIME table — fall back to the
+    # pinned mapping before the opaque ``application/octet-stream`` default,
+    # which NotebookLM cannot infer a parser for (processing fails, #1627).
```
teng-lin: The rationale here restates, nearly verbatim, the comment already at the `_EXTENSION_CONTENT_TYPES` definition site. Keep the "why" in one canonical place and leave a short pointer here:

```suggestion
    # Fall back to the pinned overrides before the opaque octet-stream default
    # (see ``_EXTENSION_CONTENT_TYPES`` for why; #1627).
```

tests/unit/test_source_upload_coverage.py
```diff
@@ -148,6 +148,35 @@ def test_resolve_upload_content_type_blank_mime_raises() -> None:
...
+
+def test_resolve_upload_content_type_unknown_suffix_falls_back_to_octet_stream(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A suffix with no MIME guess and no override still falls back to octet-stream."""
+    import mimetypes
+    from pathlib import Path
+
+    monkeypatch.setattr(mimetypes, "guess_type", lambda _name: (None, None))
+    assert _resolve_upload_content_type(Path("blob.unknownext"), None) == "application/octet-stream"
```
teng-lin, line 177: Nice coverage of the fallback + case-insensitivity. Worth also pinning the precedence the new code now relies on, so a future refactor can't silently let the override shadow an explicit mime / a real guess, or false-positive on `notes.md.txt`:

```suggestion
    assert _resolve_upload_content_type(Path("blob.unknownext"), None) == "application/octet-stream"


def test_resolve_upload_content_type_override_precedence(
    monkeypatch: pytest.MonkeyPatch,
) -> None:
    """The pinned override sits last in precedence: explicit mime > guess > override.

    The override only fills the gap when ``mimetypes`` misses (#1627); it must not
    shadow an explicit mime_type or a successful guess, and it keys on the final
    suffix only (so ``notes.md.txt`` is not mistaken for markdown).
    """
    import mimetypes
    from pathlib import Path

    # Explicit mime_type wins over the pinned ``.md`` override.
    assert _resolve_upload_content_type(Path("notes.md"), "text/plain") == "text/plain"
    # A successful guess wins over the override (sentinel proves the guess path ran).
    monkeypatch.setattr(mimetypes, "guess_type", lambda _name: ("application/x-sentinel", None))
    assert _resolve_upload_content_type(Path("notes.md"), None) == "application/x-sentinel"
    # On a miss, the override keys on the actual suffix: ``.md.txt`` -> ``.txt`` -> octet-stream.
    monkeypatch.setattr(mimetypes, "guess_type", lambda _name: (None, None))
    assert _resolve_upload_content_type(Path("notes.md.txt"), None) == "application/octet-stream"
```

teng-lin (commented):
Ran a polish pass (code-simplifier → Claude + Codex review → ultrathink). The fix is correct and merge-ready — no Critical/Important findings. Two committable suggestions below; a couple of non-blocking FYIs:

- **Import tidy (optional):** the two new tests use function-local `import mimetypes` / `from pathlib import Path`. `Path` is already imported elsewhere via local imports too — hoisting both to the module top and dropping the per-test copies reads cleaner, but it's purely cosmetic and spans lines outside this diff, so I left it as a note rather than a suggestion.
- **Heads-up:** avoid running `uv sync` in a clean checkout for this branch — it bumps `ruff` in `uv.lock` (0.15.17 → 0.15.18), which shouldn't ride along in this PR.

Verified locally: `pytest tests/unit/test_source_upload_coverage.py` green, `ruff`/`mypy` clean, public-API compat unaffected.

teng-lin:
@Leslielu thanks for the fix! I ran a polish pass and left a review with two committable suggestions:

1. Collapse the duplicated rationale comment in `_resolve_upload_content_type` into a one-line pointer.
2. Add a `test_resolve_upload_content_type_override_precedence` test pinning the precedence the new code relies on (explicit mime > guess > override, and `notes.md.txt` not mis-resolved as markdown).

Both are "Commit suggestion"-ready and verified green locally. Could you apply/address them when you get a chance? Also a heads-up: avoid letting `uv sync` bump `ruff` in `uv.lock` into this PR. Thanks!

teng-lin:
@claude review

PR #1635, opened 2026-06-27
fix(chat): surface oversized-request errors instead of "No parseable chunks"

### LINKED ISSUE #1634: Chat parser skips batchexecute error frame tag "e" — masks size-limit errors (code 132/133) as "No parseable chunks"
## Summary

The streamed-chat parser in `_chat/wire.py` only recognizes server error frames tagged `"er"`, but the batchexecute chat endpoint emits oversized-request errors under the single-letter tag `"e"` (e.g. `["e",4,null,null,132]`). The `"e"` frame is silently skipped, every `wrb.fr` chunk in the response has a `null` inner JSON, and with zero parseable chunks the parser raises the generic:

```
ChatResponseParseError: No parseable chunks in streaming chat response (6 lines scanned).
The response was empty or the API wire format may have changed.
```

…instead of surfacing the real server error code. Reported in discussion #1472.

## Reproduced

**1. Deterministic** — feeding the captured wire frame to the parser:

```python
from notebooklm._chat.wire import parse_streaming_chat_response
raw = ')]}\'\n\n38\n[["wrb.fr",null,null,null,null,[3]]]\n57\n' \
      '[["di",198],["af.httprm",197,"-1460015816955467607",6]]\n25\n[["e",4,null,null,132]]\n'
parse_streaming_chat_response(raw)
# -> ChatResponseParseError: No parseable chunks in streaming chat response (6 lines scanned). ...
```

**2. Live API** — a long (high-token) question against a real notebook reproduces the
identical `6 lines scanned` error. Capturing the raw response at the `parse_streaming_chat_response`
call site shows the `"e"` error frame:

```
)]}'

38
[["wrb.fr",null,null,null,null,[3]]]
58
[["di",219],["af.httprm",218,"-6359417103605345913",11]]
25
[["e",4,null,null,133]]
```

Note: my live capture returned code **133**; discussion #1472 reported **132**. Both are
`"e"` error frames in the same position and both are silently dropped — the exact code is
size/condition-dependent. Short questions return a normal answer, confirming this is a
request-size/token ceiling, not malformed content.

## Root cause

`_chat/wire.py` `_extract_chunk_with_parseable()`:

```python
if tag == "er":
    _raise_chat_error_frame(item)
```

The endpoint's error tag for this class is `"e"`, not `"er"`, so the frame never reaches
`_raise_chat_error_frame`. The accompanying `["wrb.fr", null, null, null, null, [3]]` chunk
also fails to count as parseable (`item[5] == [3]` doesn't match the `UserDisplayableError`
shape `raise_if_rate_limited` looks for), so the response yields zero parseable chunks.

## Suggested fix

Handle `"e"` alongside `"er"` in `_extract_chunk_with_parseable()`:

```python
if tag in ("er", "e"):
    _raise_chat_error_frame(item)
```

`_raise_chat_error_frame` already reads the code slot via `StreamFrameRow(item).error_code`,
so the surfaced `ChatError` would carry `code 132`/`133` instead of the misleading generic
parse error. Worth pairing with a test built from the captured frame above (and ideally a
clearer message for this code, e.g. "request too large").

## Environment

- Reproduced on `main` (HEAD `313bde74`), Python 3.11, live free-tier account.
- Original report: discussion #1472 + attached findings (notebooklm v0.8.0).

teng-lin:
## Summary

An oversized chat request (a very long question, or a large accumulated conversation history) returns a null-inner `wrb.fr` frame carrying the bare status payload `[3]` (gRPC `INVALID_ARGUMENT`) and no answer chunk. The streamed-chat parser only special-cased the rate-limit payload (status `8`, `RESOURCE_EXHAUSTED` via `UserDisplayableError`), so the `[3]` status silently fell through to zero parseable chunks and raised the generic:

```
ChatResponseParseError: No parseable chunks in streaming chat response (6 lines scanned)...
```

…instead of surfacing the real cause. This now raises a clear `ChatError`.

## Related Issue

Closes #1634

## Changes

- `_chat/wire.py`: in the null-inner-`wrb.fr` branch (the server-side error channel), after the existing rate-limit check, surface any remaining error payload as a `ChatError` echoing its leading gRPC-style status code via a new `_raise_chat_status_error_frame` helper. Status `3` adds a "request too large" hint since that is its common trigger.
- `_row_adapters/chat.py`: add `ErrorPayloadRow.status_code` (the `error_payload[0]` slot) and pin it in the position contract.
- Tests: the exact #1634 wire capture now raises `ChatError` (status 3); a regression test asserts a normal answer ending in the byte-count trailer still parses; an unrecognized status code surfaces a generic `ChatError`.

## Root cause

A successful streamed answer **always** carries a string `inner_json` (verified across all chat cassettes), so a null-inner `wrb.fr` frame only ever arrives on the error channel, where `item[5]` leads with a gRPC-style status code. Only `8` was handled; every other code (including `3`) degraded to the misleading parse error.

## Note on the issue's hypothesis

The issue pinned the error on the trailing `["e", n, null, null, N]` frame and proposed treating `tag in ("er", "e")` as an error. Investigation showed that frame is a **batchexecute byte-count trailer** (`N` = cumulative response bytes, present on *every* streamed response — e.g. `["e",24,null,null,441040]` ends a successful 440 KB answer). The `132`/`133` in the report were simply the byte size of the small error body, **not** an error code. Keying off the `"e"` tag would have raised on every chat call (it broke 8 chat integration tests in a local trial); the real, stable signal is the `wrb.fr` `[3]` status payload. A regression test guards against the byte-count-trailer false positive.

## Test Plan

- [x] Tests pass (`uv run pytest ... --cov-fail-under=90` → 10424 passed, 92% coverage)
- [x] Linting and formatting pass (`uv run ruff check .` / `ruff format --check .`)
- [x] No architectural shape change; no ADR needed.

## Notes

The fix handles the error channel generically (any non-rate-limit status code surfaces a `ChatError` instead of collapsing into a parse error), so future server-side status codes on this channel are no longer masked. The position-sensitive `item[5][0]` read is centralised in `ErrorPayloadRow` per the repo's row-adapter convention.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DFpW8w1gMQVXN9kHS5MTLD)_

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1635?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streamed chat error handling so rejected or malformed server responses now surface clearer error messages instead of a generic parse failure.
  * Oversized requests now report a more specific invalid-argument message.
  * Non-rate-limit server status codes are now shown in the error details.
  * Preserved correct parsing for valid streamed responses, including benign trailer frames.
* **Tests**
  * Added coverage for oversized requests, unknown server error codes, and valid trailer handling.

### DIFF
--- a/src/notebooklm/_chat/wire.py
+++ b/src/notebooklm/_chat/wire.py
@@ -329,9 +329,20 @@ def _extract_chunk_with_parseable(
             # recognized error payload is not a successfully decoded
             # envelope. The error-payload path raises, so flow only
             # reaches the next iteration when item[5] was absent/unusable.
+            #
+            # A successful streamed answer ALWAYS carries a string ``inner_json``
+            # (the answer JSON), so a null-inner ``wrb.fr`` only ever arrives on
+            # the server-side error channel. The first slot of ``item[5]`` is a
+            # gRPC-style status code: ``8`` (RESOURCE_EXHAUSTED) is the rate
+            # limit handled above; ``3`` (INVALID_ARGUMENT) is what an oversized
+            # request returns (``["wrb.fr", null, ..., [3]]``). Before this
+            # guard, every non-rate-limit status silently fell through to "no
+            # parseable chunks", masking real errors like the request-size
+            # ceiling — surface it instead (issue #1634).
             error_payload = frame.error_payload
             if error_payload is not None:
                 raise_if_rate_limited(error_payload)
+                _raise_chat_status_error_frame(error_payload)
             continue
 
         try:
@@ -428,6 +439,37 @@ def _raise_chat_error_frame(item: list) -> NoReturn:
         f"Chat request failed: the server returned an error frame{detail}. "
         "This usually means the request was rejected or the conversation "
         "could not be served; try again."
+    )
+
+
+def _raise_chat_status_error_frame(error_payload: list) -> NoReturn:
+    """Surface a non-rate-limit ``wrb.fr`` error payload as a ``ChatError``.
+
+    A streamed answer always carries a string ``inner_json``; a null-inner
+    ``wrb.fr`` frame is the server-side error channel, where ``error_payload``
+    (``item[5]``) leads with a gRPC-style status code. ``8``
+    (RESOURCE_EXHAUSTED) is the rate limit handled by
+    :func:`raise_if_rate_limited` before this call; the load-bearing case here
+    is ``3`` (INVALID_ARGUMENT), which an oversized request returns
+    (``["wrb.fr", null, ..., [3]]``). Such errors previously collapsed into the
+    generic ``ChatResponseParseError`` ("No parseable chunks …"), masking the
+    real cause (issue #1634). The status code is echoed; code ``3`` adds the
+    request-size hint since that is its common trigger.
+    """
+    code = ErrorPayloadRow(error_payload).status_code
+    if code == 3:
+        raise ChatError(
+            "Chat request failed: the server rejected it (status code 3, "
+            "INVALID_ARGUMENT). This usually means the request was too large — "
+            "the question, or the accumulated conversation history, exceeded the "
+            "server's size limit; shorten the question or start a new "
+            "conversation and try again."
+        )
+    detail = f" (status code {code!r})" if code is not None else ""
+    raise ChatError(
+        f"Chat request failed: the server returned an error{detail} with no "
+        "answer payload. The request was rejected or could not be served; "
+        "try again."
     )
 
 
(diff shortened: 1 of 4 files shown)


teng-lin:
Closing as a duplicate of #1636. Both PRs land the same fix for #1634 (null-inner `wrb.fr` frame carrying bare gRPC status `[3]`/INVALID_ARGUMENT was silently swallowed → now surfaced as `ChatError`), and they edit the same source files, so only one can merge.

#1636 supersedes this PR: it was produced by a run with **live auth**, so it carries a real integration cassette (`chat_ask_oversized_rejection.yaml`) and an empirically-confirmed request-size ceiling, whereas this PR (sandbox run) proved the logic but couldn't validate the boundary against the live server.

Superseded by #1636.

PR #1644, opened 2026-06-28
test(curl_cffi): skip-if-missing instead of importorskip so allow_no_vcr stays collected

teng-lin:
## What

The two curl_cffi integration suites used a module-level `pytest.importorskip("curl_cffi")`, which **drops the whole module from collection** when the optional `[impersonate]` extra is absent. That makes their entries in `integration_allow_no_vcr_files.txt` look *stale* to the test-taxonomy inventory (a file is "stale" if allowlisted but not currently collected with the `allow_no_vcr` marker), failing `test_integration_allow_no_vcr_taxonomy_policy_is_current` in any environment without curl_cffi installed.

## Why

The adapter import (`notebooklm._curl_cffi_transport`) does **not** need curl_cffi at import time (the real import is lazy), so collection succeeds without the extra. There's no reason to drop the module from collection — only to skip the tests at runtime.

## How

Replace the module-level `importorskip` with a `skipif` pytestmark: the module stays **collected** (its `allow_no_vcr` marker visible to the inventory) and the tests **skip at runtime** when curl_cffi is missing. The notebooklm imports become proper top-level imports (no more `# noqa: E402`).

## Verification
Without curl_cffi installed: the 11 tests **collect and skip** (previously: silently dropped), `scripts/test_taxonomy_inventory.py --check` exits 0, and `test_integration_allow_no_vcr_allowlist` passes. ruff clean.

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1644?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of an optional dependency so related tests are skipped at runtime instead of being dropped from collection.

* **Tests**
  * Kept integration test coverage visible in test runs even when the optional browser/client package is not installed.
  * Preserved test markers so the test suite reports remain more consistent.

### DIFF
--- a/tests/integration/test_curl_cffi_cassette_conformance.py
+++ b/tests/integration/test_curl_cffi_cassette_conformance.py
@@ -23,6 +23,7 @@ curl_cffi bypasses VCR (libcurl, not httpx) so this is local-server-backed
 from __future__ import annotations
 
 import gzip
+import importlib.util
 import json
 import threading
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

PR #1700, opened 2026-06-29
fix(auth): interactive `notebooklm login` no longer hangs 5 min (wait_until="commit") (#1697)

### LINKED ISSUE #1697: Interactive `notebooklm login` hangs 5 min and never saves auth — wait_for_url defaults to wait_until="load", but notebooklm.google.com never fires the load event
## Summary

`notebooklm login` opens a headed Chromium, the user completes the Google sign-in
successfully (cookies land in the persistent browser profile), but the CLI's
login-success detector never fires. It hangs for the full 5 minutes, prints
"Login not detected within 5 minutes", and **never writes `storage_state.json`** —
even though authentication actually succeeded.

Root cause: the detector uses `page.wait_for_url(..., timeout=300_000)` with no
`wait_until`, so it defaults to Playwright's `wait_until="load"`. But
`notebooklm.google.com` is a streaming SPA that reaches `DOMContentLoaded`
(`document.readyState === "interactive"`) and **never fires the `load` event**, so
the wait blocks until timeout.

## Environment

- notebooklm-py **0.7.2** (also reproduces on current `main`)
- Playwright **1.60.0**, bundled Chromium
- Python 3.11, Linux (WSL2)

## Steps to reproduce

1. Have a valid Google session in the persistent browser profile (i.e. you can
   sign in successfully).
2. Run `notebooklm login`.
3. Complete the Google login in the browser window.

**Expected:** login is detected within a second or two and `storage_state.json` is
written.
**Actual:** the CLI sits on "Waiting for login (up to 5 minutes)…", then fails with
"Login not detected within 5 minutes." `storage_state.json` is never written,
despite the full Google cookie set (`SID`, `__Secure-1PSID`, …) being present in the
persistent profile.

## Root cause

On current `main` the detector is in `src/notebooklm/_auth/browser_capture.py`
(in released 0.7.2 the same code is in `cli/services/playwright_login.py`):

- `browser_capture.py:611` — `page.wait_for_url(f"{get_base_url()}/**", timeout=300_000)`
  → no `wait_until` ⇒ defaults to `"load"`.
- `browser_capture.py:532` — initial `page.goto(f"{get_base_url()}/", timeout=30000)`
  → also defaults to `"load"`.

`notebooklm.google.com` keeps a long-lived connection open and never fires the
`load` lifecycle event, so a `load`-gated wait blocks indefinitely.

## Evidence

Against an already-authenticated persistent profile (headless, for isolation):

| call | result |
|------|--------|
| `page.goto("https://notebooklm.google.com/")` (default `load`) | **TimeoutError** after 30s |
| `page.wait_for_load_state("load", timeout=60000)` | **never fires** (`readyState` stays `interactive`) |
| `page.wait_for_url(".../**")` (default `load`) | **times out** even though `page.url` already matches |
| `page.wait_for_url(".../**", wait_until="commit")` | **returns in ~0.0s** |

So the URL/glob is fine — it's purely the `load` lifecycle gate.

## Suggested fix

Pass `wait_until="commit"` on the detector (and the initial goto):

```python
# browser_capture.py:532
page.goto(f"{get_base_url()}/", wait_until="commit", timeout=30000)
# browser_capture.py:611
page.wait_for_url(f"{get_base_url()}/**", wait_until="commit", timeout=300_000)
```

There's already precedent for this in the *same function*: the regional
cookie-forcing gotos at lines 641/649 use `wait_until="commit"` (added for #214).
`commit` is the right level here — detection only needs to confirm we reached the
authenticated host so cookies can be captured; it shouldn't wait for a `load` that
never comes.

## Notes

- The bare `wait_for_url(..., timeout=300_000)` was introduced in #415 (auto-detect
  login) and relocated into `browser_capture.py` by the #1512 refactor; it defaults
  to `load` in both places.
- The newer headless re-auth / `connect_over_cdp` / master-token paths don't cover
  the default headed `notebooklm login` flow, which is what most users hit.
- Happy to send a small PR with the two-line change (plus a regression test if you'd
  like one).

## Workaround for anyone landing here

Your sign-in actually succeeded — the cookies are in the persistent profile. You can
materialize `storage_state.json` from that profile (visit `accounts.google.com`
first so the rotating `__Secure-1PSIDTS` cookie is set), then `notebooklm auth check`.

---

<sub>🤖 Filed by @nkoz18. This issue was researched and drafted by Claude Code
(Anthropic) and posted with my review and permission.</sub>

teng-lin:
## Summary

Fixes #1697.

`notebooklm login` opened headed Chromium, the user signed in successfully (cookies landed in the persistent profile), but the success detector never fired — it hung the full 5 minutes, printed "Login not detected within 5 minutes", and **never wrote `storage_state.json`**, despite authentication having succeeded.

## Root cause

`page.wait_for_url(...)` and `page.goto(...)` default to Playwright's `wait_until="load"`. `notebooklm.google.com` is a streaming SPA that reaches DOMContentLoaded (`readyState === "interactive"`) but **never fires the `load` event** (long-lived connection stays open), so a `load`-gated wait blocks until timeout even though the URL already matches and cookies are present.

The reporter's evidence confirms the key link: the same `wait_for_url` with `wait_until="commit"` returns in ~0.0s while the `load` default times out.

The defect has been latent since #415 (the bare `wait_for_url(..., timeout=300_000)` was never given a `wait_until`), relocated verbatim into `browser_capture.py` by #1512. It reproduces identically on 0.7.2 and `main` with no relevant code change between them — the fingerprint of a Google-side change making the page more aggressively streaming, on a path most users never re-run once they have cached auth.

## Fix

Add `wait_until="commit"` to the three NotebookLM-host navigations that defaulted to `load`:

- `browser_capture.py` — initial login `goto`
- `browser_capture.py` — success-detector `wait_for_url` (the headline symptom)
- `browser_capture.py` — `connect_over_cdp` re-auth `goto` (same defect; would otherwise waste 30s → `TimeoutError`)

`commit` resolves once response headers are processed — the same level the regional cookie-forcing gotos already use (#214) — which is all detection needs. Cookies are still read later at `context.storage_state()` (after the cookie-forcing round-trips), so resolving early is safe. This also **repairs** the already-logged-in and headless re-auth branches, whose initial `goto` hit the same `load` timeout today.

`master_token.py`'s `goto(_EMBEDDED_SETUP_URL)` was considered and left unchanged — different host (`accounts.google.com`), fires `load` normally, and polls the cookie jar rather than URL-waiting.

## Tests

Regression assertions at all three sites verify the **invariant** — `wait_until` is set and is not the `load` default (i.e. in `{"commit", "domcontentloaded"}`) — rather than the literal `"commit"`, so a future switch to `domcontentloaded` won't spuriously fail but a regression to the default will. Verified red→green against the unfixed code (the unfixed sites have `wait_until` absent → `.get()` returns `None` ∉ the set → assertion fails).

- `tests/unit/cli/test_login.py` — initial goto + success-detector `wait_for_url`
- `tests/unit/test_browser_capture_cdp_arm.py` — CDP re-auth goto

Full local sweep: 294 passed across `test_login` / `test_browser_capture_*` / `test_auth_*`; ruff + mypy clean.

https://claude.ai/code/session_01YG1NZU5zNDAX2gWGNubjwi

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1700?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved login and re-auth navigation so browser sign-in flows complete more reliably without hanging on page load.
  * Reduced timeouts when opening the app during interactive and automated login steps.
  * Better handling of streaming web pages, making sign-in and landing detection smoother.

* **Tests**
  * Strengthened automated checks to ensure browser navigation uses faster, non-default wait behavior during login flows.

### DIFF
--- a/src/notebooklm/_auth/browser_capture.py
+++ b/src/notebooklm/_auth/browser_capture.py
@@ -529,7 +529,13 @@ def run_browser_capture(
             # Retry navigation on transient connection errors with backoff
             for attempt in range(1, LOGIN_MAX_RETRIES + 1):
                 try:
-                    page.goto(f"{get_base_url()}/", timeout=30000)
+                    # wait_until="commit": notebooklm.google.com is a streaming
+                    # SPA that never fires the "load" event (readyState stays
+                    # "interactive"), so Playwright's default wait_until="load"
+                    # would block until timeout. "commit" resolves once response
+                    # headers are processed -- enough to land on the host and
+                    # classify page.url. See #1697 (and the #214 precedent below).
+                    page.goto(f"{get_base_url()}/", wait_until="commit", timeout=30000)
                     break
                 except PlaywrightError as exc:
                     error_str = str(exc)
@@ -608,7 +614,14 @@ def run_browser_capture(
                 io.emit("2. Authentication will be saved automatically once login is detected\n")
                 io.emit("[dim]Waiting for login (up to 5 minutes)...[/dim]")
                 try:
-                    page.wait_for_url(f"{get_base_url()}/**", timeout=300_000)
+                    # wait_until="commit", not the default "load": the SPA never
+                    # fires "load", so a load-gated wait hangs the full 5 min even
+                    # though sign-in already succeeded and the URL already matches
+                    # (the #1697 symptom). "commit" returns as soon as we reach the
+                    # host. Cookies are read later at storage_state() (after the
+                    # cookie-forcing round-trips), so resolving early is safe;
+                    # page.content() at capture time is best-effort/None-tolerant.
+                    page.wait_for_url(f"{get_base_url()}/**", wait_until="commit", timeout=300_000)
                 except PlaywrightTimeout:
                     io.emit(
                         "[red]Login not detected within 5 minutes.[/red]\n"
@@ -828,7 +841,11 @@ def run_cdp_capture(
             # This avoids navigating (and thereby disrupting) a tab the operator
             # is actively using.
             page = context.new_page()
-            page.goto(f"{get_base_url()}/", timeout=30000)
+            # wait_until="commit": same streaming-SPA reason as the headed login
+            # arm -- the default "load" never fires on notebooklm.google.com, so
+            # this CDP re-auth goto would otherwise waste 30s then TimeoutError
+            # before landing classification. See #1697.
+            page.goto(f"{get_base_url()}/", wait_until="commit", timeout=30000)
 
             # SAME landing classification as the headless launch arm: if we did
             # not land on the NotebookLM host, the attached browser's Google
--- a/tests/unit/cli/test_login.py
+++ b/tests/unit/cli/test_login.py
@@ -386,8 +386,12 @@ class TestLoginCommand:
         goto_calls = mock_page.goto.call_args_list
         # 3 calls: initial NOTEBOOKLM_URL, then accounts.google.com, then NOTEBOOKLM_URL
         assert len(goto_calls) == 3
-        assert goto_calls[1].kwargs.get("wait_until") == "commit"
-        assert goto_calls[2].kwargs.get("wait_until") == "commit"
+        # Every goto in the login flow must use an early lifecycle state, NOT the
+        # Playwright default "load" -- the NotebookLM SPA never fires "load", so a
+        # load-gated goto hangs (#1697). Assert the invariant (not the literal
+        # "commit") so a future switch to "domcontentloaded" doesn't spuriously fail.
+        for call in goto_calls:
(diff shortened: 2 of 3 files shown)


teng-lin:
@claude review

PR #1701, opened 2026-06-29
perf(mcp): batch source-ref resolution to list sources once (#1689)

### LINKED ISSUE #1689: MCP: batch source-ref resolution (avoid N concurrent sources.list calls)
Follow-up from PR #1688 (review by agy).

`artifact_generate` and `chat_ask` resolve multiple `source_ids` via `asyncio.gather(resolve_source(...) for ref in refs)`. Each non-UUID ref (title/prefix) independently calls `client.sources.list(notebook_id)`, so passing N non-UUID refs fires N concurrent identical list RPCs. Full-UUID refs take the no-list fast path, so this only bites the mixed/title path.

Fix: a shared batch-resolve helper (e.g. in `mcp/_resolve.py`) that lists sources once and resolves all refs against the single snapshot, applied to BOTH tools (keep them consistent — do not optimize one and not the other).

Deferred from #1688 because the proper fix is a cross-cutting refactor of the pre-existing resolution pattern, out of #1680's scope.

teng-lin:
## Summary

`chat_ask` and `artifact_generate` resolved N source refs via `asyncio.gather(resolve_source(...) for ref in refs)`. Each non-UUID ref independently called `client.sources.list(notebook_id)` → N identical concurrent list RPCs for a single tool call.

This adds one shared batch resolver that lists the notebook's sources **at most once**:

- **`resolve_sources(client, notebook_id, refs) -> list[str]`** in `mcp/_resolve.py`:
  - `validate_id` every ref first (so an empty/whitespace ref raises `ValidationError` before any resolution).
  - If **every** validated ref is a full canonical UUID → return verbatim, **no list call** (parity with the single-ref fast-path).
  - Else list **once**, then resolve each ref over that single snapshot via the existing single-ref helpers (full-UUID → verbatim, hex → `_resolve_hex`, else `_resolve_by_title`). Same matching rules, same exceptions (`SourceNotFoundError` / `AmbiguousIdError`), **input order preserved**.
- Both call sites swapped to `await resolve_sources(client, nb_id, refs)`, keeping each caller's existing `if refs else None` guard (empty/omitted still means "all sources", #1652).
- Dropped the now-unused `import asyncio` and `resolve_source` import from both tool modules.

## Behavior delta

Resolution becomes sequential over one snapshot instead of N concurrent lists. Errors are now **deterministic** (raise in input order) rather than subject to `gather`'s first-to-complete race — a strict improvement. The single-ref `resolve_source` is untouched (still used by other tools).

## Tests

- `test_resolve.py`: all-UUID → zero list calls; mixed / title-only → exactly **one** list call with order preserved; non-match → `SourceNotFoundError`; ambiguous → `AmbiguousIdError`; whitespace ref in batch → `ValidationError` before listing.
- `test_chat.py` / `test_artifacts.py`: two non-UUID refs → `sources.list` awaited **exactly once**, resolved-id order matches input order.

Verification: `uv run pytest tests/unit/mcp/` (448 passed), `mypy`, `ruff check/format`, `audit_public_api_compat.py --check-stale`, and the MCP manifest/freshness checks all green.

Closes #1689

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1701?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added bulk resolution of multiple source references for notebook tools, returning canonical source IDs in the same order they’re provided.
* **Bug Fixes**
  * Avoids repeated lookups by using a single sources listing per batch when needed, and returns full IDs unchanged.
  * Preserves consistent handling for missing, ambiguous, and invalid references.
* **Tests**
  * Added unit tests covering batch behavior, single-lookup execution, order preservation, and error cases.

### DIFF
--- a/src/notebooklm/mcp/_resolve.py
+++ b/src/notebooklm/mcp/_resolve.py
@@ -50,7 +50,7 @@ from ..exceptions import (
 if TYPE_CHECKING:
     from ..client import NotebookLMClient
 
-__all__ = ["resolve_note", "resolve_notebook", "resolve_source"]
+__all__ = ["resolve_note", "resolve_notebook", "resolve_source", "resolve_sources"]
 
 #: A token made only of hex digits and dashes routes to the id/prefix path; any
 #: other character (a space, a letter outside ``a-f``, punctuation) routes to the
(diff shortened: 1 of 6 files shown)


teng-lin:
@claude review

teng-lin:
Addressed in 4e0917f0: documented `resolve_sources`'s empty-input semantics in the docstring — an empty `refs` returns `[]` (not `None`), so call sites must keep their own `if refs else None` guard, since `[]` forwarded to the backend means zero sources (#1652) rather than all. Thanks for the catch.

(CodeRabbit: no actionable comments; cubic: neutral; gemini: not triggered on this PR.)

Issue #1679, opened 2026-06-29
MCP: source_list filtering (status/label) + add-time failure signaling


teng-lin:
## What to build
Two related source-listing gaps from a live session: (1) no way to filter source lists — to find failed sources an agent had to pull all 50 and eyeball `status`; add a `status` filter (and a `label` filter once labels land). (2) Silent import failures: a failed `source_add` (status 3) persists as a ghost row (null url/type) and is not surfaced synchronously — only discoverable via re-list/`source_wait`. Surface add-time failure clearly.

## Acceptance criteria
- [ ] `source_list` accepts an optional `status` filter (e.g. only failed/ready).
- [ ] `source_add` surfaces a failed add more clearly (or documents the async-failure discovery path).
- [ ] Tests.

## Source
src/notebooklm/mcp/tools/sources.py (source_list, source_add). Empirical (Claude Desktop).

## Blocked by
None - can start immediately

PR #1702, opened 2026-06-29
feat(mcp): warn on ready-but-untitled url/youtube source adds (#1698)

### LINKED ISSUE #1698: MCP: dead/soft-404 URLs ingested as status:ready (ghost sources) — content-sanity warning at add time
## Problem
A live MCP session added a deliberately bogus 404 URL (`…/this-url-does-not-exist-zzz-404`). Instead of failing, it was ingested as `status: ready`, titled "Article | Heavy Chain Engineering" — a thin/garbage soft-404 page. A dead link becomes a **ghost source masquerading as success**.

#1679's add-time failure-signaling can't catch this: the backend genuinely returns `status: ready` (not ERROR) for a soft-404 that serves a 200 page. We only `validate_url` for format/SSRF at add time — there is no content-liveness/sanity check.

## Why a naive fix won't work
- We do NOT fetch the URL ourselves (NotebookLM ingests it), so an HTTP-status probe means a separate fetch + latency — and a **soft-404 returns 200 anyway**, so status alone misses it.

## Proposed (preferred): post-add heuristic warning, non-blocking
- After add, inspect the created source's `char_count` (and optionally empty/boilerplate title). When suspiciously thin, surface a `warning` on the add result via the existing `_add_result_payload` channel (where #1679 flags failed imports) — e.g. "Added but the fetched page is very short (N chars); may be a dead link or paywall — verify with source_get_content."
- MUST be a warning, never a rejection (don't break legitimately short sources). Applies to both single-add and the batch (`urls=[...]`) per-item path.
- Threshold = small documented constant.

## Acceptance
- [ ] A thin/soft-404 source add returns a `warning` (single + batch), still `status: ready`.
- [ ] Legit short sources not rejected; no false-failure.
- [ ] Tests for thin-content warning + normal case.

## Source
Empirical (Claude Desktop live session). Related: #1679, #1673.

teng-lin:
## What

Dead/soft-404 URLs can ingest as `status:ready` ghost sources. A soft-404 serves a 200, so #1679's add-time ERROR signaling never fires and the dead link looks like a clean add.

This adds a **zero-RPC, title-only** heuristic: when a `url`/`youtube` add lands READY with a blank/empty title, attach a non-blocking top-level `warning` ("…the fetched page has no title — it may be a dead link or an empty/paywalled page; verify with source_get_content."). It **never rejects** — a legitimately untitled live page keeps importing, just flagged.

## How

- New `_thin_source_warning(source) -> str | None` helper + a single shared `_THIN_SOURCE_WARNING` constant.
- Applied to **`url`/`youtube` single adds only** and the **URL-only batch** path. Deliberately NOT inside the generic `_add_result_payload` — that projector also serves `text`/`drive`/`file`, where a null title is legitimate (e.g. untitled pasted text). Regression-guard tests pin that text/drive/file are not flagged.
- `is_error` warning takes precedence over the thin-title warning (single + batch).
- Single and batch share one constant so the two paths can't drift.

## Scope / known ceiling (deliberate)

Per the locked plan + momus fix: this is **title-only with no extra RPC**. `char_count` lives on `SourceFulltext`, not the `Source` echoed by the add response, so a content-length check would require a separate fetch (rejected here). A `ponytail:` comment in `_thin_source_warning` names this ceiling. Consequence: it catches *empty-title* soft-404s but **not** soft-404s that return a plausible title (e.g. the issue's own "Article | Heavy Chain Engineering" repro). Upgrade path if false-negatives bite = a ready-gated fulltext probe.

## Tests

`tests/unit/mcp/test_sources.py`: single url warn (None/`""`/`"   "` titles), youtube warn, real-title no-warn, **text/drive/file ready+blank-title NO warning** (scope guard), not-ready+blank-title no-warn (single + batch), error-precedence (single + batch), batch warn/no-warn. Full `tests/unit/mcp/` suite green (454 passed); mypy + ruff clean; public-API audit shows no breaks.

Closes #1698

<a href="https://cubic.dev/pr/teng-lin/notebooklm-py/pull/1702?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a warning when a newly added URL or YouTube source finishes as ready but has no meaningful title, helping flag likely dead links.
  * Kept existing error warnings unchanged so failure messages still take priority.
  * Applied the same behavior consistently for both single-item and batch source additions.
  * Ensured the warning only appears for ready URL/YouTube sources and does not affect text, file, or drive sources.

### DIFF
--- a/src/notebooklm/mcp/tools/sources.py
+++ b/src/notebooklm/mcp/tools/sources.py
@@ -84,6 +84,30 @@ def _source_view(source: Any) -> dict[str, Any]:
     return view
 
 
+#: Surfaced when a URL/YouTube add lands READY with no title — the zero-RPC
+#: signal that the page is likely a dead link / empty / paywalled (a soft-404 that
+#: served a 200). Shared by single + batch so the two paths cannot drift. Never a
+#: rejection: a legitimately untitled live page keeps importing, just flagged.
+_THIN_SOURCE_WARNING = (
+    "Added and ready, but the fetched page has no title — it may be a dead link "
+    "or an empty/paywalled page; verify with source_get_content."
+)
+
+
+def _thin_source_warning(source: Any) -> str | None:
+    """Flag a ready-but-untitled add (likely a dead/soft-404 link), else ``None``.
+
+    ponytail: title-only is the ceiling — a zero-RPC heuristic off the add
+    response. It won't catch a soft-404 that returns a plausible title; the only
+    upgrade is a ready-gated fulltext probe (extra RPC), deferred unless
+    false-negatives bite. ``char_count`` lives on ``SourceFulltext``, not the
+    ``Source`` echoed here, so it is not available without that extra fetch.
+    """
+    if source.is_ready and not (source.title or "").strip():
+        return _THIN_SOURCE_WARNING
+    return None
+
+
 def _add_result_payload(source: Any, base: dict[str, Any]) -> dict[str, Any]:
     """Project a ``source_add`` result: enrich the added source + flag failure.
 
@@ -385,7 +409,9 @@ def register(mcp: Any) -> None:
         only AFTER processing. Confirm the outcome with ``source_wait`` or
         ``source_list(status="error")``. When the add response ALREADY reflects a
         failed import, ``source_add`` flags it inline (``status_label="error"`` plus
-        a top-level ``warning``) instead of looking like a clean add.
+        a top-level ``warning``) instead of looking like a clean add. A ``url`` /
+        ``youtube`` add that lands READY with no title is similarly flagged with a
+        ``warning`` (likely a dead link / empty / paywalled page) — never rejected.
 
         **Batch mode** — pass ``urls`` (a list of **http/https URLs**, YouTube links
         included) to add many in one call instead of one round-trip each. Each entry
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
Thanks @claude — addressing the three non-blocking observations:

1. **Batch YouTube coverage comment** — fair; `_add_url_batch` always dispatches `source_type="url"`, so a YouTube URL in `urls=[…]` takes the identical thin-title path. Happy to add the clarifying test comment if this PR is kept (holding off on a CI-churning push while the title-only-vs-`source_wait` approach is being decided with the maintainer).

2. **Batch vs single-mode error-warning text divergence** — confirmed pre-existing (introduced in #1694, the batch PR), not this change. Out of scope here; better as a small follow-up that unifies both `is_error` messages on a shared constant (same anti-drift treatment this PR gives the thin-title text). Can file a tracking issue.

3. **`ponytail:` convention undocumented** — also repo-wide / pre-existing (the marker is used in `_curl_cffi_transport.py` and `_auth/master_token.py`). Documenting it in `docs/development.md`/`CONTRIBUTING.md` is a reasonable separate housekeeping PR rather than scope-creep here.

Note on the CodeRabbit "Linked Issues" pre-merge warning: it flags the absence of a `char_count`/thin-content check. That is the deliberate, maintainer-locked scope of this PR — `char_count` lives on `SourceFulltext`, not the `Source` echoed by the add response, so a content-length check needs an extra RPC. The `ponytail:` comment in `_thin_source_warning` names this ceiling and the upgrade path. An approach that moves the content-sanity check to `source_wait` (where real content is cheaply available) is under separate discussion.

teng-lin:
Superseded by option A (#1698) — the content-sanity check is moving to source_wait where real char_count is available, so it can catch soft-404s that return a plausible title (which the title-only add-time heuristic misses). Closing in favor of the source_wait implementation.

PR #1717, opened 2026-06-30
fix(cli): reject chat configure --mode + custom fields (transport-neutral) (#1716)

### LINKED ISSUE #1716: CLI chat configure: --mode silently drops --response-length/--goal (align with MCP reject)
Surfaced during review of #1715 (MCP chat_mode selector).

`notebooklm chat configure --mode detailed --response-length longer` **silently ignores** `--response-length` because the underlying `execute_configure` short-circuits on `chat_mode` (`_app/chat.py:297`). The CLI's own help even shows that combination as an example (`cli/chat_cmd.py`), so it's actively misleading.

PR #1715 made the **MCP** `chat_configure` tool *reject* the combination with a ValidationError instead of silently dropping the custom fields. The CLI should align — either reject the combo with a clear error, or drop the misleading help example.

Small, transport-neutral fix (the guard could live in `_app.chat.execute_configure` so both CLI and MCP inherit it; MCP's tool-level guard would then be redundant but harmless).

## Source
Review of #1715; `cli/chat_cmd.py` configure_cmd, `_app/chat.py:execute_configure`.

teng-lin:
Closes #1716.

## What

`notebooklm chat configure --mode X --response-length Y` **silently dropped** `--response-length` because `execute_configure` short-circuits on `chat_mode` — and the CLI help even showed that combo as an example. PR #1715 made the *MCP* tool reject the combination; this aligns the CLI.

The fix is **transport-neutral**: the preset-vs-custom mutual-exclusion guard moves into `_app.chat.execute_configure`, so the CLI and the MCP tool now share one rule. The duplicate guard added to the MCP tool in #1715 is removed (its rejection tests still pass via the tool→core path).

## Behavior

- `chat_mode` + a custom `persona`/`response_length` → `ValidationError` ("A chat_mode preset cannot be combined with a custom persona/goal or response_length…").
- Empty `persona=""` stays a no-op (does **not** block a preset); any explicit `response_length` (incl. `"default"`) is a real setting and is rejected.
- **Intentional behavior change**: the CLI now errors on `--mode X --response-length Y` instead of silently applying just the mode. Not a public-API/allowlist break.
- Misleading CLI help example fixed.

## Tests

Core: parametrized persona/response_length rejection + empty-persona-OK (asserts `set_mode` still fires). CLI: the combo exits non-zero with "cannot be combined" and fires no RPC. MCP rejection tests unchanged and still green (now exercising the core guard). Full unit suite green; mypy + ruff clean.

## Note: `uv.lock`

This PR also reconciles `uv.lock` (`ruff 0.15.18 → 0.15.20`) to match the existing `pyproject.toml` pin (`ruff==0.15.20`). The committed lock lagged the pin and `uv lock --check` **failed on main**; this is the corrective lock update (ruff-only diff), not accidental churn.

https://claude.ai/code/session_012ZUzFEAC3XBxwjYQ435QFQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented conflicting chat setup options from being used together. Mode presets now clearly reject custom persona or response-length settings instead of silently ignoring them.
  * Improved empty persona handling so it no longer blocks preset-based chat configuration.
  * The CLI now returns a clear error when incompatible mode and response-length options are combined.

* **Documentation**
  * Updated command examples and usage notes to reflect the valid chat configuration combinations.


teng-lin:
@claude review

teng-lin:
Addressed @claude review in 39ca8a33:
- **Minor 1**: added `client.chat.configure.assert_not_called()` to `test_execute_configure_mode_with_empty_persona_ok`, matching the sibling short-circuit test.
- **Minor 2** (CLI covers only the `--response-length` arm): leaving as-is — both arms are covered at the core (`@parametrize`), and you flagged the single-arm CLI wiring as an acceptable tradeoff.
- **Minor 3** (uv.lock): acknowledged as corrective, no change.

PR #1737, opened 2026-07-02
test: fix auth-load heartbeat CI flake at the source (cheaper tick)

teng-lin:
## Problem

`tests/unit/concurrency/test_auth_load_blocks_loop.py` failed on **macOS Python 3.13** in the post-merge run for #1735 ([job](https://github.com/teng-lin/notebooklm-py/actions/runs/28589016822/job/84767740674)): `only 4 heartbeats fired in 0.5s (expected >= 5)`.

## Diagnosis — flake, not a regression

The two tests prove `AuthTokens.from_storage` / `fetch_tokens_with_domains` run their blocking storage save on a **worker thread** by counting heartbeats from a sibling task during the save: **frozen** loop ⇒ 0–1 ticks, **alive** ⇒ ~10. Evidence it's a flake: 15/16 jobs in the same run passed on identical code; #1735 touches no auth/storage/concurrency files; first failure of this test in the last 15 main runs.

**Root cause (mechanistic):** each tick did `await asyncio.wait_for(stop.wait(), timeout=…)` — which builds a future, arms a timer, and raises+catches a `TimeoutError` *every iteration*. `coverage.py` (CI runs `--cov`) traces all that machinery, so the effective tick period inflated on a contended arm64 runner until the count sagged into the `>= 5` bound. It was **not** the loop actually blocking (it ticked 4×, proving it was alive) and **not** GIL contention (`time.sleep` releases the GIL).

## Fix — at the source, not just the threshold

- **Cheaper tick:** replace the `wait_for(Event.wait())` tick with a bare `asyncio.sleep(interval)` (also *simpler* — drops the try/except). Per-tick overhead stays tiny under coverage, so the healthy count returns to ~10 with wide margin. This removes the actual flake source rather than nudging the bar down inside the noisy regime.
- **Threshold 5 → 3:** kept as a defensive floor cleanly above the 0–1 frozen ceiling — the number is now a floor, not a throughput target.

Frozen-vs-alive detection is unchanged: a genuinely blocked loop still can't fire the sleep during the save, so it stays at the 0–1 baseline and the test still catches the real regression.

The flaked job on main has also been re-run to clear its status.

## Test

`pytest tests/unit/concurrency/` → 26 passed · ruff check + format clean

## Summary by CodeRabbit

* **Tests**
  * Updated concurrency regression coverage to be less timing-sensitive and more reliable across environments.
  * Strengthened the check that the app remains responsive during a delayed save operation.
  * Adjusted heartbeat timing expectations to better reflect normal scheduling behavior.

### DIFF
--- a/tests/unit/concurrency/test_auth_load_blocks_loop.py
+++ b/tests/unit/concurrency/test_auth_load_blocks_loop.py
@@ -16,7 +16,7 @@ This test monkeypatches ``notebooklm._auth.storage.save_cookies_to_storage`` to
 ``time.sleep(0.5)``. While ``AuthTokens.from_storage`` is mid-save, a
 concurrently scheduled async task increments a counter every 50 ms.
 Pre-fix the counter is ~0–1 (loop frozen by the sync sleep); post-fix
-it is >= 5 (the sleep runs on a thread, loop keeps ticking).
+it ticks ~10 times (the sleep runs on a thread, loop keeps spinning).
 
 These are unit-style regression tests under ``tests/unit/concurrency/`` so the
 blocking-I/O contract is exercised without the integration harness.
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude review

teng-lin:
@claude review — updated: fixes the flake at the source (bare `asyncio.sleep` tick instead of `wait_for(Event)`, whose per-tick future/timer/exception machinery coverage.py amplified) rather than only lowering the threshold.

PR #1747, opened 2026-07-02
fix(mcp): chat_configure fails loud on partial/empty custom writes (#1744)

### LINKED ISSUE #1744: MCP: chat_configure fail-loud on partial/empty writes (silent goal/length reset)
Surfaced by the 2026-07-02 multi-model MCP gap review (`docs/notes/mcp-gap-review-2026-07-02.md`). **Med.** The one item every reviewer (Claude ×3 + Codex + agy) independently flagged.

`chat_configure`'s custom branch writes the *full* chat-settings block, defaulting any omitted field to its backend default. So `chat_configure(nb, response_length="longer")` after a prior `chat_configure(nb, goal="tutor")` **silently wipes the goal** back to default. A fully-empty `chat_configure()` call also writes defaults.

- Tool: `src/notebooklm/mcp/tools/chat.py:237`
- Core: `src/notebooklm/_app/chat.py` → `src/notebooklm/_chat/api.py:735` (full-block write, `None→DEFAULT`).

Currently only a docstring warns; a schema-only agent still clobbers prior config.

**Fix (fail-loud — no client getter exists for a true read-merge):**
- In the MCP tool, reject a partial custom write: if exactly one of `goal`/`response_length` is supplied, raise a VALIDATION error telling the agent to pass both (or add an explicit `reset_omitted=true` opt-in).
- Reject a fully-empty `chat_configure()` call.
- A proper read-modify-write is out of scope (needs a GET_NOTEBOOK chat-settings decoder in the neutral core, shared with the CLI) — track separately if desired.

**Constraints:** the write path is shared with the CLI via `execute_configure` — keep CLI behavior consistent (validate at the MCP boundary, or push the fail-loud into the neutral core if it improves the CLI too). Add unit tests under `tests/unit/mcp/`.

teng-lin:
## Summary

`chat_configure`'s custom branch (`goal` / `response_length`) drives `client.chat.configure`, which writes the **full** chat-settings block with no server-side merge (`_chat/api.py`, `None → DEFAULT`). So a partial custom write silently reset the omitted field:

- `chat_configure(nb, response_length="longer")` after a prior `chat_configure(nb, goal="tutor")` **silently wiped the goal** back to default.
- A bare `chat_configure(nb)` silently wrote all defaults.

Only a docstring warned; a schema-only agent still clobbered prior config. This was the one item every reviewer in the 2026-07-02 MCP gap review independently flagged.

## Fix (fail loud at the MCP boundary)

The custom branch now requires **both** `goal` and `response_length` together, and a bare call is rejected — each with a `VALIDATION` error that names the fix. No client getter exists for a true read-modify-write, so this is the fail-loud interim (a GET_NOTEBOOK chat-settings decoder for a real read-merge is out of scope).

- The guard runs **before** `resolve_notebook`, so an invalid call costs no network round-trip.
- "Supplied" mirrors the neutral core exactly: an empty goal (`""`) is a no-op (core uses `if persona:`) and any explicit `response_length` — incl. `"default"` — is a real setting.
- To change only verbosity without a custom goal, use a `chat_mode` preset (`concise` = shorter, `detailed` = longer) — the error message points there.

### Why MCP-boundary, not the shared core

The write path is shared with the CLI via `_app.chat.execute_configure`. The guard stays MCP-only on purpose: the core / CLI **deliberately support** a no-/single-field "no changes" `configure` (pinned by `tests/unit/app/test_app_chat.py::test_execute_configure_no_flags_leaves_goal_and_length_none` and `tests/unit/cli/test_chat.py::test_configure_no_flags_json`). Pushing the guard down would regress a supported CLI affordance. The existing preset-vs-custom mutual-exclusion invariant stays in the core; this "don't silently reset" policy is an agent-safety concern local to the schema-only surface.

The CLI shares the same full-block clobber footgun — **tracked as follow-up, not fixed here** (it would need the read-merge decoder above, or a CLI behavior change out of this issue's scope).

## Tests

`tests/unit/mcp/test_chat.py`:
- Bare call, goal-only, length-only, `response_length="default"`-only, `goal=""`-only, and `goal="" + response_length` (partial) all rejected with no RPC.
- A rejection by notebook **title** asserts `notebooks.list` is never called — pins that the guard short-circuits ahead of resolution.
- Both-supplied and all `chat_mode` preset paths unchanged.

`tests/e2e/test_mcp.py`: `test_configure_then_ask` switched from length-only to `chat_mode="concise"` (same intent, now-valid).

`docs/mcp-guide.md`: `chat_configure` row documents the both-required custom contract.

Reviewed via momus + Codex + agy (converged round 1); Codex findings folded (before-resolve test, `bool(goal)` regression test, docs, CLI follow-up note).

Verified: `mypy src/notebooklm` clean, ruff clean, targeted unit + app + CLI + guardrail suites green.

Closes #1744

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat configuration now rejects incomplete custom updates, preventing accidental partial resets.
  * Custom chat settings must be provided as a complete pair (persona + response length) rather than partial or empty requests.
  * Invalid notebook selection by title is now detected earlier, avoiding unnecessary follow-up actions.
* **Documentation**
  * Clarified chat configuration behavior, including how mode interacts with persona and response length and when to use length presets.
* **Tests**
  * Updated end-to-end and unit tests to cover the new validation rules and error cases.


teng-lin:
@claude review

PR #1770, opened 2026-07-03
fix(mcp): catch upload mkdtemp OSError → clean 500, not raw 500 (#1754)

### LINKED ISSUE #1754: MCP upload route: mkdtemp/OSError before the stream loop becomes a raw 500
Split out of PR #1752 review (claude[bot]) as a pre-existing, out-of-scope gap — not introduced by the jti single-use change.

In `src/notebooklm/mcp/_fileroutes.py`, the upload POST/PUT handler calls `tempfile.mkdtemp(prefix="nblm-mcp-ul-")` at the top of the middle `try` block, **before** the innermost `try` whose `except OSError` handler lives. If `mkdtemp` (or the subsequent `os.open`) raises `OSError` (e.g. ENOSPC), it propagates unhandled and the user sees a raw Starlette **500** instead of a clean error.

Accounting is already safe: the inner `finally` decrements `_inflight_uploads` and the outer `finally` rolls back the jti claim — this is purely a UX/observability concern.

The **download** route already handles + tests this (`test_download_mkdtemp_failure_releases_slot`); the upload route lacks the equivalent handling and test.

**Fix:** catch `OSError` around the upload `mkdtemp`/temp-path setup and return a clean 500/507 (mirroring the download route), plus a `test_upload_mkdtemp_failure_*` regression test.

Refs #1746, PR #1752, ADR-0024.

teng-lin:
## Summary
- The `/files/ul` upload POST/PUT route called `tempfile.mkdtemp()` **outside** the inner spool `try` (whose `except OSError` maps a bad *filename* to a 400), so an `OSError` there (e.g. ENOSPC) escaped as a **raw Starlette 500**.
- Wrap `mkdtemp` in its own `except OSError` returning a clean, header-bearing **500** — a failed dir-create is a server storage problem, not bad input, so it stays distinct from the 400 `os.open` path. `500` (not `507`) because `OSError ⊋ ENOSPC` and it matches the download route's status for the same failure.
- Accounting was already safe — the outer `finally`s release the concurrency slot and roll back the single-use `jti`; the early `return` unwinds through both.

Fixes #1754.

## Test plan
- [x] New `test_upload_mkdtemp_failure_is_clean_500_releases_slot_and_frees_jti` — asserts the caught 500 + body, slot release (`_inflight_uploads` restored), and jti rollback (same link retried → 200). Uses the **default** `TestClient` (`raise_server_exceptions=True`), so an uncaught `OSError` would be re-raised — proving the exception is now genuinely caught.
- [x] Verified the test **fails** against unfixed `main` (raw `OSError` propagates) and **passes** with the fix.
- [x] `tests/unit/mcp/` (706 passed), ruff, mypy, module-size + mcp-boundary guardrails all green.

## Deferred polish findings (out of scope for this minimal fix)
- The **download** route (`/files/dl`) has the same latent `mkdtemp`→raw-500 (its outer `finally` only releases the slot). Symmetry fix is a reasonable follow-up; not changed here to keep the fix minimal and #1754-scoped.
- The sibling `os.open`→400 path omits the `no-store`/`no-referrer` headers the new 500 response carries — a pre-existing inconsistency, not a regression.

https://claude.ai/code/session_01Xay22TTv3tDvaLStyha2PL

## Summary by CodeRabbit

* **Bug Fixes**
  * Upload failures caused by temporary storage creation issues now return a clear server error instead of being mistaken for a client-side processing problem.
  * Upload retries are handled more reliably when a temporary storage error occurs, with upload slots and one-time tokens properly reset.
* **Tests**
  * Added coverage for failed upload storage creation and retry behavior.

### DIFF
--- a/src/notebooklm/mcp/_fileroutes.py
+++ b/src/notebooklm/mcp/_fileroutes.py
@@ -481,7 +481,23 @@ def register_file_routes(mcp: FastMCP, config: FileTransferConfig) -> None:
                 raw_mime = payload.get("mime") or request.headers.get("content-type")
                 mime = raw_mime.split(";")[0].strip() if raw_mime else None
 
-                temp_dir = tempfile.mkdtemp(prefix="nblm-mcp-ul-")  # mkdtemp is 0o700
+                try:
+                    temp_dir = tempfile.mkdtemp(prefix="nblm-mcp-ul-")  # mkdtemp is 0o700
+                except OSError:
+                    # Temp-dir creation failed (e.g. ENOSPC) — the exact temp-disk
+                    # exhaustion the concurrency cap guards against. mkdtemp sits OUTSIDE
+                    # the spool ``try`` below (whose ``except OSError`` maps a bad
+                    # *filename* to a 400), so without this its OSError would escape as a
+                    # raw Starlette 500. A failed dir-create is a SERVER storage problem,
+                    # not bad input, so 500 (mirrors the download route's status) — kept
+                    # distinct from the 400 os.open path. Accounting stays safe: the outer
+                    # finallys release the slot + roll back the jti; no dir was made, so
+                    # nothing to clean.
+                    return PlainTextResponse(
+                        "Upload could not be stored (server storage error).",
+                        status_code=500,
+                        headers={"Cache-Control": "no-store", "Referrer-Policy": "no-referrer"},
+                    )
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
@codex review

PR #1772, opened 2026-07-03
fix(mcp): catch download mkdtemp OSError → clean 500, not raw 500 (#1771)

### LINKED ISSUE #1771: MCP download route: mkdtemp OSError escapes as a raw 500 (symmetry with #1754)
Follow-up split from #1754 (PR #1770), flagged by the claude[bot] review.

In `src/notebooklm/mcp/_fileroutes.py`, the **download** route (`/files/dl`) calls `tempfile.mkdtemp(prefix="nblm-mcp-dl-")` at `download_route` (~line 332). Its outer `finally` correctly releases the concurrency slot on failure (`test_download_mkdtemp_failure_releases_slot` pins this), but an `OSError` there (e.g. ENOSPC) still escapes as a **raw Starlette 500** — it is not wrapped in an `except OSError` returning a clean response.

#1770 fixed the equivalent gap on the **upload** route (clean 500 + regression test). This issue tracks the symmetric fix for download so both `/files/*` routes emit a clean, header-bearing 500 on temp-dir-creation failure.

**Fix:** wrap the download `mkdtemp` in `except OSError` → clean 500 (mirroring the upload route), and update `test_download_mkdtemp_failure_releases_slot` (or add a sibling) to assert the clean 500 body rather than a raw one.

Refs #1754, PR #1770.

teng-lin:
## Summary
- Symmetry follow-up to #1770. The `/files/dl` download route called `tempfile.mkdtemp(prefix="nblm-mcp-dl-")` inside a `try` whose only handlers were `except ValidationError`/`except NotebookLMError` + a `finally` (slot release) — so an `OSError` there (e.g. ENOSPC) escaped as a **raw Starlette 500**.
- Wrap **only** `mkdtemp` in a nested `except OSError` returning a clean, header-bearing **500** (`Cache-Control: no-store` + `Referrer-Policy: no-referrer`), mirroring the upload route. Scoped to `mkdtemp` so post-fetch OSErrors (`FileResponse` init, path resolution) keep their existing path — **not** swallowed.
- The early `return` still unwinds the outer `finally` (`success=False` → slot released; `temp_dir` stays `None` → cleanup no-op), so slot accounting is unchanged.

Fixes #1771.

## Test plan
- [x] Renamed `test_download_mkdtemp_failure_releases_slot` → `_is_clean_500_releases_slot`; now asserts the caught 500 **body** (`"server storage error"`) + slot release via the **default** `TestClient` (`raise_server_exceptions=True`), so an uncaught `OSError` would be re-raised — proving the exception is now caught.
- [x] Verified the test **fails** against `main` (raw `OSError` propagates) and **passes** with the fix.
- [x] `tests/unit/mcp/` (706 passed), ruff, mypy (confirms `temp_dir` narrowing holds), module-size + mcp-boundary guardrails all green.

## Deferred polish findings (out of scope)
- The upload (#1770) and download mkdtemp guards are now near-duplicates; a shared helper is a possible future DRY refactor, not done here to keep the fix minimal.
- agy raised a hypothetical "a *stricter* mypy config might not narrow `temp_dir`" → add `assert temp_dir is not None`. This repo's mypy passes clean and both other reviewers confirmed the narrowing; the assert would be redundant dead code, so not added.

https://claude.ai/code/session_01Xay22TTv3tDvaLStyha2PL

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved download failure handling when temporary storage cannot be created, returning a clearer 500 response instead of an unhandled server error.
  * Added safer cleanup so active download tracking is released correctly even when the download cannot start.
  * Included coverage for the storage-failure path to confirm the response remains user-friendly and the service state stays consistent.

### DIFF
--- a/src/notebooklm/mcp/_fileroutes.py
+++ b/src/notebooklm/mcp/_fileroutes.py
@@ -329,7 +329,21 @@ def register_file_routes(mcp: FastMCP, config: FileTransferConfig) -> None:
         temp_dir: str | None = None
         success = False
         try:
-            temp_dir = tempfile.mkdtemp(prefix="nblm-mcp-dl-")
+            try:
+                temp_dir = tempfile.mkdtemp(prefix="nblm-mcp-dl-")
+            except OSError:
+                # Temp-dir creation failed (e.g. ENOSPC) — the exact temp-disk
+                # exhaustion this cap defends against. Scoped to mkdtemp ONLY (not the
+                # whole try) so it stays distinct from the post-fetch paths; without it
+                # the OSError escapes as a RAW Starlette 500 (the outer finally still
+                # releases the slot, but the response is bare). Return a clean 500,
+                # mirroring the upload route. No dir was created, so the outer finally's
+                # temp_dir cleanup is a no-op and success stays False → slot released.
+                return PlainTextResponse(
+                    "Download could not be prepared (server storage error).",
+                    status_code=500,
+                    headers={"Cache-Control": "no-store", "Referrer-Policy": "no-referrer"},
+                )
             temp_path = os.path.join(temp_dir, f"artifact{spec.extension}")
             try:
                 args: dict[str, object] = {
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
@codex review

Issue #1768, opened 2026-07-03
REST server accepts bearer token via --token CLI flag — leaks through `ps aux` (MCP forbids exactly this)


teng-lin:
## Problem
`server/__main__.py` accepts the bearer token as a **CLI flag**:
```python
parser.add_argument("--token", default=os.environ.get(SERVER_TOKEN_ENV), ...)
...
if args.token:
    os.environ[SERVER_TOKEN_ENV] = args.token
```
A token passed as `notebooklm-server --token <secret>` is visible to any local user via `ps aux` and lands in shell history.

The MCP server deliberately does **not** do this. Its module docstring is explicit:
> All secrets are **env-only** (never CLI flags, so they cannot leak via `ps aux`).

Same class of secret — a bearer fronting a full Google account — with opposite handling across the two servers. The REST side is the weaker posture.

## Fix
Drop the `--token` flag; read the token from `NOTEBOOKLM_SERVER_TOKEN` only, matching MCP. The existing fail-closed `_check_token_configured()` guard already errors clearly when it's unset, so removing the flag doesn't regress UX — it just removes the leaky path.

Add a test asserting `--token` is rejected/absent (a regression guard so the flag can't be reintroduced).

Found while surveying config resolution across the three entry points (see #1765). Sibling: REST also has no `--profile` flag at all — separate issue.

teng-lin:
Superseded by #1767, which handles this more completely: it adds `_reject_argv_token()` (refuses the `--token` flag outright — a token on argv leaks via the process table) **and** a `--token-file` alternative that keeps the secret off the command line. That's strictly better than the "env-only, drop the flag" fix proposed here.

Leaving this open to auto-close when #1767 merges; no separate PR needed.

teng-lin:
Fixed on main by #1767 (merged f76cd4f1): `_reject_argv_token` refuses the `--token` flag and `--token-file` provides the off-argv alternative. Closing as completed.

Issue #1765, opened 2026-07-03
OAuth state_path ignores the --profile CLI flag (only NOTEBOOKLM_PROFILE env)


teng-lin:
Split from review of #1764 (gemini-code-assist, medium). **Pre-existing**, not introduced by #1764 — that PR only added a startup warning about `state_path is None`.

## Problem
`mcp/_oauth.py::get_oauth_config()` resolves the OAuth persistence `state_path` from `NOTEBOOKLM_HOME` + `NOTEBOOKLM_PROFILE` read directly from `os.environ`. But `notebooklm-mcp --profile work --transport http` passes `args.profile` to `create_server(profile=...)` for the client runtime while `get_oauth_config()` never sees it. So with `--profile` set but `NOTEBOOKLM_PROFILE` unset, OAuth state is resolved under the wrong profile (or not persisted at all), diverging from the profile the server actually drives.

Note: the primary OAuth deploy (docker-compose) sets both `NOTEBOOKLM_HOME` and `NOTEBOOKLM_PROFILE` env vars explicitly, so it is unaffected; impact is the `--profile` local-dev-over-http combination.

## Suggested fix
Thread the resolved profile into `get_oauth_config(profile: str | None = None)` and call it from `mcp/__main__.py::main()` as `get_oauth_config(profile=args.profile)`, falling back to `NOTEBOOKLM_PROFILE` when the flag is absent — matching how the client runtime resolves the profile. Add a test for the flag-vs-env precedence. Also reconcile the new startup-warning wording (which currently tells the operator to set `NOTEBOOKLM_HOME`/`NOTEBOOKLM_PROFILE`) with the `--profile` path.

teng-lin:
## Scope: this is a 3-way divergence, not just the `--profile` flag

Investigated while surveying entry-point config resolution. The suggested fix (thread `profile` + fall back to `NOTEBOOKLM_PROFILE`) closes only one of **three** ways `get_oauth_config()` diverges from the profile the server actually drives.

The client runtime resolves through the canonical `paths.resolve_profile(profile)` (precedence: arg → `set_active_profile` → `NOTEBOOKLM_PROFILE` env → config `default_profile` → `"default"`) and `get_profile_dir()` (which defaults home to `~/.notebooklm`). `get_oauth_config()` bypasses all of it with `if home and profile:` reading raw env, so it diverges when:

1. **`--profile` flag set, `NOTEBOOKLM_PROFILE` unset** — the filed case.
[...]

PR #1775, opened 2026-07-03
fix(serve): one loopback/bind guard across mcp+server + REST --profile (#1769)

### LINKED ISSUE #1769: Entry-point config parity: REST server has no --profile; bind/loopback guards duplicated + drifting across mcp/server __main__
## Problem
The three entry points each re-derive shared serve-config independently, and they've drifted. #1765 (OAuth profile divergence) is one symptom; two more:

### 1. REST server has no `--profile` flag
`server/app.py::create_app()` calls `from_storage()` with **no** profile argument, and `server/__main__.py` has no `--profile`. So per-process profile support is:
- `notebooklm` (CLI): full — `--profile`, `use`, env, config.
- `notebooklm-mcp`: `--profile` flag threaded to the runtime (partial — OAuth still diverges, #1765).
- `notebooklm-server`: **none** — only `NOTEBOOKLM_PROFILE` env or `notebooklm use`.

You can't pin a profile per REST-server process without mutating global env. Add `--profile` (default `os.environ.get("NOTEBOOKLM_PROFILE")`) and thread it into `create_app(profile=...)` → `from_storage(profile=...)`, matching the MCP entry point.

### 2. Bind/loopback/log guards are copy-pasted and already diverging
`_LOOPBACK_HOSTNAMES`, `_is_loopback`, `_check_bind_allowed`, `_configure_logging`, `ALLOW_EXTERNAL_BIND_ENV` are duplicated in both `mcp/__main__.py` and `server/__main__.py`. They've already drifted:
- MCP `_is_loopback`: `host.strip().lower()` → `LOCALHOST` / ` localhost ` count as loopback.
- Server `_is_loopback`: bare `host in {...}` → `--host LOCALHOST` gets refused.

Safe direction this time, but this is security-adjacent code maintained in two places — exactly how the *unsafe* drift eventually ships. Extract a shared `mcp`/`server`-neutral serve-guard helper (loopback check + bind guard + log config) so there's one implementation, and add a ratchet/test so a new entry point can't re-copy it.

## Note
These plus #1765 and the token-leak issue are all the same root cause: three entry points, three independent config-resolution paths. This is the real seam behind the "should we consolidate `notebooklm-mcp`/`notebooklm-server` into `notebooklm mcp serve` / `server serve`" question — the CLI sugar is cosmetic; the shared resolution is the bug-preventing win.

teng-lin:
## Summary
Closes the remaining entry-point config gaps from the #1765/#1768/#1769 sweep (#1765 shipped in #1773; #1768 fixed by #1767).

**1. One loopback/bind guard (was drifted 3 ways).** #1767 built a robust, version-independent loopback check in `server/_auth._addr_is_loopback` (unwraps `ipv4_mapped`, so `::ffff:127.0.0.1` is correctly loopback across CPython patch levels — the naive check returned `False` on some macOS 3.10/3.11 runners). But the two **bind-path** guards stayed naive: `server/__main__` and `mcp/__main__` each had their own `_is_loopback` that mishandled the IPv4-mapped form and differed in normalization. New `notebooklm/_serving.py` is the single source (`is_loopback` / `addr_is_loopback` / `check_bind_allowed`); `server/_auth` and both `__main__` modules route through it. Fixes the `::ffff:127.0.0.1` bind-refusal bug on **both** servers.

**2. REST server gains `--profile`.** Only the CLI and MCP could pin an auth profile per process. Added `--profile` (default `$NOTEBOOKLM_PROFILE`), threaded through `create_app(profile=)` → `from_storage(profile=)` — the **default factory only**; an injected `client_factory` stays zero-arg. `main()` also `.strip()`s `--host` once so the guard and the actual bind agree.

## Design notes
- The per-`__main__` helper **names** (`_is_loopback`, `_check_bind_allowed`, `_check_http_bind_allowed`) are kept as **thin delegators** binding each server's label + override env var — this preserves the existing test seams (zero test repointing) while making the *logic* single-source.
- `_serving.py` imports no `click`/`rich`/`cli` (MCP stdio import contract) — enforced by a direct AST test (the dir-scoped mcp/server boundary guards don't cover a top-level module).
- Dropped the now-unused `import ipaddress` from `server/_auth`, `server/__main__`, `mcp/__main__`.

## Test plan
- [x] `tests/unit/test_serving.py` — `is_loopback` incl `::ffff:127.0.0.1` True, `LOCALHOST`/whitespace, `0.0.0.0`/`example.com` False; `check_bind_allowed` empty→SystemExit, non-loopback→SystemExit naming the passed `allow_env`; no-cli AST guard
- [x] `test_main.py` — `--profile` default/env/flag; `main()` passes `profile` to `create_app`
- [x] `test_app_lifecycle.py::test_create_app_default_factory_threads_profile` — `create_app(profile=X)` reaches `from_storage(profile=X)` through the real closure (closes the mocked-`create_app` blind spot)
- [x] existing bind/token/auth tests green (delegator seams preserved); `_auth` request-path loopback tests still pass
- [x] mypy (302 files) + ruff + doc-freshness clean

Closes #1769.

## Summary by CodeRabbit

* **New Features**
  * Added a `--profile` option for the REST server (supports `NOTEBOOKLM_PROFILE`) and propagates it into the app’s default client startup.
  * Improved host handling for both the REST server and MCP transport with shared loopback detection and bind safety checks.

* **Bug Fixes**
  * Prevented empty/whitespace host values from bypassing bind restrictions.
  * Enhanced loopback detection (including IPv4-mapped IPv6 and hostname normalization).

* **Documentation**
  * Updated configuration and architecture docs to reflect the shared bind-validation flow.

* **Tests**
  * Added coverage for profile propagation and the new bind/loopback behavior.


teng-lin:
@claude review

teng-lin:
@codex review

Issue #1791, opened 2026-07-04
server_info health probe targets the wrong profile when the server is launched with an explicit --profile flag (no env/config)


teng-lin:
## Summary

`server_info` (both the MCP tool and the REST `/v1/server/info` route) resolves the
profile it probes/reports via `paths.resolve_profile()`. That precedence chain is:
explicit arg > module `_active_profile` > `NOTEBOOKLM_PROFILE` env > config
`default_profile` > `"default"`.

The server entry points bind a profile by passing `--profile` **only** into
`NotebookLMClient.from_storage(profile=...)`:

- `src/notebooklm/mcp/__main__.py` → `create_server(profile=args.profile)`
- `src/notebooklm/server/__main__.py` → `create_app(profile=args.profile)`

Neither sets `_active_profile` nor exports `NOTEBOOKLM_PROFILE`, and `AuthTokens`
keeps only the resolved `storage_path` (it discards the profile *name*). So when a
user starts the server with an explicit flag and no matching env/config —
`notebooklm-server --profile work` (env unset, no config `default_profile`) — the
live client is bound to `work`, but `resolve_profile()` inside `server_info`
returns `"default"`.

## Impact

This is more than a cosmetic label: `server_info`'s auth-health probe runs against
`get_storage_path(resolve_profile())`, so in this scenario it checks
**`default`'s** `storage_state.json`, not `work`'s. The pre-flight health check can
therefore report `authenticated: false` (or a stale health for the wrong profile)
while the live client is happily authenticated as `work` — undermining the point of
the probe.

This predates #1790 (which only fixed `auth.profile` coming back `null`); that PR
swapped `get_active_profile()` → `resolve_profile()`, which made the reported field
agree with the storage the probe already measured, but did not close this gap.

## Proposed fix

Make the server's bound profile authoritative for `resolve_profile()`. Simplest:
in each server `__main__`, call `paths.set_active_profile(args.profile)` when a
profile is explicitly provided (the server is single-tenant, so a process-wide
active profile is coherent). Alternatively, thread the bound profile through the
server state (MCP `AppState` / FastAPI `app.state`) and have `server_info` read it
from there rather than re-resolving.

Add coverage: start the server with `create_server(profile="work")` /
`create_app(profile="work")` (env + `_active_profile` cleared) and assert both
`auth.profile` **and** the auth-check storage path resolve to `work`.

## Severity

Low–medium. Only affects the explicit-`--profile`-flag-without-env deployment; the
[...]

PR #1792, opened 2026-07-04
fix(server_info): report the resolved profile, not the raw active one (#1790)

### LINKED ISSUE #1790: server_info returns auth.profile=null on a healthy session, undercutting its own diagnostic guidance
## Summary

`server_info`'s docstring states that "the `profile` name + booleans are
sufficient to diagnose auth health." In practice `auth.profile` comes back `null`
even on a fully authenticated, healthy session — so half of that guidance is
inert.

## Repro

```
server_info(include_account=true)
# ->
# {
#   "server": "notebooklm",
#   "version": "0.8.0",
#   "auth": {
#     "authenticated": true,
#     "storage_exists": true,
#     "json_valid": true,
#     "cookies_present": true,
#     "sid_cookie": true,
#     "profile": null            <-- null despite authenticated=true
#   },
#   "account": { "email": "…", "authuser": 0, "available": true,
#                "notebook_limit": 100, "source_limit": 50,
#                "output_language": null }
# }
```

## Proposed fix

Either:

1. Populate `auth.profile` with the active storage profile name when one is in
   use, or
2. Document that `null` is the normal/expected value when no named profile is
   configured, and adjust the docstring so it doesn't lean on `profile` as a
   health signal.

## Why it matters

Low severity, but the docstring currently points users at a field that carries no
information in the common case.

## Environment

- notebooklm-py `v0.8.0a1`, remote (http) MCP connector, server `0.8.0`

teng-lin:
## Summary

Closes #1790.

`server_info` — the MCP `server_info` tool and the REST `GET /v1/server/info`
route — reported `auth.profile` as `null` even on a fully authenticated, healthy
session. The docstring pointed diagnostics at that field, so the guidance was
inert.

Root cause: both call sites read `paths.get_active_profile()`, which returns the
raw module-level `_active_profile`. That global is only ever set by the CLI's
`--profile` flag / `use` command; the long-running MCP/REST server never sets it,
so it was always `None`. Meanwhile the auth-health probe's storage path was
already resolved via `resolve_profile()` (inside `get_storage_path`), so the
*reported* profile was inconsistent with the profile the probe actually ran
against.

## Fix

- Swap both call sites (`mcp/tools/meta.py`, `server/routes/meta.py`) from
  `get_active_profile()` to `resolve_profile()`. `resolve_profile()` never
  returns `None` (falls back to `"default"`), follows the full precedence chain,
  and is the same value the CLI `status` command reports — so `auth.profile` now
  names the profile the probe measured.
- Tighten the MCP docstring so it no longer over-claims `profile` as a
  "sufficient" health signal: the booleans are the real signals, `profile` is a
  label for *which* profile was probed.

## Tests

- `tests/unit/mcp/test_meta.py`: `auth.profile` is the resolved profile
  (`"default"` with no named profile) and reflects a `NOTEBOOKLM_PROFILE`-named
  profile. Both fail against the old `get_active_profile()` code (verified).
- `tests/server/test_meta.py`: parallel assertion for the REST route.
- Test isolation uses object-target `monkeypatch.setattr(paths, "_active_profile", None)`
  (ADR-0007 forbids string-target monkeypatch) plus the autouse
  `NOTEBOOKLM_HOME` tmp-dir fixture and `NOTEBOOKLM_PROFILE` control.

## Verification

- `mypy src/notebooklm --ignore-missing-imports` — clean
- `ruff check . && ruff format --check .` — clean (whole tree)
- `pytest tests/unit/mcp tests/server` — 1002 passed
- ADR-0025 schema-char budget + ADR-0007 monkeypatch guardrails — pass

## Deferred (out of scope)

A pre-existing, separate correctness gap surfaced during review: a server started
with an explicit `--profile X` flag (no matching `NOTEBOOKLM_PROFILE` env / config
`default_profile`) binds the client to `X` but `resolve_profile()` still returns
`"default"`, so the probe checks the wrong storage. This predates #1790 and is
tracked in #1791.

## Summary by CodeRabbit

* **Bug Fixes**
  * The server info response now reports the actual resolved profile used by the auth health check, preventing a null profile value in healthy sessions.
  * The reported auth profile now matches the CLI status output more consistently.

* **Tests**
  * Added coverage to verify server info returns a resolved, non-null profile in both default and named profile scenarios.

### DIFF
--- a/src/notebooklm/mcp/tools/meta.py
+++ b/src/notebooklm/mcp/tools/meta.py
@@ -27,7 +27,7 @@ from fastmcp import Context
 from ..._app.auth_check import AuthCheckPlan, run_auth_check
 from ..._version_info import version_string
 from ...exceptions import NotebookLMError
-from ...paths import get_active_profile, get_storage_path
+from ...paths import get_storage_path, resolve_profile
 from .._confirm import READ_ONLY
 from .._context import get_client
 from .._errors import mcp_errors, redact
@@ -120,13 +120,21 @@ def register(mcp: Any) -> None:
         fields degrade to ``{available: False, reason: ...}`` (identity still
         included) rather than failing the whole call.
 
+        ``profile`` names the resolved storage profile the probe ran against
+        (e.g. ``"default"``); the booleans are the actual health signals.
+
         The absolute on-disk storage path is deliberately **not** returned: it
         leaks the server-host OS username / filesystem layout to any (possibly
-        remote) caller, while telling the agent nothing it can act on. The
-        ``profile`` name + booleans are sufficient to diagnose auth health.
+        remote) caller, while telling the agent nothing it can act on.
         """
         with mcp_errors():
-            profile = get_active_profile()
+            # Report the *resolved* profile (never ``None``) rather than the raw
+            # module-level active profile: the MCP server never sets an active
+            # profile, so ``get_active_profile()`` was always ``None`` here even
+            # though ``get_storage_path`` resolves and checks a concrete profile
+            # (#1790). ``resolve_profile()`` names the profile the auth probe
+            # actually ran against — the same value the CLI ``status`` reports.
+            profile = resolve_profile()
             storage_path = get_storage_path(profile)
             plan = AuthCheckPlan(
                 storage_path=storage_path,
--- a/src/notebooklm/server/routes/meta.py
+++ b/src/notebooklm/server/routes/meta.py
@@ -33,7 +33,7 @@ from ..._redact import redact
 from ..._version_info import version_string
 from ...client import NotebookLMClient
 from ...exceptions import NotebookLMError
-from ...paths import get_active_profile, get_storage_path
+from ...paths import get_storage_path, resolve_profile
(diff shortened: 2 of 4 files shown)


teng-lin:
@claude review

teng-lin:
Thanks @claude — noted all three minor observations; none require a change here:

1. **Issue number in inline comments** — kept intentionally: the comment explains a non-obvious regression class and the `#1790`/`#1791` refs anchor the "why" for the next maintainer. Consistent with the existing convention of referencing the resolved issue in fix comments.
2. **REST coverage symmetry** — the named-profile case is exercised through the shared `resolve_profile()` path by the MCP `test_server_info_profile_reflects_named_profile`; the REST route adds no logic of its own, so I left the REST suite at the not-null assertion to avoid a redundant test.
3. **Comment length** — it documents genuinely subtle `_active_profile`-is-always-`None`-in-server behavior, so I'm keeping it explicit over terse.

CodeRabbit reported no actionable comments. The deferred `--profile`-startup gap is tracked in #1791.

PR #1795, opened 2026-07-04
fix(mcp): rename research id params to poll_task_id, alias old names (#1789)

### LINKED ISSUE #1789: Research tools
## Summary

The research tools (`research_start`, `research_status`, `research_import`,
`research_cancel`) pass several distinct identifiers through parameters whose
names don't consistently match the field the value comes from. The docstrings
spend significant prose warning callers which id goes where — a signal that the
parameter naming itself is the UX problem.

Observed while reviewing `v0.8.0a1` tool schemas during live testing. (Not
triggered against the live API to avoid kicking off a real research run.)

## The overload, as documented

- `research_cancel(run_id=…)` wants `research_status.task_id`.
  - **But** for a *deep* run that value is `research_start`'s `report_id`, **not**
    its `task_id` — which is itself a `sessionId`. The docstring explicitly says to
    prefer the `research_status` value "to avoid the trap."
- `research_import(task_id=…)` wants the `poll_task_id` from
  `research_start` / `research_status`, not the `task_id` field name it's called.
- `research_status(task_id=…)` disambiguates by `poll_task_id`.

So three different tools accept an argument named `task_id`/`run_id` that actually
wants a `poll_task_id` (or `report_id`, in the deep-run case). Copy-pasting the
field literally named `task_id` from one tool's output into another's `task_id`
input can silently target the wrong task.

## Proposed fix

If the underlying ids can't be unified, **rename the parameters to match their
source field verbatim** so copy-paste "just works":

- `research_import(poll_task_id=…)` instead of `task_id`
- `research_cancel(poll_task_id=…)` instead of `run_id`
- `research_status(poll_task_id=…)` for the pin

Keep the old names as accepted aliases for one release with a deprecation note.

Alternatively, return a single canonical `run_ref` from `research_start` /
`research_status` that every other research tool accepts, and stop surfacing the
raw `task_id`/`report_id`/`sessionId` distinction in the public contract.

## Environment

- notebooklm-py `v0.8.0a1`, remote (http) MCP connector, server `0.8.0`

teng-lin:
## Summary

Closes #1789.

The research MCP tools passed the single id that `research_start` surfaces as
`poll_task_id` under mismatched parameter names — `task_id` on
`research_status` / `research_import`, `run_id` on `research_cancel` — which
forced long docstring warnings about which id goes where and invited copy-paste
mistakes (pasting a fast run's `task_id` where a deep run needs its `report_id`,
etc.).

All three tools now accept the id under the **same** name, `poll_task_id`, so the
value copies verbatim from one tool's output into the next.

## What changed

- **`research_status`** — pin param `task_id` → `poll_task_id`.
- **`research_import`** — `task_id` → `poll_task_id`.
- **`research_cancel`** — `run_id` → `poll_task_id`.
- The old `task_id` / `run_id` names remain **accepted aliases for one release**
  (removed in **v0.9.0**). Passing an alias resolves identically, emits a gated
  `DeprecationWarning`, and adds a caller-visible `deprecation` note to the tool
  result (FastMCP swallows the Python warning at the tool boundary, so the
  agent-visible note is the real signal). Passing both names with *different*
  values is a validation error; same value → canonical wins silently.
- Responses surface both the canonical `poll_task_id` and the legacy key
  (`run_id` / `task_id`) for one release, matching `research_status`'s existing
  dual-key response.
- A small `_resolve_poll_task_id(...)` helper centralizes the alias fold /
  warning / note / conflict check.

## Budget & docs

- The now-redundant id-routing docstring prose is trimmed so the extra alias
  param stays within the ADR-0025 schema-char budget (measured **36934 ≤ 37020**).
- `docs/mcp-guide.md`, `docs/deprecations.md`, and `CHANGELOG.md` updated.

## Tests

- Helper-level `pytest.warns(DeprecationWarning)` (deterministic; the tool
  boundary swallows warnings) + conflict / same-value / canonical cases.
- Through-the-tool alias tests asserting the alias resolves to identical
  behavior and surfaces the `deprecation` note, for all three tools.
- Migrated existing unit / mcp_vcr / e2e tests to the canonical `poll_task_id`.

## Verification

- `uv run mypy src/notebooklm --ignore-missing-imports` — clean (302 files)
- `uv run ruff check . && uv run ruff format --check .` — clean
- `uv run pytest` — **11689 passed, 77 skipped, 1 xfailed**
- `scripts/check_deprecation_targets.py` + `scripts/audit_public_api_compat.py` — clean (no new API breaks)

## Summary by CodeRabbit

* **New Features**
  * MCP research tools now consistently use a single `poll_task_id` field across status, import, and cancel workflows.

* **Bug Fixes**
  * Legacy `task_id` / `run_id` aliases remain supported but now emit deprecation warnings and include deprecation notes in responses when used.
  * Clear validation errors occur when canonical and alias identifiers are provided with different values; empty/whitespace identifiers are rejected.
  * Cancel behavior and returned identifiers are now consistent with the canonical `poll_task_id`.

* **Documentation**
  * Updated the MCP guide and changelog to reflect `poll_task_id` and the deprecations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **MCP research tools now accept the poll id under one name, `poll_task_id`.**
+  `research_status`, `research_import`, and `research_cancel` previously took the
+  value that `research_start` / `research_status` surface as `poll_task_id` under
+  mismatched parameter names (`task_id` on status/import, `run_id` on cancel),
+  forcing docstring warnings about which id goes where and inviting copy-paste
+  mistakes. All three now accept it as `poll_task_id`, so the value pastes
+  verbatim from one tool's output into the next. The old `task_id` / `run_id`
+  names still work as **deprecated aliases** (removed in v0.9.0): passing one
+  emits a `DeprecationWarning` and adds a `deprecation` note to the result;
+  passing both names with different values is a validation error.
+  ([#1789](https://github.com/teng-lin/notebooklm-py/issues/1789))
+
 ### Fixed
 
 - **The `.mcpb` desktop bundle now ships for pre-releases and launches the
--- a/docs/deprecations.md
+++ b/docs/deprecations.md
@@ -21,6 +21,7 @@ the broader stability policy (semver promise, supported Python versions, the
 | Deprecated | Replacement | Since | Removal | Notes |
 |------------|-------------|-------|---------|-------|
 | Awaiting `NotebookLMClient.from_storage(...)` | `async with NotebookLMClient.from_storage(...) as client:` | v0.5.0 | v1.0 | The `__await__` form still works. Warning emitted via `src/notebooklm/_deprecation.py::warn_deprecated`; suppress with `NOTEBOOKLM_QUIET_DEPRECATIONS=1` ([#1369](https://github.com/teng-lin/notebooklm-py/issues/1369)) |
+| MCP `research_status(task_id=…)` / `research_import(task_id=…)` / `research_cancel(run_id=…)` | The same value under `poll_task_id=…` on all three | v0.8.0 | v0.9.0 | The three tools each accept the id that `research_start` / `research_status` surface as `poll_task_id` — renamed so the value copies verbatim between tools. The old `task_id` / `run_id` param names still work as aliases but emit a `DeprecationWarning` (via `warn_deprecated`) and add a `deprecation` note to the tool result; passing both names with different values is a validation error. ([#1789](https://github.com/teng-lin/notebooklm-py/issues/1789)) |
 
 > The v0.8.0 error-contract runways (`get()`-returns-`None`, the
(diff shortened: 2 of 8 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the review. Addressed in 72696c92:

- **`warn_deprecated` stacklevel** — now passes `stacklevel=4` (chain is `warn_deprecated` → `_resolve_poll_task_id` → tool coroutine → caller), so a real consumer's warning points at the tool boundary instead of the helper.
- **Whitespace-only alias fired a spurious warning before rejection** — the helper now only emits the `DeprecationWarning` + note for a *substantive* alias (`alias.strip()`); a blank/whitespace alias is handed back unwarned so the tool's own empty-id guard rejects it with no deprecation signal spent. Added a helper-level test (blank alias → no warning) and a through-tool test (`task_id="   "` → VALIDATION, no warning, no RPC).
- **`research_status` echoes `result.task_id` for `poll_task_id`** — agreed this is pre-existing (the backend `task_id` field vs the input pin) and out of scope for this rename; left unchanged, worth tracking separately if a deep-run status VCR cassette ever surfaces it.

PR #1854, opened 2026-07-09
fix(sources): decode Drive-hosted PDF type_code 14 by MIME, not URL (#1832)

### LINKED ISSUE #1832: Drive source_list still mislabels Drive-hosted PDFs as google_spreadsheet (disambiguate type code 14 at decode)
Follow-up to #1828 / PR #1831.

## Context

PR #1831 fixed the **add** and **add-and-wait** paths for Drive PDFs: `execute_source_add_drive` stamps the returned `Source._type_code` from the caller's declared `mime_type`, and `source_add_and_wait` re-stamps the waited source before aggregating. So `source_add(source_type="drive", mime_type="pdf")` now reports `kind="pdf"`.

**What it does NOT fix:** the plain **`source_list` / `GET_NOTEBOOK` decode path**. There, the only signal is the raw `_type_code`, and per #1828 the backend returns `14` for Drive-hosted PDFs — which `_SOURCE_TYPE_CODE_MAP` maps to `GOOGLE_SPREADSHEET`. With no declared mime available at list time, a Drive PDF still lists as `kind="google_spreadsheet"`.

## Root cause

`_type_code == 14` is **ambiguous for Drive sources**: it is the code for a native Google Sheet (`docs/rpc-reference.md:104`, from a live probe) **and**, per the #1828 reporter's observations, the code returned for Drive-hosted PDFs. `Source.kind` is a pure function of `_type_code` (`src/notebooklm/_types/sources.py:151-153`), so it cannot distinguish the two.

## Why NOT to just remap 14

Blanket-remapping `14` away from `GOOGLE_SPREADSHEET` would **relabel real Google Sheets** and is a breaking change to a documented contract (Sheets consumers rely on `kind=google_spreadsheet`). Do not do this.

## Proposed API-safe fix: disambiguate 14 by source URL at decode time

A native Google Sheet and a Drive-hosted file carry different URLs:
- Sheet → `https://docs.google.com/spreadsheets/d/<id>/…`
- Drive file (PDF/other) → `https://drive.google.com/file/d/<id>/…` (exact shape **to be confirmed by capture**)

When `_type_code == 14`, decode `kind` using the source URL: a spreadsheets URL → `GOOGLE_SPREADSHEET`; a Drive-file URL → `PDF` (or a more precise Drive-file type if one exists). This corrects **only** the buggy Drive-PDF case and leaves real Sheets untouched — so it is a **bug fix, not a breaking change** (nobody should rely on a PDF being labeled a spreadsheet). It also naturally covers the list path, which the declared-mime approach can't reach.

`Source` already exposes both `_type_code` and `url`, so `kind` (or the `SourceRow` decode) can incorporate the URL.

## Blocker — needs a live capture first

Per repo policy (don't guess a wire shape ahead of a live capture), this needs a real capture from an authed account to lock down:

1. The exact **URL** a Drive-hosted PDF source carries in `GET_NOTEBOOK` (vs. a native Sheet).
2. Confirmation of whether **all** Drive-hosted non-native files bucket under `14`, or only PDFs (do a Drive `google-sheets` add and a Drive `pdf` add and compare rows).

Suggested: `scripts/scrub_rpc_har.py` on a `GET_NOTEBOOK` that lists a Drive PDF + a real Google Sheet, and paste both source rows here.

## Acceptance criteria

- [ ] With a Drive-PDF capture in hand, `kind` decodes `_type_code == 14` by URL: Drive-file URL → `PDF` (or precise type), spreadsheets URL → `GOOGLE_SPREADSHEET`.
- [ ] Real Google Sheets continue to decode as `GOOGLE_SPREADSHEET` (no regression / no API break).
- [ ] `source_list(detail="full")` on a Drive PDF reports `kind="pdf"`, matching `source_add`.
- [ ] Regression tests for both the Sheet and Drive-PDF `_type_code=14` rows; update `docs/rpc-reference.md` to note the `14` overload.
- [ ] Add a `stability`/changelog note that Drive-PDF `kind` was corrected from `google_spreadsheet` → `pdf`.

## Links
- Parent bug: #1828 (add-path fixed by #1831)
- PR: #1831

teng-lin:
## Summary

Fixes #1832. The NotebookLM backend returns source `type_code == 14` for **both** a native Google Sheet **and** a Drive-hosted PDF, so the read paths (`source_list` / `GET_NOTEBOOK` **and** `source fulltext` / `source_read` / `GET_SOURCE`) mislabeled Drive PDFs as `kind="google_spreadsheet"`. This completes the **read-path** half of the fix that #1831 started on the add path.

## Why not disambiguate by URL (the issue's original proposal)

A live capture (authed account) showed a Drive-hosted source carries **no URL on the wire** — `metadata[0]`, `[5]`, `[7]` (the URL precedence slots the adapter reads) are all `null`, and there's no `drive.google.com/file/d/…` anywhere in the row. So URL-based disambiguation has nothing to key on.

## The real signal: MIME

The wire already carries the true type as MIME, at the same positions for every Drive-hosted source:

| type_code | `metadata[19]` | `metadata[9]` (drive-file descriptor) |
|---|---|---|
| Drive PDF `14` | `"application/pdf"` | `["<id>", 5, "application/pdf", ""]` |
| Native Sheet `14` | `"application/vnd.google-apps.spreadsheet"` | `["<id>", 8, "application/vnd.google-apps.spreadsheet", ""]` |

## Change

- **`SourceRow.mime`** — reads `metadata[19]`, falling back to the drive-file descriptor `metadata[9][2]`.
- **`_disambiguate_type_code(type_code, mime)`** — for `type_code == 14`, remaps only live-verified MIMEs (`_TYPE_CODE_14_MIME_OVERRIDE = {"application/pdf": 3}`); native Sheets (google-apps mime), no MIME, and unrecognized MIMEs all stay `GOOGLE_SPREADSHEET` (conservative — never relabels a real Sheet, never emits `UNKNOWN`).
- Wired into `Source.from_row` (list path) **and** `SourceContentRenderer.get_fulltext` (the `GET_SOURCE` / `source_read` path, live-confirmed to carry the same `type_code=14 + mime` shape), and complements #1831's declared-mime stamping on the **add path** (no conflict — all resolve a Drive PDF to `PDF`).
- Position constants + canary contract (`TestSourceRowPositionContract`) updated for `metadata[9]`/`[19]`.
- `docs/rpc-reference.md` documents the code-`14` overload; CHANGELOG note added.

## Testing

- Regression tests for **both** `type_code=14` rows (Drive PDF → `pdf`, native Sheet → `google_spreadsheet`), plus MIME-accessor and conservative-fallback cases.
- Full unit + guardrail suite green (10,723 passed); mypy + ruff clean.
- **Verified live** against a real notebook: Drive PDF now lists as `kind=pdf`, native Sheet stays `google_spreadsheet`, native Doc unchanged.

> Note: the issue is labeled `do-not-implement-yet`; opening this PR was explicitly authorized.

## Summary by CodeRabbit

* **Bug Fixes**
  * Drive-hosted PDFs are now correctly identified as PDFs instead of spreadsheets.
  * Source and full-text views now use MIME information to resolve ambiguous source types more accurately.
* **Documentation**
  * Updated the source type code reference to explain the overloaded type and how it is distinguished.
  * Clarified metadata expectations for source records.
* **Tests**
  * Added coverage for PDF and spreadsheet disambiguation, plus fallback behavior when MIME data is missing or unexpected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Drive-hosted PDFs no longer list as `google_spreadsheet`.** The backend
+  returns type code `14` for both native Google Sheets and Drive-hosted PDFs,
+  and Drive sources carry no URL, so the read paths (`source_list` /
+  `GET_NOTEBOOK` **and** `source fulltext` / `source_read` / `GET_SOURCE`)
+  mislabeled Drive PDFs as `kind="google_spreadsheet"`. `kind` now disambiguates
+  code `14` by the source MIME (`metadata[19]` / `metadata[9][2]` from a live
+  capture): `application/pdf` → `pdf`, while real Google Sheets
+  (`application/vnd.google-apps.spreadsheet`) are unchanged. This completes the
+  read-path half of the fix started in #1831 (the add path).
+  ([#1832](https://github.com/teng-lin/notebooklm-py/issues/1832))
+
 ## [0.8.0]
 
 The headline of 0.8.0 is **integrations**: NotebookLM is now reachable from AI
--- a/docs/rpc-reference.md
+++ b/docs/rpc-reference.md
@@ -101,11 +101,13 @@ Internal integer codes returned by `GET_NOTEBOOK` / `LIST_SOURCES` and consumed 
 | 10 | `MEDIA` | Audio / video upload |
 | 11 | `DOCX` | Word document |
 | 13 | `IMAGE` | Image upload |
-| 14 | `GOOGLE_SPREADSHEET` | Google Sheets source |
+| 14 | `GOOGLE_SPREADSHEET` | Google Sheets source **and** Drive-hosted binaries (see overload note) |
 | 16 | `CSV` | CSV upload |
 | 17 | `EPUB` | EPUB upload (added in v0.4.0) |
 
 > Codes outside this map are surfaced as `SourceType.UNKNOWN` and emit `UnknownTypeWarning` on first occurrence so unmapped types don't crash callers.
+
+> **Code `14` is overloaded** (live-captured #1828/#1832): the backend returns `14` for a native Google Sheet *and* for a Drive-hosted PDF. Drive sources carry no URL (`metadata[0]/[5]/[7]` are all null), so the two are disambiguated by the MIME at `metadata[19]` (fallback `metadata[9][2]`): `application/vnd.google-apps.spreadsheet` → `GOOGLE_SPREADSHEET`, `application/pdf` → `PDF`. See `_disambiguate_type_code` in `src/notebooklm/_types/sources.py`.
 
 ---
 
--- a/src/notebooklm/_row_adapters/sources.py
+++ b/src/notebooklm/_row_adapters/sources.py
@@ -102,6 +102,11 @@ class SourceRow:
     7      url block; ``[7][0]`` is the canonical source URL when
            present (takes precedence over ``metadata[5][0]`` and
            ``metadata[0]``).
+    9      Drive-file descriptor for Drive-hosted sources:
+           ``[drive_id, kind_int, mime, ""]``; ``[9][2]`` is the MIME.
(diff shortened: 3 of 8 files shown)


teng-lin:
@claude review

Issue #1875, opened 2026-07-12
concurrency: null-conversation chat asks do not serialize with explicit follow-ups


teng-lin:
> Surfaced by a 3-model codebase review (codex, verified).

## Problem
`chat.ask` with `conversation_id=None` locks by **notebook**, while explicit follow-ups lock by **conversation id**, so two asks targeting the same current conversation do not serialize.

## Evidence
- `src/notebooklm/_chat/api.py:442/446/461` — the two lock paths.

## Failure scenario
Task A: `chat.ask(nb, "Q1")` (no `conversation_id`). Task B (concurrent): `chat.ask(nb, "Q2", conversation_id=current_id)`. B can build/send stale history before Q1 lands → divergent local vs server conversation ordering.

## Suggested fix
Normalize a null-conversation ask to the resolved current-conversation lock key before acquiring, so both paths serialize on the same key.

Issue #1870, opened 2026-07-12
perf: multi-source wait fans out N full-notebook GET_NOTEBOOK polls (O(N^2))


teng-lin:
> Surfaced by a 3-model codebase review (Claude / codex / agy). **2/3 models flagged this** (codex: high, Claude: medium), both verified the fan-out path.

## Problem
Waiting on multiple sources fans out one poller per source, and each poller re-fetches the **entire** notebook source list — O(N²) payload growth and N× the RPC pressure per poll tick.

## Evidence
- `src/notebooklm/_source/polling.py:63/159` — per-source pollers each call `wait_until_ready` → `get_or_none`.
- `src/notebooklm/_sources.py:247` — `get_source=self.get_or_none`.
- `src/notebooklm/_source/listing.py:45` — `get_or_none` issues a fresh `GET_NOTEBOOK` returning the whole source list.

## Failure scenario
`source_wait` / batch add over a 100-source notebook issues ~100 full-source-list `GET_NOTEBOOK` calls **per poll tick**, amplifying load and rate-limit risk (and interacting badly with #the-rate-limit paths).

## Suggested fix
Fetch the notebook source list **once per poll tick** and dispatch statuses to the waiting sources from that single snapshot, instead of one full fetch per source.

teng-lin:
Resolved by #1882 (merged as 44bd9fb5): multi-source wait now fetches the notebook source list ONCE per poll tick (single-snapshot loop in `_source/polling.py`) and dispatches to all pending sources, replacing the O(N²) per-source GET_NOTEBOOK fan-out. #1877's MAX_WAIT_SOURCE_IDS backstop + input-order + per-source outcomes are preserved; a one-list-per-tick test pins it.

PR #1876, opened 2026-07-12
fix(mcp): reject non-loopback Host headers on the loopback HTTP transport (#1869)

### LINKED ISSUE #1869: security: MCP loopback HTTP has no Host guard — DNS-rebinding exposure
> Surfaced by a 3-model codebase review (Claude / codex / agy). **2/3 models independently flagged this** (agy: critical, codex: high), both verified against the REST server.

## Problem
The MCP HTTP server, when bound to loopback, starts with `auth=None` (no bearer token required) **and has no `Host`/`Origin` validation** — unlike the REST server, which guards both the unspoofable peer address and the `Host` header (`src/notebooklm/server/_auth.py:139-149`). This leaves the loopback MCP HTTP endpoint open to a **DNS-rebinding** attack.

## Evidence
- `src/notebooklm/mcp/__main__.py:104` — loopback binds skip auth.
- `src/notebooklm/mcp/server.py:124` — `FastMCP(..., auth=auth)` is built with no Host-guard middleware; targeted search found no `Host`/`Origin` check on the MCP app.
- REST has the guard the MCP side is missing: `server/_auth.py` `_peer_is_loopback` + `_host_is_loopback`.

## Failure scenario
A user runs `notebooklm-mcp --transport http --host 127.0.0.1` without `NOTEBOOKLM_MCP_TOKEN`. They visit a malicious page; the attacker rebinds `evil.example → 127.0.0.1` (short TTL) and sends MCP JSON-RPC with `Host: evil.example`, driving tools — source reads, chat, `*_delete`, even `share_set_access(..., confirm=true)` — against the victim's account.

## Suggested fix
Mirror the REST guard onto the MCP HTTP ASGI app: reject requests whose `Host` header is not loopback (and, where reachable, enforce the loopback peer address), gated by the same external-bind escape hatch. Add a hostile-`Host` MCP test (the current `tests/unit/mcp/test_entrypoint.py:90` only asserts `auth is None` on loopback).

teng-lin:
Fixes #1869. **Security — DNS rebinding.**

## Problem
The MCP HTTP transport skips bearer auth on a loopback bind (matching the REST server's local-dev posture) but — unlike the REST server (`server/_auth.py`) — had **no `Host`-header validation**. A malicious page that resolves its own domain to `127.0.0.1` can then drive the local MCP server with `Host: evil.example`, reaching every tool: source reads, chat, `*_delete`, even `share_set_access(..., confirm=true)`.

Independently flagged by 2 of 3 review models (agy: critical, codex: high), both contrasting it against the REST guard the MCP side was missing.

## Fix
- New `LoopbackHostGuardMiddleware` (pure ASGI) rejects any request whose `Host` header isn't a loopback literal, wired into `server.run(transport="http", …, middleware=[…])` (FastMCP forwards `middleware` to `run_http_async`).
- Skipped when `NOTEBOOKLM_MCP_ALLOW_EXTERNAL_BIND=1` — that path already **mandates** bearer/OAuth auth via `_check_http_auth_required`, which owns access control there.
- The loopback `Host` parsing (optional `:port`, bracketed IPv6 `[::1]`) lives in the neutral **`_serving`** module as `host_header_is_loopback`, so the CLI/MCP/server import boundary stays intact (no `mcp → server` import).

## Test plan
- [x] `host_header_is_loopback`: loopback variants (with/without port, `localhost`, `[::1]`, `[::1]:9420`) pass; `evil.example`, `127.0.0.1.evil.com`, `[::1]evil.com`, `0.0.0.0`, empty rejected.
- [x] ASGI middleware behavior: hostile `Host` → `403` **without reaching the wrapped app**; loopback passes through; `allow_external=True` and non-http (lifespan/ws) scopes bypass.
- [x] Entrypoint wiring: the guard is installed on every http run with the correct `allow_external` (False on loopback, True on external bind).
- [x] ruff + mypy clean; import-boundary guardrail green.

The current `tests/unit/mcp/test_entrypoint.py` previously only asserted `auth is None` on loopback (which this PR now complements with the Host guard).

Follow-up for #1870–#1875 (the other review findings) tracked separately.

## Summary by CodeRabbit

* **Security Enhancements**
  * Added a loopback `Host` header guard to the loopback-bound MCP HTTP transport to help prevent DNS-rebinding-style access.
  * Requests with empty/malformed or non-loopback `Host` values are rejected with HTTP 403.
  * Loopback `Host` values (IPv4/IPv6, including optional ports) continue to be allowed.
  * When external binding is enabled, the guard is installed in bypass mode to avoid blocking external hosts.

* **Tests**
  * Added unit test coverage for host validation and bypass behavior.

* **Documentation**
  * Updated architecture notes to describe the new host-header guard.


teng-lin:
@claude review

teng-lin:
@codex review

PR #1886, opened 2026-07-13
test(sources): regression guard for #1878 overall-duration bound

### LINKED ISSUE #1878: source_wait: bound overall request duration, not just per-source timeout
## Context
Split out from the #1877 review (Codex P2).

`_app.source_wait.wait_all_sources` runs an 8-worker pool where **each** queued source receives the full `timeout`. When more than `MAX_WAIT_CONCURRENT_SOURCES` (8) sources are still processing, the overall call can run for up to `ceil(N / 8) * timeout` — with the current caps (`N ≤ 100`, `timeout ≤ 3600s`) that is a worst case of ~13 × 3600 ≈ 13h, far longer than a caller passing `timeout=3600` expects.

This is **pre-existing behavior** (the bounded pool was originally in the REST route; #1877 hoisted it into `_app` and made MCP match it — so REST and MCP are now consistent). It is therefore out of scope for the #1877 consolidation, which is intentionally byte-identical for REST.

## Proposal
Give queued items a shared **deadline / remaining budget** instead of the full per-source `timeout`:
- Compute a monotonic deadline once at entry (`deadline = now + timeout`).
- Each worker passes `remaining = max(0, deadline - now)` as the per-source `wait_until_ready` timeout.
- A source that can't start before the deadline returns a `SourceWaitTimeout` outcome immediately.

This bounds the **overall** request duration at ~`timeout` regardless of N, while preserving the typed-outcome contract. It changes both adapters' timeout semantics, so it needs its own PR + parity re-verification (and a note that REST is no longer byte-identical to pre-#1877).

## Acceptance
- `wait_all_sources(timeout=T)` completes in ≈ T wall-clock even when N ≫ concurrency.
- Sources not reached before the deadline surface as timed-out, not dropped.
- REST + MCP behave identically; a unit test pins the overall-duration bound.

teng-lin:
## Summary

GitHub issue #1878 ("source_wait: bound overall request duration, not just per-source timeout") was filed against the OLD bounded-concurrency wait pool, where each queued source received the full `timeout`, so the overall multi-source wait scaled with N (~`ceil(N/workers)*timeout`).

**#1878 was already resolved by #1870 (merged as PR #1882).** That PR rewrote the multi-source path into a single-snapshot loop (`SourcePoller.wait_all_until_ready`) driven by ONE shared `RuntimeDeadline.start(timeout)`:

- Every poll tick fetches the whole notebook source list ONCE and resolves every still-pending source against that single snapshot (no per-source fan-out).
- On `deadline.expired()`, every still-pending source is resolved to its own `SourceTimeoutError` at once via `_fill_timeouts`.
- The per-tick sleep is clamped with `deadline.clamp_sleep`, so accumulated sleeps can never exceed the single `timeout`.

The overall wait is therefore bounded by ~one `timeout` regardless of N. Both typed adapters — the MCP `source_wait` / `source_add_and_wait` tools (via `mcp/tools/_waitagg.py`) and the REST `POST /sources/wait` route — route through `_app.source_wait.wait_all_sources` → `client.sources.wait_all_until_ready`, so they inherit the bound. `MAX_WAIT_CONCURRENT_SOURCES` is gone.

## Why this PR

The existing timeout test (`test_wait_all_until_ready_timeout_reports_each_sources_own_last_status`) only exercised **2 sources in a single tick**, so it did not pin the N-independence the issue is specifically about. This adds a focused, deterministic regression guard:

- Drives `wait_all_until_ready` with **25 sources** that all stay `PROCESSING` forever (`list_sources` always returns the same PROCESSING snapshot), against an injected fake clock whose only time source is the loop's own clamped sleeps.
- Asserts (a) total simulated elapsed time is bounded by ~one `timeout` (NOT `N*timeout`), (b) all 25 come back as `SourceTimeoutError`, (c) each carries its own id and last status.

Test-only; no production change.

Closes #1878

## Summary by CodeRabbit

* **Tests**
  * Added coverage verifying that source readiness polling applies one overall timeout across multiple pending sources.
  * Confirmed polling remains efficient and bounded, regardless of the number of sources.
  * Verified timed-out sources report the correct identifier and processing status.

### DIFF
--- a/tests/unit/test_source_polling_service.py
+++ b/tests/unit/test_source_polling_service.py
@@ -441,6 +441,77 @@ async def test_wait_all_until_ready_timeout_reports_each_sources_own_last_status
 
 
 @pytest.mark.asyncio
+async def test_wait_all_until_ready_bounds_overall_duration_regardless_of_source_count(
+    poller: SourcePoller,
+    logger: logging.Logger,
+) -> None:
+    """Regression for #1878: the OVERALL multi-source wait is bounded by ONE
+    ``timeout``, not ``N * timeout``.
+
+    The issue was filed against the pre-#1870 bounded-concurrency wait pool, where
+    each queued source got the full ``timeout`` so the overall wait grew with N
+    (~``ceil(N / workers) * timeout``). #1870 rewrote this into a single-snapshot
+    loop over one shared :class:`RuntimeDeadline`, so N never multiplies the wait.
+
+    Drive 25 sources that ALL stay PROCESSING forever (``list_sources`` always
+    returns the same PROCESSING snapshot) against an injected fake clock whose only
+    time source is the loop's own clamped sleeps. Assert the total *simulated*
+    elapsed time is bounded by ~one ``timeout`` (NOT 25 * timeout), that all 25 come
+    back as :class:`SourceTimeoutError`, and that each carries its own last status.
+    """
+    source_count = 25
+    timeout = 5.0
+    ids = [f"s{i}" for i in range(source_count)]
+    # Every source is stuck PROCESSING and never leaves that state.
+    stuck = [Source(id=sid, status=SourceStatus.PROCESSING) for sid in ids]
+    # The same PROCESSING snapshot on every tick — nothing ever becomes ready.
+    list_sources = _sequenced_list_sources([stuck])
+
+    sleeps: list[float] = []
+    clock = 0.0
+
+    def monotonic() -> float:
+        return clock
+
+    async def sleep(seconds: float) -> None:
+        nonlocal clock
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude @codex please review — test-only. #1878 (bound overall wait duration) was already resolved by #1870 (PR #1882): the multi-source wait now shares one `RuntimeDeadline` across all sources (no more 8-worker pool / per-source-timeout multiplication). Existing tests only exercised 2 sources; this adds a regression guard proving the bound is N-independent (25 stuck sources return in ≤ one `timeout`, all as SourceTimeoutError with their own last_status). `Closes #1878`.

PR #1902, opened 2026-07-15
chore: prepare v0.8.0b2

teng-lin:
Second beta of the 0.8.0 cycle. Prep only — tag/publish follows separately with the confirm-gates.

## Version
`0.8.0b1 → 0.8.0b2` (pyproject + `desktop-extension/manifest.json` + `uv.lock`; `uv lock --check` clean).

## What changed since b1
The **only** delta is [#1898](https://github.com/teng-lin/notebooklm-py/issues/1898) — a blank `FASTMCP_STATELESS_HTTP` no longer crash-loops the remote MCP server (compose passthrough removed + an import-time env guard). Added to `[0.8.0]` Fixed.

## E2E scope (deliberate)
b2 touches **no RPC/API code** (`rpc/` and every live call path are identical to b1, which already passed full E2E + RPC health hands-on). So the standalone step-8 E2E/RPC-health dispatches are **skipped as redundant** — but **Verify Package still runs unit+integration+E2E against the published wheel** (not skippable), so the packaged artifact is E2E-verified regardless.

## Gates (local)
- ✅ Public-API compat audit (0 unapproved breaks) · ruff (945 files) · mypy (312) · desktop-extension version-match.

https://claude.ai/code/session_01RTdQA3fnD98jve4fB6Aocf

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented remote MCP servers from repeatedly crashing when `FASTMCP_STATELESS_HTTP` is provided as an empty or whitespace-only value.
  * Empty configuration values are now handled as unset for more reliable server startup.

* **Chores**
  * Updated the application and desktop extension version to 0.8.0b2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -473,6 +473,14 @@ get-returns-None / kwarg-alias deprecation machinery — has been **removed**
 
 ### Fixed
 
+- **A blank `FASTMCP_STATELESS_HTTP` no longer crash-loops the remote MCP server.**
+  The deploy compose passed the variable through as `${FASTMCP_STATELESS_HTTP:-}`,
+  which injects an **empty string** when unset — and FastMCP's settings bool-parse
+  it at import and raise, so a docker deploy that didn't set the var crash-looped.
+  The compose passthrough is removed (the upload widget auto-enables stateless in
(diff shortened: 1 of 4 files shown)


teng-lin:
@claude review

Issue #1938, opened 2026-07-16
Tracking: deferred MCP upload-widget + file-transfer follow-ups (#1889/#1891/#1894/#1896)


teng-lin:
## Summary
Tracking issue for follow-up work explicitly deferred (as "out of scope" / "follow-ups, not here" / "deferred") across the in-app upload-widget + mobile-file-upload PRs merged this week, none of which currently has a tracking issue. Grouped so they don't get lost now that the parent issues are closed.

### From #1894 (multi-file widget upload)
- [ ] **Serial multi-file uploads can 403 late files when tokens expire mid-batch.** All per-file tokens are minted up front with one short TTL; on a slow link a later large file may start after its token expired. Author accepted as a known v1 limitation. Fix: refresh/mint tokens per-file, or parallelize. (chatgpt-codex, `_uploadwidget.py:146`)

### From #1891 (in-app upload widget, Phase 3)
- [ ] Progress bar in the widget
- [ ] Fallback ladder: `window.openai.uploadFile` → direct-PUT → signed link
- [ ] Auto-wire `await_upload` so the model confirms the add without a second prompt

### From #1889 (await_upload + short-link)
- [ ] Phase 4: `sha256` completion record (client-hashing) for the completion map
- [ ] `ctx.report_progress` keepalive during long uploads
- [ ] Phase 2b probe / Phase 3 direct-PUT widget path
  - (Phase 5 SEP-2631 transport migration is already tracked by #1656 — excluded here.)

### From #1896 (fold source-add composites into source_add)
- [ ] Remaining ADR-0025 optional folds punted to "separate PRs": `studio_get_prompt` → `studio_list` fold, and `source_add_drive_file` docstring sharpen (parent #1890 closed with these open)

## Notes
All items are minor/enhancement-tier and were deliberate deferrals, not misses — this issue just gives them a home. Prioritize as convenient; several (progress bar, fallback ladder, token-per-file) improve the mobile widget UX together.

_Compiled from a review/PR-body sweep of PRs merged this week._

PR #1942, opened 2026-07-16
fix(mcp): give the upload widget's token pool a longer TTL (#1894)

teng-lin:
### Problem
`source_add_widget` mints a **pool** of single-use upload tokens up front (one per file, cap 10) but the widget uploads them **sequentially** through one browser. Every token shared the 15-min single-link `UPLOAD_TTL`, all stamped at the same mint instant. On a slow mobile link a *later* file can start uploading after its token has already expired → the `/files/ul` POST returns an opaque **HTTP 403** and the file is silently lost. The widget author accepted this as a known v1 limitation.

### Fix
Add a dedicated `WIDGET_UPLOAD_TTL` (1 h) and mint the widget pool with it, so the last token outlives the sum of every earlier file's transfer plus the user's file-picking think-time. `FileTransferConfig.upload_url()` gains an optional `ttl` kwarg that **defaults to `UPLOAD_TTL`** — the single-link broker (`_broker_upload`) is byte-for-byte unchanged.

Chosen over the two alternatives because they don't fit the architecture: *lazy per-file minting* / *client-requests-fresh-token* both require an on-demand mint endpoint the static widget would call **without an MCP credential** — an unauthenticated upload-token minter for any notebook, a security regression. A longer TTL on the already-minted pool keeps the signed, MCP-authenticated mint as the sole authorization point.

### Why the longer window is safe
It's a smaller risk class than the *human* link the tight `UPLOAD_TTL` guards. A pool token is:
- **single-use** — burned via `ConsumedJtiStore` on the one successful add, *regardless of TTL* (this PR does not touch the replay guard);
- **notebook-scoped** (the signed `nb` claim);
- **never surfaced** to the user or placed in a URL bar / history / `Referer` — it lives only inside the widget's `structuredContent` / `fetch` body.

The 15-min value exists to bound the *human_upload* link (opened in a browser: URL bar, history, referer — a genuinely higher-exposure surface). The widget pool never enters that surface.

### Preserves
The single-use / replay guarantee (`ConsumedJtiStore`), the HMAC signed-token model, `op="ul"` route-matching, and per-token `jti` injection — all unchanged. Only the token lifetime for the widget pool changes.

### Tests
- `test_filelink.py`: `upload_url(ttl=…)` honors a custom TTL and still stamps `op`/`jti`; default stays `UPLOAD_TTL`.
- `test_uploadwidget.py`: every pool token carries `WIDGET_UPLOAD_TTL` (> `UPLOAD_TTL`) and remains a valid single-use `ul` token.

`mypy`, `ruff check`/`format`, and the MCP unit suite + `tests/_guardrails` (module-size + schema-char budgets) all green.

Addresses part of #1938 (other items open).

https://claude.ai/code/session_01XmN63FN4Kz2nkmC4eX99jC

## Summary by CodeRabbit

* **Bug Fixes**
  * Extended in-app upload widget token validity for multi-file, sequential batches to prevent late-file failures.
  * Kept the shorter expiration for standard single-file upload links unchanged.
* **Documentation**
  * Clarified upload-token expiration behavior for widget pool uploads vs single upload links.
  * Updated architecture and the experimental upload-widget guide text for clarity.
* **Tests**
  * Added unit tests validating both default and widget-specific TTL behavior for minted upload tokens.


teng-lin:
@claude review

teng-lin:
@codex review

teng-lin:
Addressed in 833fb3b9: updated the two stale `≤ UPLOAD_TTL` comments in `ConsumedJtiStore` (`_seen` cap + `_results` map) to say the sweep bound is the token's own TTL (`UPLOAD_TTL`, or `WIDGET_UPLOAD_TTL` for a widget-pool token) — `_sweep` already reads the real `exp`, so this is doc-accuracy only. Also surfaced the pool's 1 h lifetime in the `architecture.md` per-file index and the mcp-guide widget section. Thanks for the careful read.

PR #1946, opened 2026-07-16
feat(mcp): emit report_progress keepalives during await_upload's poll (#1889)

teng-lin:
### What & why
`await_upload` polls the in-process completion map for up to ~45 s (bounded just under claude.ai's ~60 s connector watchdog). On a slow upload, an MCP host watching for a stalled call could consider that single poll window dead. This wires `ctx.report_progress` into the `await_upload` tool so it emits a progress notification at **t=0 and each ~2 s tick**, giving the host visible liveness across the wait.

The re-invoke loop stays the **load-bearing** completion path — the keepalive is purely a "still alive" signal, so `progress` climbs a plain tick counter with a status message rather than pretending to be an upload %-done (the tool polls; it does not transfer the bytes).

### Scope note (coordinated with the lead)
The brief listed `_fileroutes.py` for this item, but `/files/ul` is a raw Starlette HTTP route (the browser→server byte transfer) with **no MCP `ctx`** — there is no progress channel there. The MCP call a host can consider stalled is the `await_upload` **tool**, which already had an unused best-effort `progress` hook in `_await_upload` (`tools/_fileupload.py`) and a ponytail TODO for exactly this. So the change lands there. It does **not** overlap `fix/consolidate-source-add-1890` (whose `_fileupload.py` edits are top-of-file docstring renames, far from `await_upload`).

### Best-effort by contract
- `report_progress` no-ops when the client sent no `progressToken`.
- `_await_upload` now **centrally swallows** any progress-callback error (`_emit_progress`), enforcing the "the design does not depend on it" contract that was previously only documented — a notify failure can never abort the wait.

### Tool surface unchanged
`await_upload(ctx, upload_link, timeout=45.0)` signature/docstring are untouched → no ADR-0025 schema-char / tool-count impact (verified: `test_tool_eval`, `test_manifest` green). `_fileupload.py` at 432/1000 lines.

### Tests (`test_await_upload.py`)
- keepalive fires at t=0 **and** each poll tick (≥2 calls over a short poll);
- a **raising** keepalive is swallowed → the poll still returns `received`;
- end-to-end: the `await_upload` **tool** (built via `create_server`) awaits `ctx.report_progress` with a status message.

`mypy`, `ruff`, MCP unit suite, and the touched guardrails all green.

Addresses part of #1938 (other items open).

https://claude.ai/code/session_01XmN63FN4Kz2nkmC4eX99jC

## Summary by CodeRabbit

* **New Features**
  * Added optional MCP progress notifications during file upload waiting, including an initial update and periodic keepalive-style updates.
  * Progress updates now include human-readable status messages and are delivered as real protocol notifications.
* **Bug Fixes**
  * Progress/keepalive reporting failures are now handled best-effort, so errors in progress callbacks no longer interrupt or fail the upload-wait flow.
* **Tests**
  * Expanded unit and end-to-end MCP coverage for keepalive/progress behavior and error handling.

### DIFF
--- a/src/notebooklm/mcp/tools/_fileupload.py
+++ b/src/notebooklm/mcp/tools/_fileupload.py
@@ -419,9 +419,8 @@ def register_file_tools(mcp: Any) -> None:
             # watching for a stalled call sees liveness across the ~45s wait — the re-invoke
             # loop stays the load-bearing completion path. ``progress`` climbs a plain tick
             # counter (an honest liveness ping, not an upload %-done — this polls, it doesn't
-            # transfer the bytes). ``report_progress`` no-ops when the client sent no
-            # progressToken, and ``_await_upload`` swallows any error, so this never fails the
-            # tool.
+            # transfer the bytes). It's best-effort by ``_await_upload``'s contract (swallowed
+            # on error; a no-op when the client sent no progressToken).
             ticks = 0
 
             async def _keepalive() -> None:
--- a/src/notebooklm/mcp/tools/_fileupload.py
+++ b/src/notebooklm/mcp/tools/_fileupload.py
@@ -325,9 +325,20 @@ async def _await_upload(
     - ``{"status": "expired_or_invalid", "hint": ...}`` — the link failed signature/
       expiry/op checks; mint a fresh one via ``source_add(source_type="file")``.
 
-    ``progress`` (best-effort) is awaited at t=0 and each tick as a keepalive; the design
-    does not depend on it.
+    ``progress`` (best-effort) is awaited at t=0 and each poll tick as a keepalive; the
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review

teng-lin:
@codex review

teng-lin:
Addressed both notes in e23cce9c, thanks @claude:
1. **Test style / true end-to-end** — replaced the hand-built `SimpleNamespace` ctx test with an in-memory `fastmcp.Client` + `progress_handler`, so it now asserts the keepalive actually reaches a client as a real progress notification over a progressToken-bearing request (not just that `ctx.report_progress` was called).
2. **Duplicated rationale** — trimmed the tool's inline comment to defer to `_await_upload`'s docstring as the single source of truth for the best-effort/swallow contract.

PR #1974, opened 2026-07-17
fix(chat): classify empty conversations as fresh

### LINKED ISSUE #1973: chat_ask reports is_follow_up=true on the FIRST question to a notebook (over-correction of #1965)
Found live-testing **v0.8.0b4** (the release that shipped the #1965 fix). This is a regression/over-correction introduced by that fix.

## Problem
`chat_ask` (no `conversation_id`) reports `is_follow_up: true` for the **first-ever** question to a notebook — one with no prior Q&A.

## Live repro (b4, remote connector)
Created a brand-new notebook, added one text source, asked its **first** question:
```
chat_ask(notebook=<fresh>, question="What is the capital of Testlandia?")
→ {"conversation_id":"3a12ce8f-…"(new), "turn_number":1, "is_follow_up":true, ...}
```
`is_follow_up:true` on a first question is wrong — there is nothing to follow up on. Reproduced on two independently-fresh notebooks; every observed `chat_ask` on the remote connector now returns `is_follow_up:true`.

## History
- b3 (the original #1965 bug): null-ask `is_follow_up` was **always false** (never detected continuations).
- b4 (#1965 fix): the null-ask path now sets `is_follow_up = (current_id is not None)` — so it's correct for genuine continuations but **always true** when a current conversation object exists.

## Root cause
`_chat/api.py` null-ask path (~L422-445):
```python
current_id = await self.get_conversation_id(notebook_id)
...
if current_id is not None:
    ...
    is_follow_up = override is not None   # true whenever a current conversation exists
```
The logic assumes `get_conversation_id(notebook) is None` ⟺ "first-ever question". But **live, `get_conversation_id` returns a non-None id for a never-chatted notebook** — the backend appears to auto-create/track a current conversation (empty) at notebook-creation / first-source time. So the first real question sees `current_id != None` → `is_follow_up=true`. The fix conflates *"a current conversation object exists"* with *"there is prior Q&A to follow up on"*.

(The b4 unit tests assert first-ever→false, but they mock `get_conversation_id → None`, which doesn't match live backend behavior.)

## Impact
`is_follow_up` is now effectively **always true on the remote connector**, so it's unreliable as a "has prior context" signal — an agent seeing `is_follow_up:true` on a user's opening question may wrongly assume conversation history exists. (Also note `turn_number` is `1` even when `is_follow_up:true`, an internal inconsistency — turn_number reflects only the stateless client's locally-cached turns, which is always ≤1 per request.)

## Fix direction
`is_follow_up` should reflect whether the resumed conversation actually has **prior turns**, not merely whether a current-conversation id exists. Options:
- Treat an **empty** current conversation (zero prior Q&A) as new → `is_follow_up=false`. Requires knowing the turn count of `current_id`; if the "current conversation" lookup can cheaply report emptiness, use it.
- If turn-count can't be known without an extra fetch, reconsider whether `is_follow_up` can be computed reliably at all on the stateless remote path (and document the limitation) rather than defaulting to true.
- Add a live-backed regression test (a genuinely-fresh notebook's first ask → `is_follow_up=false`); the current unit test's `get_conversation_id → None` mock hid this.

_Found via a live MCP connector stress test of v0.8.0b4._

loulanyue:
## Summary

Treat an auto-created current conversation with zero server-side turns as a fresh conversation, so the first `chat_ask` reports `is_follow_up=false`.

## Related Issue

Closes #1973

## Changes

- reuse locally cached turns when available to avoid an extra metadata request
- otherwise query at most one server-side turn before classifying an implicit continuation
- keep the history probe best-effort so metadata failures do not block the chat request
- add a regression test for a non-null current conversation ID with an empty turn list

## Test Plan

- [x] Full test and coverage suite passes (`uv run pytest -n auto --dist loadgroup --cov=src/notebooklm --cov-report=term-missing --cov-report=json:coverage.json --cov-fail-under=90`): 12,387 passed, 64 skipped, 1 xfailed; 96.45% coverage
- [x] No architectural shape change; an ADR is not required

## Notes

The additional `khqZz` lookup only runs when the current conversation has no locally cached turns. If that best-effort lookup fails, the existing-ID behavior is retained and the ask proceeds.

This contribution was developed with AI assistance. I reviewed the resulting diff and ran the test, coverage, lint, formatting, and type-check commands listed above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected follow-up detection for chat requests by deriving follow-up status from detected prior turn history for the server-assigned current conversation.
  * Prevented fresh conversations from being incorrectly marked as follow-ups when no prior turns exist.
  * Ensured history-probe failures raise an error and stop subsequent chat submission.

* **Tests**
  * Added unit regression coverage for empty vs. non-empty history and probe-failure behavior.
  * Extended test fixtures/mocks and updated VCR-based integration tests to support the added follow-up probe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Fresh conversations are no longer reported as follow-ups.** A null-conversation
+  ask now checks whether the server-assigned current conversation has any turns
+  before setting `AskResult.is_follow_up` ([#1973](https://github.com/teng-lin/notebooklm-py/issues/1973)).
+
 ## [0.8.0]
 
 The headline of 0.8.0 is **integrations**: NotebookLM is now reachable from AI
--- a/src/notebooklm/_chat/api.py
+++ b/src/notebooklm/_chat/api.py
@@ -439,10 +439,33 @@ class ChatAPI(LoopBoundPrimitive):
                     # conversation for the null POST, so drop the override and
                     # recover the real id post-POST, not the deleted one (#1875).
                     override = None if current_id in self._deleted_conversations else current_id
-                    # Resuming a live current conversation is a genuine
-                    # follow-up; a deleted one makes the server start fresh, so
-                    # it is not (override is None ⇒ new conversation) (#1965).
-                    is_follow_up = override is not None
+                    # A current id can refer to an auto-created, zero-turn
+                    # conversation. Only classify it as a follow-up when it has
+                    # local or server-side history (#1973).
+                    is_follow_up = False
+                    if override is not None:
+                        cached_turns = self._cache.get_cached_conversation(override)
+                        if cached_turns:
+                            is_follow_up = True
+                        else:
+                            try:
+                                turns_data = await self.get_conversation_turns(
+                                    notebook_id, override, limit=1
+                                )
+                                is_follow_up = bool(
+                                    unwrap_conversation_turns(
+                                        turns_data, source="_chat.ask.follow_up_probe"
+                                    )
+                                )
+                            except Exception as e:  # noqa: BLE001 - metadata probe is best-effort
+                                # Preserve ask availability when the metadata probe
+                                # fails; an existing id remains the safest fallback.
+                                logger.warning(
+                                    "Failed to inspect current conversation %s before ask: %s",
+                                    override,
+                                    e,
+                                )
+                                is_follow_up = True
                     posted = await perform_request(
                         conversation_history=None,
                         active_conversation_id=None,
--- a/src/notebooklm/_chat/api.py
+++ b/src/notebooklm/_chat/api.py
@@ -440,22 +440,18 @@ class ChatAPI(LoopBoundPrimitive):
                     # recover the real id post-POST, not the deleted one (#1875).
                     override = None if current_id in self._deleted_conversations else current_id
                     # A current id can refer to an auto-created, zero-turn
-                    # conversation. Only classify it as a follow-up when it has
-                    # local or server-side history (#1973).
+                    # conversation. Query the server even when local turns are
+                    # cached because another client may have deleted them (#1973).
                     is_follow_up = False
                     if override is not None:
-                        cached_turns = self._cache.get_cached_conversation(override)
-                        if cached_turns:
-                            is_follow_up = True
-                        else:
-                            turns_data = await self.get_conversation_turns(
-                                notebook_id, override, limit=1
+                        turns_data = await self.get_conversation_turns(
+                            notebook_id, override, limit=1
+                        )
+                        is_follow_up = bool(
+                            unwrap_conversation_turns(
+                                turns_data, source="_chat.ask.follow_up_probe"
                             )
-                            is_follow_up = bool(
-                                unwrap_conversation_turns(
-                                    turns_data, source="_chat.ask.follow_up_probe"
-                                )
-                            )
+                        )
                     posted = await perform_request(
                         conversation_history=None,
                         active_conversation_id=None,
--- a/src/notebooklm/_chat/api.py
+++ b/src/notebooklm/_chat/api.py
@@ -448,24 +448,14 @@ class ChatAPI(LoopBoundPrimitive):
                         if cached_turns:
                             is_follow_up = True
                         else:
-                            try:
-                                turns_data = await self.get_conversation_turns(
-                                    notebook_id, override, limit=1
+                            turns_data = await self.get_conversation_turns(
+                                notebook_id, override, limit=1
+                            )
+                            is_follow_up = bool(
+                                unwrap_conversation_turns(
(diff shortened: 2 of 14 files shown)


teng-lin (commented):
Thanks for this — and for jumping on #1973 so quickly. 👏 This is a strong fix: it independently arrives at the same shape a careful review of this bug converges on — a warm local-cache fast path, then a **pre-POST raw `get_conversation_turns(limit=1)` + `unwrap_conversation_turns` existence probe** (correctly *not* `get_history`), placed **inside `_get_conversation_lock(current_id)`, after the `_deleted_conversations` recheck, skipped when `override is None`**. A multi-model review panel (Claude + Codex + Gemini) all independently validated the placement/concurrency, the existence check, and the cache fast-path as correct. Nice work.

There's **one blocker** plus some test-coverage gaps before it's ready to merge.

## 🔴 Critical — the probe-failure fallback reintroduces #1973

`src/notebooklm/_chat/api.py` (~L460–468):
```python
except Exception as e:  # noqa: BLE001 - metadata probe is best-effort
    ...
    is_follow_up = True
```
Defaulting to `True` on probe failure recreates the exact bug this PR fixes. The probe only runs on a **cold cache** — which is **every CLI `notebooklm ask` and every first MCP ask** (the cache is per-process). So the failure window (`cold cache + first question to an auto-created zero-turn conversation`) is *precisely* the #1973 scenario: a transient khqZz blip / auth hiccup / server 5xx → the genuinely-first question reports `is_follow_up=True`.

**Recommended fix: remove the `try/except` entirely and let the probe error propagate.** Because the probe is **pre-POST**, propagation fails the ask *before* any answer is generated — nothing paid-for is lost (unlike the post-POST hPTbtc recovery path). If you'd rather prioritize availability, default to **`False`**, never `True`: we only reach the probe when the local cache holds zero evidence of history, so "not a follow-up" is the evidence-consistent default.

Separately, `except Exception` is **too broad** — it swallows `UnknownRPCMethodError` (genuine schema drift) that the codebase deliberately surfaces (compare `get_history`, which catches only `(ChatError, NetworkError)` and lets drift propagate). Removing the catch resolves this too.

## 🟠 Important — pin the behavior with tests

1. **No test covers the disputed fallback.** Whatever you decide in the Critical, pin it: make khqZz raise (e.g. a 500 / malformed payload) and assert the chosen behavior — the ask raises **before** any `GenerateFreeFormStreamed` POST (assert the library-mapped `ServerError`, not raw `httpx`), or `is_follow_up=False`.
2. **The cold-cache *positive* branch is untested.** There's a cold-cache-0-turns→`False` case, and warm-cache→`True` is covered by the existing `test_ask_implicit_continue_is_follow_up` (which hits the fast path, no probe). But **cold cache + conversation-with-turns → `True` via the probe** has no assertion. Add one: empty cache, khqZz returns ≥1 turn, assert `is_follow_up is True` **and** that khqZz was actually requested.
3. **`mock_get_conversation_id`'s khqZz response is `is_optional=True, is_reusable=True`** (`tests/conftest.py:427`), so tests can pass whether the probe ran, was skipped, or was swallowed — "green" doesn't prove the probe fired. Your new test guarding with `assert any("rpcids=khqZz" ...)` is exactly the right pattern; consider asserting probe-fired / probe-not-fired explicitly in the cold vs warm cases so the shared optional fixture can't mask a regression.

## 🟡 Minor

4. **CHANGELOG** — this is a user-visible fix to `AskResult.is_follow_up`; please add a `### Fixed` entry under `[Unreleased]` referencing #1973.
5. **`turn_number` / `is_follow_up` can disagree on a cold cache** — once the probe correctly returns `True`, `cache_turn` (derived from the empty per-request cache) still reports `turn_number=1`. This is **pre-existing** (not introduced here) and we're tracking it as a separate issue, so no action needed in this PR — just flagging so it's not a surprise.

Once the fallback direction is fixed (propagate, or `False`) and the behavior is pinned by tests, this is good to go. Really solid contribution — thank you! 🙏

teng-lin (approved):
🎉 This fully addresses the review — and fast. Re-reviewed the revision and verified it locally (131 chat/conversation tests pass, mypy + ruff clean).

**All findings resolved:**
- ✅ **Critical — probe-failure now propagates.** No more `except … is_follow_up = True`; the probe (`get_conversation_turns` + `unwrap_conversation_turns`) errors surface, failing the ask *pre-POST* (no answer paid for), and the default is `is_follow_up = False`. `except Exception` is gone, so genuine schema drift (`UnknownRPCMethodError`) propagates too. `test_conversation_probe_failure_raises_before_chat_post` pins it (asserts `ServerError` before any chat POST). 👍
- ✅ **Cross-client cache — fast path dropped (option 1).** "Query the server even when local turns are cached because another client may have deleted them (#1973)" — exactly the right call. Always server-authoritative, no Web-UI-delete staleness window, one fewer branch to test.
- ✅ Placement/concurrency preserved (probe inside `_get_conversation_lock`, after the deleted-id recheck, skipped when `override is None` → deleted stays `False`, #1875 intact); raw `limit=1` existence check; CHANGELOG entry added.

`turn_number` is correctly left untouched (tracked separately in #1976).

LGTM — ready to merge once CI is green. Thank you for the thorough, quick turnaround! 🙏

PR #2005, opened 2026-07-22
fix(cli): clarify Windows login session failures

### LINKED ISSUE #2004: notebooklm login fails with spawn UNKNOWN when launching Chromium in a sandboxed/non-interactive Windows session
Description
notebooklm login fails with BrowserType.launch_persistent_context: spawn UNKNOWN when launching the bundled Chromium in a Windows automation/sandbox environment that has no interactive desktop/window station session (e.g. a CI-like agent shell). The --browser-cookies chrome fallback also fails, with "Could not decrypt chrome cookies" (whose hint text only covers macOS Keychain, not Windows DPAPI).

Steps to Reproduce
uv tool install --python 3.12 "notebooklm-py[browser,cookies]"
Run notebooklm login
On first run it correctly reports Chromium isn't downloaded yet; ran python -m playwright install chromium in the tool's venv to fix that.
Run notebooklm login again (or notebooklm -vv login) — this time it launches but fails with spawn UNKNOWN (full traceback below). Reproduced identically from both Git Bash (MSYS) and native Windows PowerShell 5.1.
As a fallback, tried notebooklm login --browser-cookies chrome — fails with a cookie-decryption error instead (see below).
Expected Behavior
notebooklm login should launch Chromium and complete the Google sign-in flow, or --browser-cookies chrome should successfully decrypt and read cookies from the local Chrome profile on Windows.

Actual Behavior
notebooklm -vv login:
Profile: default
Opening Chromium for Google login...
Using persistent profile:
C:\Users\<user>\.notebooklm\profiles\default\browser_profile
00:02:46 DEBUG [notebooklm.cli.services.playwright_login] Login failed: BrowserType.launch_persistent_context: spawn UNKNOWN
Call log:
  - <launching> C:\Users\<user>\AppData\Local\ms-playwright\chromium-1228\chrome-win64\chrome.exe --disable-field-trial-config --disable-background-networking --disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-back-forward-cache --disable-breakpad --disable-client-side-phishing-detection --disable-component-extensions-with-background-pages --disable-component-update --no-default-browser-check --disable-default-apps --disable-dev-shm-usage --disable-edgeupdater --disable-extensions --disable-features=AvoidUnnecessaryBeforeUnloadCheckSync,BoundaryEventDispatchTracksNodeRemoval,DestroyProfileOnBrowserClose,DialMediaRouteProvider,GlobalMediaControls,HttpsUpgrades,LensOverlay,MediaRouter,PaintHolding,ThirdPartyStoragePartitioning,Translate,AutoDeElevate,RenderDocument,OptimizationHints,msForceBrowserSignIn,msEdgeUpdateLaunchServicesPreferredVersion --enable-features=CDPScreenshotNewSurface --allow-pre-commit-input --disable-hang-monitor --disable-ipc-flooding-protection --disable-popup-blocking --disable-prompt-on-repost --disable-renderer-backgrounding --force-color-profile=srgb --metrics-recording-only --no-first-run --password-store=basic --use-mock-keychain --no-service-autorun --export-tagged-pdf --disable-search-engine-choice-screen --unsafely-disable-devtools-self-xss-warnings --edge-skip-compat-layer-relaunch --disable-infobars --disable-search-engine-choice-screen --disable-sync --enable-unsafe-swiftshader --no-sandbox --disable-blink-features=AutomationControlled --password-store=basic --user-data-dir=C:\Users\<user>\.notebooklm\profiles\default\browser_profile --remote-debugging-pipe about:blank

Traceback (most recent call last):
  File "...\notebooklm-py\Lib\site-packages\notebooklm\cli\services\playwright_login.py", line 791, in run_playwright_login
    context = p.chromium.launch_persistent_context(**launch_kwargs)
  File "...\playwright\sync_api\_generated.py", line 16816, in launch_persistent_context
    self._sync(
  File "...\playwright\_impl\_sync_base.py", line 115, in _sync
    return task.result()
  File "...\playwright\_impl\_browser_type.py", line 165, in launch_persistent_context
    result = await self._channel.send_return_as_dict(
  File "...\playwright\_impl\_connection.py", line 83, in send_return_as_dict
    return await self._connection.wrap_api_call(
  File "...\playwright\_impl\_connection.py", line 563, in wrap_api_call
    raise rewrite_error(error, f"{parsed_st['apiName']}: {error}") from None
playwright._impl._errors.Error: BrowserType.launch_persistent_context: spawn UNKNOWN

Unexpected error: BrowserType.launch_persistent_context: spawn UNKNOWN
This may be a bug. Please report at https://github.com/teng-lin/notebooklm-py/issues

notebooklm login --browser-cookies chrome:
Note: sibling-product cookies not included by default. Pass
--include-domains=<docs, mail, myaccount, youtube> (or =all) to extract them.
Reading cookies from chrome...
Could not decrypt chrome cookies.
On macOS, allow Keychain access when prompted, or try a different browser.

Environment
OS: Windows 11 Home Single Language, build 10.0.26200
Python version: 3.12.13 (pinned via uv tool install --python 3.12; originally tried on 3.13.14, where rookiepy==0.5.6 has no prebuilt Windows wheel and its sdist fails to build under uv due to a pyproject.toml missing project.version)
notebooklm-py version: 0.7.3 (notebooklm --version → NotebookLM CLI, version 0.7.3)
Install method: uv (uv tool install)
Surface: CLI
Playwright: 1.61.0, Chromium build 1228 (chrome-win64), installed via python -m playwright install chromium in the tool's own venv
Shell: reproduced identically in both Git Bash (MSYS) and native Windows PowerShell 5.1
Note: this environment is a sandboxed automation shell (Claude Code) without an interactive Windows desktop/window station session, which is the suspected root cause for both failures below (Chromium can't spawn a GUI process, and Chrome's DPAPI-based cookie store isn't accessible)

Debug Output
notebooklm doctor --json

{
  "profile": "default",
  "profile_source": "default",
  "checks": {
    "migration": { "status": "pass", "detail": "complete" },
    "profile_dir": { "status": "warn", "detail": "C:\\Users\\<user>\\.notebooklm\\profiles\\default (permissions: 0o777, expected: 0o700)" },
    "auth": { "status": "fail", "detail": "not authenticated" },
    "config": { "status": "pass", "detail": "not present (using defaults)" }
  }
}
notebooklm status --paths --json

{
  "paths": {
    "home_dir": "C:\\Users\\<user>\\.notebooklm",
    "home_source": "default (~/.notebooklm)",
    "profile": "default",
    "profile_source": "default",
    "profile_dir": "C:\\Users\\<user>\\.notebooklm\\profiles\\default",
    "storage_path": "C:\\Users\\<user>\\.notebooklm\\profiles\\default\\storage_state.json",
    "context_path": "C:\\Users\\<user>\\.notebooklm\\profiles\\default\\context.json",
    "config_path": "C:\\Users\\<user>\\.notebooklm\\config.json",
    "browser_profile_dir": "C:\\Users\\<user>\\.notebooklm\\profiles\\default\\browser_profile"
  }
}

notebooklm auth check --test --json

{
  "status": "error",
  "checks": {
    "storage_exists": false,
    "json_valid": false,
    "cookies_present": false,
    "sid_cookie": false,
    "token_fetch": null
  },
  "details": {
    "storage_path": "C:\\Users\\<user>\\.notebooklm\\profiles\\default\\storage_state.json",
    "auth_source": "file (C:\\Users\\<user>\\.notebooklm\\profiles\\default\\storage_state.json)",
    "cookies_found": [],
    "cookie_domains": [],
    "error": "Storage file not found: C:\\Users\\<user>\\.notebooklm\\profiles\\default\\storage_state.json"
  }
}
Suggested improvement
Even if there's no direct fix for running inside a non-interactive Windows session, it would help to:

Give spawn UNKNOWN a friendlier hint (similar to the existing "Playwright was just installed" box) pointing at a possible missing interactive desktop/window station.
Extend the "Could not decrypt chrome cookies" hint to cover Windows DPAPI (currently only mentions macOS Keychain).

Checklist
I verified this bug exists on the latest version of notebooklm-py (Done)
I searched existing issues and this is not a duplicate (Done_

Sanjays2402:
## Summary

Show actionable Windows guidance when Playwright cannot launch Chromium from a non-interactive session, and explain the DPAPI requirement when browser-cookie decryption fails.

## Related Issue

Closes #2004

## Changes

- Map Windows `spawn UNKNOWN` launch failures to an interactive-session hint.
- Add Windows DPAPI context to cookie-decryption errors.
- Cover both diagnostics with unit regressions.

## Test Plan

- [ ] Tests pass (`uv run pytest --cov=src/notebooklm --cov-report=term-missing --cov-report=json:coverage.json --cov-fail-under=90`)
- [ ] Linting and formatting pass (`uv run pre-commit run --all-files`)

## Notes

The 68 focused login tests pass; changed-file Ruff, formatting, Mypy, and compilation checks also pass. This does not make GUI login possible without an interactive Windows desktop—the new diagnostic states that constraint explicitly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cookie decryption error messages with clearer macOS Keychain and Windows App-Bound Encryption guidance.
  * Added a recommended Firefox cookie-source command for affected Chrome and Edge configurations.

* **Tests**
  * Updated validation to cover the new Windows guidance and Firefox recommendation.

### DIFF
--- a/src/notebooklm/cli/services/login/rookiepy_errors.py
+++ b/src/notebooklm/cli/services/login/rookiepy_errors.py
@@ -34,6 +34,8 @@ def _handle_rookiepy_error(e: Exception, browser_name: str) -> str:
     if "keychain" in msg or "decrypt" in msg:
         return (
             f"[red]Could not decrypt {browser_name} cookies.[/red]\n"
-            "On macOS, allow Keychain access when prompted, or try a different browser."
+            "On macOS, allow Keychain access when prompted. On Windows, DPAPI cookie "
+            "decryption requires the same interactive user session that owns the browser "
+            "profile. Otherwise, try a different browser."
         )
     return f"[red]Failed to read cookies from {browser_name}:[/red] {e}"
--- a/src/notebooklm/cli/services/login/rookiepy_errors.py
+++ b/src/notebooklm/cli/services/login/rookiepy_errors.py
@@ -34,8 +34,10 @@ def _handle_rookiepy_error(e: Exception, browser_name: str) -> str:
     if "keychain" in msg or "decrypt" in msg:
         return (
             f"[red]Could not decrypt {browser_name} cookies.[/red]\n"
-            "On macOS, allow Keychain access when prompted. On Windows, DPAPI cookie "
-            "decryption requires the same interactive user session that owns the browser "
-            "profile. Otherwise, try a different browser."
+            "On macOS, allow Keychain access when prompted. On Windows, Chrome 127+ and "
+            "current Edge protect the cookie database with App-Bound Encryption (ABE): the "
+            "key is bound to the browser process via a Windows service, so no external "
+            "process can read it and no flag bypasses it. Use Firefox as the cookie source "
+            "instead (plain SQLite, no ABE): notebooklm login --browser-cookies firefox"
         )
     return f"[red]Failed to read cookies from {browser_name}:[/red] {e}"
--- a/src/notebooklm/cli/services/playwright_login.py
+++ b/src/notebooklm/cli/services/playwright_login.py
@@ -458,17 +458,29 @@ def run_playwright_login(plan: PlaywrightLoginPlan, io: LoginIO) -> None:
     io.emit(f"[yellow]Opening {browser_label} for Google login...[/yellow]")
     io.emit(f"[dim]Using persistent profile: {plan.browser_profile}[/dim]")
 
-    result: CaptureResult = run_browser_capture(
-        BrowserCapturePlan(
-            browser=plan.browser,
-            browser_profile=plan.browser_profile,
-            storage_path=plan.storage_path,
-            include_domains=plan.include_domains,
-        ),
-        io,
-        headless=False,
-        interactive=True,
-    )
+    try:
+        result: CaptureResult = run_browser_capture(
+            BrowserCapturePlan(
+                browser=plan.browser,
+                browser_profile=plan.browser_profile,
+                storage_path=plan.storage_path,
+                include_domains=plan.include_domains,
+            ),
+            io,
+            headless=False,
+            interactive=True,
+        )
+    except Exception as exc:
+        if sys.platform == "win32" and "spawn unknown" in str(exc).lower():
+            io.emit(
+                "[red]Chromium could not start in this Windows session.[/red]\n"
+                "Playwright needs an interactive desktop or window station to open the "
+                "login browser.\n"
+                "Run notebooklm login from an interactive Windows session, or, if the "
+                "browser profile is accessible, try: notebooklm login --browser-cookies chrome"
+            )
+            io.fail(1)
+        raise
 
     repair_playwright_account_metadata(plan.storage_path, io, page_html=result.page_html)
 
--- a/src/notebooklm/cli/services/playwright_login.py
+++ b/src/notebooklm/cli/services/playwright_login.py
@@ -458,29 +458,17 @@ def run_playwright_login(plan: PlaywrightLoginPlan, io: LoginIO) -> None:
     io.emit(f"[yellow]Opening {browser_label} for Google login...[/yellow]")
     io.emit(f"[dim]Using persistent profile: {plan.browser_profile}[/dim]")
 
-    try:
-        result: CaptureResult = run_browser_capture(
-            BrowserCapturePlan(
-                browser=plan.browser,
-                browser_profile=plan.browser_profile,
-                storage_path=plan.storage_path,
-                include_domains=plan.include_domains,
-            ),
-            io,
-            headless=False,
-            interactive=True,
-        )
-    except Exception as exc:
-        if sys.platform == "win32" and "spawn unknown" in str(exc).lower():
-            io.emit(
-                "[red]Chromium could not start in this Windows session.[/red]\n"
-                "Playwright needs an interactive desktop or window station to open the "
-                "login browser.\n"
-                "Run notebooklm login from an interactive Windows session, or, if the "
-                "browser profile is accessible, try: notebooklm login --browser-cookies chrome"
-            )
-            io.fail(1)
-        raise
+    result: CaptureResult = run_browser_capture(
+        BrowserCapturePlan(
+            browser=plan.browser,
+            browser_profile=plan.browser_profile,
+            storage_path=plan.storage_path,
+            include_domains=plan.include_domains,
+        ),
+        io,
+        headless=False,
+        interactive=True,
+    )
 
     repair_playwright_account_metadata(plan.storage_path, io, page_html=result.page_html)
 
--- a/tests/unit/cli/test_playwright_login_render_contract.py
+++ b/tests/unit/cli/test_playwright_login_render_contract.py
@@ -649,19 +649,6 @@ class TestLoginProgressSuccess:
 # ---------------------------------------------------------------------------
 class TestLoginErrorRender:
     @pytest.mark.requires_playwright
-    def test_windows_spawn_unknown_explains_interactive_session_requirement(self, runner):
-        with patch.object(_pl, "sys", SimpleNamespace(platform="win32")):
-            result, _ = _drive_login(
-                runner,
-                launch_side=Exception("BrowserType.launch_persistent_context: spawn UNKNOWN"),
-            )
-
-        assert result.exit_code == 1
-        assert "Chromium could not start in this Windows session" in result.output
-        assert "interactive desktop or window station" in result.output
-        assert "notebooklm login --browser-cookies chrome" in result.output
-
-    @pytest.mark.requires_playwright
     def test_retry_exhausted_connection_error_help(self, runner):
         from playwright.sync_api import Error as PlaywrightError
 
--- a/tests/unit/cli/test_playwright_login_render_contract.py
+++ b/tests/unit/cli/test_playwright_login_render_contract.py
@@ -649,6 +649,19 @@ class TestLoginProgressSuccess:
 # ---------------------------------------------------------------------------
 class TestLoginErrorRender:
     @pytest.mark.requires_playwright
+    def test_windows_spawn_unknown_explains_interactive_session_requirement(self, runner):
+        with patch.object(_pl, "sys", SimpleNamespace(platform="win32")):
+            result, _ = _drive_login(
+                runner,
+                launch_side=Exception("BrowserType.launch_persistent_context: spawn UNKNOWN"),
+            )
+
+        assert result.exit_code == 1
+        assert "Chromium could not start in this Windows session" in result.output
+        assert "interactive desktop or window station" in result.output
+        assert "notebooklm login --browser-cookies chrome" in result.output
+
+    @pytest.mark.requires_playwright
     def test_retry_exhausted_connection_error_help(self, runner):
         from playwright.sync_api import Error as PlaywrightError
 
--- a/tests/unit/cli/test_rookiepy_errors.py
+++ b/tests/unit/cli/test_rookiepy_errors.py
@@ -43,8 +43,8 @@ def test_decryption_branch(error_text):
     assert "Could not decrypt edge cookies" in msg
     assert "Keychain access" in msg
     assert "Windows" in msg
-    assert "DPAPI" in msg
-    assert "interactive user session" in msg
+    assert "App-Bound Encryption (ABE)" in msg
+    assert "--browser-cookies firefox" in msg
 
 
 def test_generic_branch_includes_original_message():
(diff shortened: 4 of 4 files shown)


teng-lin (changes requested):
Thanks for this — the structure is right (CLI-side messaging, `NoReturn` `io.fail` used correctly, the bare `raise` correctly re-raising non-matching exceptions), and CI is green across all 15 matrix legs. But two of the technical claims in the messages are wrong, and one of them is now asserted in a test, so I can't merge as-is.

### Blocking 1 — the DPAPI claim is incorrect, and the advice it implies doesn't work

```
On Windows, DPAPI cookie decryption requires the same interactive user session that owns the browser profile.
```

Chrome 127+ and current Edge don't use DPAPI for this — they use **App-Bound Encryption (ABE)**, where the key is bound to the browser process via a Windows service. The distinction matters because it inverts the remediation: with ABE, running in the same interactive user session that owns the profile **still fails**. There is no session, privilege, or flag that makes external decryption work.

This repo already documents that, in `docs/troubleshooting.md` (the Windows `Could not decrypt` section):

> Chrome 127+ (and current Edge) protect the cookie database with **App-Bound Encryption (ABE)**: the decryption key is bound to the browser process via a Windows service, so no external process can read it. This blocks every cookie-extraction library (`rookiepy`, `browser-cookie3`, `pycookiecheat`), not just `notebooklm-py`. There is no flag that bypasses ABE.

So the new message tells a Windows user to do something that cannot succeed, and contradicts our own docs. Suggested replacement, reusing the doc's own framing: name ABE, say plainly that no external process can read the cookies, and point to the workaround that actually works — `--browser-cookies firefox` (Firefox stores cookies in plain SQLite, no ABE).

`tests/unit/cli/test_rookiepy_errors.py` now asserts `"DPAPI" in msg`, which would lock the error in as contract — please update alongside the message.

### Blocking 2 — the `spawn UNKNOWN` message routes users into a path that's broken on Windows

```
try: notebooklm login --browser-cookies chrome
```

Per Blocking 1, `--browser-cookies chrome` on Windows fails with ABE. This sends a user from one failure straight into another known-broken one — and it's exactly what happened to the reporter of #2004, who hit `spawn UNKNOWN` and then got `Could not decrypt chrome cookies` on the fallback.

Better fallbacks: `--browser chrome` / `--browser msedge` (note: **not** `--browser-cookies`), which launch the system browser via Playwright's channel mechanism from `Program Files` — a different path from `%LOCALAPPDATA%\ms-playwright`, so it sidesteps a path-scoped block. Failing that, the documented headless/CI path in `docs/installation.md`: sign in on a machine with a display, then copy `storage_state.json` or set `NOTEBOOKLM_AUTH_JSON`.

### Important — the "interactive desktop or window station" diagnosis is probably wrong

`spawn UNKNOWN` is libuv's `UV_UNKNOWN`: the Win32 error returned by `CreateProcessW` had **no entry in libuv's translation table**. That's diagnostically narrowing:

- a missing or partial executable surfaces as `spawn ENOENT` (and Playwright pre-checks existence anyway)
- a plain ACL denial surfaces as `spawn EACCES`
- what actually falls through to `UNKNOWN` is policy/AV-shaped: `ERROR_ACCESS_DISABLED_BY_POLICY` (AppLocker / SRP / WDAC) or `ERROR_VIRUS_INFECTED` / `ERROR_VIRUS_DELETED`

Upstream reports of this exact error resolve the same way — microsoft/playwright#35363 (group policy blocking execution inside the `ms-playwright` directory), #28307, #28858.

A missing window station doesn't make `CreateProcess` fail: a GUI process spawns fine in a non-interactive window station and only fails later, when it tries to create a window. Since this failure is at spawn time, an execution veto on the binary is the better explanation. Worth reflecting in the message — telling an AppLocker-blocked user to "run from an interactive session" leaves them stuck.

Suggested direction: name it as a Windows execution veto (AppLocker/WDAC/endpoint security blocking `%LOCALAPPDATA%\ms-playwright`), suggest `--browser chrome`, and state that headless does **not** help — it spawns the identical binary through the identical path.

### Nit — test patches the whole `sys` module

```python
with patch.object(_pl, "sys", SimpleNamespace(platform="win32")):
```

This swaps out the entire module in that namespace, so it breaks if anything on that code path later touches `sys.executable` (the chromium pre-flight does, and there's in-flight work in that area). `patch.object(_pl.sys, "platform", "win32")` is narrower and won't rot.

### Coordination

There's parallel work touching `_auth/browser_capture.py` — specifically the launch-failure branch gated on `browser in CHANNEL_BROWSERS`, which excludes bundled `chromium` and is *why* this failure reached the user as a raw traceback in the first place. That work deliberately does not touch `playwright_login.py`, so it should compose with this PR rather than conflict. There's also in-flight work on `ensure_chromium_installed` in this same file; whichever merges second will need a rebase.

Happy to see this land once the ABE/DPAPI correction and the fallback command are fixed — the instinct to convert a raw traceback into guidance is exactly right.

teng-lin:
One more finding, verified rather than inferred — it's a **merge-order hazard**, not a criticism of the approach here.

There's parallel work extending the launch-failure branch in `_auth/browser_capture.py` to bundled Chromium (today it's gated on `browser in CHANNEL_BROWSERS`, which is why this failure reached the user as a raw traceback). Once that lands, **this PR's `spawn UNKNOWN` handler becomes unreachable**, and its test goes vacuous without failing.

The mechanism:

- This PR wraps `run_browser_capture(...)` in `except Exception as exc:`.
- The other change classifies the failure *inside* `run_browser_capture` and terminates via `io.fail(1)`.
- For the CLI sink, `io.fail` → `exit_with_code` (`src/notebooklm/cli/playwright_login_io.py:73`) → `raise SystemExit` (`src/notebooklm/cli/error_handler.py:58-60`).
- `SystemExit` derives from `BaseException`, **not** `Exception` — confirmed: `issubclass(SystemExit, Exception)` is `False`.

So the inner handler wins and `except Exception` never sees it. Your test would keep passing on this branch and only quietly stop exercising anything after both merge — the worst kind of regression, since nothing goes red.

The resolution I'd suggest: let the neutral core own the launch-failure classification, and **keep the `rookiepy_errors.py` half here** (with the ABE correction from my review above). The core placement covers callers this file can't reach — the layer-3 headless re-auth arm and any non-CLI `_auth` consumer — whereas a `playwright_login.py`-level handler only covers the interactive CLI path.

That means the most useful shape for this PR is: keep and fix the cookie-decryption message, and drop the `playwright_login.py` handler in favour of the core one. If you'd rather keep the handler here, that's workable too — but then the core change needs to not terminate, and someone has to own that ordering explicitly rather than discovering it after the fact.

Also confirmed while checking: `run_cdp_capture` has no equivalent exposure — it uses `connect_over_cdp` and never spawns a process, so `spawn UNKNOWN` is unreachable there. No change needed.

Sanjays2402:
You're right on both, and I've taken the narrower shape you suggested. Pushed 3c55278.

The DPAPI line was wrong — I wrote it from the old pre-127 model and didn't check the repo's own troubleshooting doc, which already says ABE and already names the working escape hatch. The message now says Chrome 127+/current Edge bind the key to the browser process via a Windows service so no external process can read it and no flag bypasses it, and points at `notebooklm login --browser-cookies firefox`. The test assertion moved from `"DPAPI" in msg` to the ABE wording plus the firefox command, so the contract locks in the correct claim rather than the wrong one.

On the spawn UNKNOWN handler: I dropped it entirely, along with its test. Your SystemExit argument settles it — `io.fail` raises through `exit_with_code`, `SystemExit` is a `BaseException`, so once the core classifies the launch failure my `except Exception` never fires and the test would keep passing against nothing. Classifying inside `run_browser_capture` also covers the layer-3 headless re-auth arm and non-CLI `_auth` callers that `playwright_login.py` can't see, so that's the right owner. No point keeping a handler here that has to be deleted the moment the other change lands.

I'll also drop the interactive-desktop diagnosis rather than relocate it — your libuv reading is more precise than mine. ENOENT/EACCES having their own codes does narrow `UV_UNKNOWN` to something without a translation entry, and `ERROR_ACCESS_DISABLED_BY_POLICY` / `ERROR_VIRUS_INFECTED` fit that and the playwright#35363 thread. A GUI process does spawn fine in a non-interactive window station and only fails at window creation, so a spawn-time failure isn't that. If you want the core-side message to say "execution veto on `%LOCALAPPDATA%\ms-playwright` (AppLocker/WDAC/endpoint security), try `--browser chrome`, headless does not help", I'm happy to write it there — say the word and I'll open a PR against whichever branch the `browser_capture.py` work lands on rather than pre-empting it here.

Also fixed the `patch.object(_pl, "sys", ...)` nit by removing the test that used it, so nothing on that path stubs the whole module anymore.

`tests/unit/cli/` is 2042 passed / 67 skipped locally, ruff clean.

Issue #2013, opened 2026-07-26
`notebooklm login` never detects success after the NotebookLM -> Gemini Notebook rebrand (notebook.google.com)


MatheusF9516:
## Bug

Since Google's rebrand of NotebookLM to "Gemini Notebook" (2026-07-16, noted in #1978), `notebooklm login` times out after 5 minutes even after a successful, completed Google sign-in in the opened Chromium window.

## Root cause

`url_matches_base_host()` in `src/notebooklm/cli/services/playwright_login.py` (and the `page.wait_for_url(f"{get_base_url()}/**", ...)` call at ~line 864) only match the configured base host (`notebooklm.google.com`). After the rebrand, the web app now redirects the browser to `notebook.google.com`, so:

- `page.wait_for_url` never matches -> "Login not detected within 5 minutes."
- Even if that check were relaxed, the follow-up defense-in-depth check (`if not url_matches_base_host(page.url): ... "Unexpected URL after login"`) would still fail for the same reason.

## Confirmed workaround (tested locally)

Patching `url_matches_base_host` to also accept `notebook.google.com`, and changing the `wait_for_url` call to use that predicate instead of a glob string tied to `get_base_url()`, fixes login end-to-end:

```python
def url_matches_base_host(url: str) -> bool:
    current_host = (urlparse(url).hostname or "").lower()
    return current_host in (get_base_host().lower(), "notebook.google.com")
```

```python
page.wait_for_url(url_matches_base_host, wait_until="commit", timeout=300_000)
```

After this, `notebooklm auth check --test --json` returns `"status": "ok"` with `"token_fetch": true`, and `notebooklm list` successfully returns real notebooks. Cookies end up set on both `notebooklm.google.com` and `notebook.google.com`, and the `batchexecute` RPC endpoint at the original `notebooklm.google.com` host keeps working unchanged — only the login-detection host check needed updating, not `NOTEBOOKLM_BASE_URL` / `_ALLOWED_BASE_HOSTS`.

## Environment

- notebooklm-py 0.7.3 (PyPI) and main branch (0.8.0rc1) — same code path, both affected
- Windows 11, Python 3.14.5
- Chromium bundled via Playwright

## Suggested fix

[...]

PR #2020, opened 2026-07-28
fix(auth): request + rank notebook.google.com binding cookies (#2013)

dan1t0:
## Summary

After the July 2026 NotebookLM → Gemini Notebook rebrand (#1978), Google now serves the app from `notebook.google.com` in addition to the legacy `notebooklm.google.com`, and sets the per-product binding cookies (`OSID` / `__Secure-OSID`) on the rebrand host.

PR #2015 made the Playwright login flow recognize `notebook.google.com` as a valid landing page (`url_matches_base_host`), but the **cookie-domain policy** in `_auth/cookie_policy.py` was not updated to match. This PR closes that gap.

## Root cause

- `REQUIRED_COOKIE_DOMAINS` did **not** include `notebook.google.com` / `.notebook.google.com`, so `notebooklm login --browser-cookies` (rookiepy extraction) never asked the browser for cookies on the rebrand host. A user whose browser only populated `OSID` on `notebook.google.com` could fail the Tier-2 secondary-binding check.
- `_auth_domain_priority` returned tier 0 (unranked) for `notebook.google.com`, vs tier 2/3 for `notebooklm.google.com` — so a same-name cookie on the rebrand host could be silently shadowed.

Confirmed via the live auth check: Google now sets `OSID` + `__Secure-OSID` on the bare `notebook.google.com` host (host-only, no leading dot), in addition to `notebooklm.google.com`.

## Changes

- `src/notebooklm/_auth/cookie_policy.py`:
  - Add `notebook.google.com` + `.notebook.google.com` to `REQUIRED_COOKIE_DOMAINS` (both dotted/bare variants, matching the `notebooklm.google.com` pattern).
  - Add matching priority tiers (3 / 2) in `_auth_domain_priority` so the rebrand host's binding cookies resolve consistently.
- `tests/unit/test_auth_cookie_policy.py`: Extended the priority parametrize cases and added `test_gemini_notebook_rebrand_host_matches_app_host_tiers`.
- `tests/unit/test_cookie_domain_split.py`: Added the rebrand host to `test_required_set_includes_runtime_auth_domains` and a new `TestGeminiNotebookRebrandHost` class covering: REQUIRED-set membership, default extraction requests it, runtime gate accepts it, `OSID`-only-on-rebrand-host survives extraction, and the legacy app host outranks the rebrand host when both carry `OSID`.
- `CHANGELOG.md`: Added entry under `[Unreleased]`.

Complements #2015 (login-host recognition) — together they complete the Gemini Notebook rebrand support for the auth path.

## Verification

```bash
# Lint + typecheck
uv run ruff check src/notebooklm/_auth/cookie_policy.py tests/unit/test_auth_cookie_policy.py tests/unit/test_cookie_domain_split.py  # All checks passed
uv run ruff format --check ...                                                                                                     # 3 files already formatted
uv run mypy src/notebooklm                                                                                                         # Success: no issues found in 314 source files

# Full test suite
uv run pytest tests/unit/ tests/server/ tests/_guardrails/ -q -n auto --dist=worksteal
# 11547 passed, 64 skipped, 1 xfailed, 20 warnings in 21.67s

# Live verification (existing cookies, post-rebrand)
uv run notebooklm auth check --test --json
# {"cookies_present": true, "token_fetch": true, "notebook_count": 3, ...}

uv run notebooklm list --json
# {"count": 3, ...}
```

Closes the cookie-policy half of #2013.

## Bug Fixes

* Fixed browser-cookie authentication for Gemini Notebook’s rebranded `notebook.google.com` domain.
* Ensured `OSID` and `__Secure-OSID` cookies are recognized from both legacy and rebranded Notebook domains.
* Prevented authentication failures when cookies exist only on the rebranded domain.
* Improved consistent cookie prioritization when legacy and rebranded domain cookies are both present.


teng-lin:
Cross-reference: there is an interaction between this PR and **#2044** (which makes the RPC-health / bundle-drift scripts use a domain-preserving cookie jar instead of a flat `name -> value` dict that pinned every cookie to `.google.com`).

Once this PR starts capturing **host-only** `OSID` / `__Secure-OSID` on `notebook.google.com`, a correctly domain-scoped jar will not send those cookies to `notebooklm.google.com` — and `_env.py` restricts `_ALLOWED_BASE_HOSTS` to `notebooklm.google.com` / `notebooklm.cloud.google.com`, so `get_base_url()` can never target the rebrand host. Requests keep going to the legacy host.

That is fine **if** the legacy host still gets its own `OSID` after the rebrand. If Google now mints the binding cookie only on `notebook.google.com`, the correctly-scoped jar would withhold the only one that exists, and the symptom would be an `"Authentication expired or invalid"` error indistinguishable from a genuine expiry (see #2038 for why that message has been misleading).

Do you have a live capture showing whether `notebooklm.google.com` still carries its own `OSID` post-rebrand? That would settle it. If it doesn't, this PR may also need to widen what host the client actually talks to — which is a bigger change than the cookie policy alone, and would be worth splitting out.

No action needed on this PR unless that capture shows the legacy host is bare. Flagging so the two changes aren't landed independently and debugged jointly.

teng-lin:
**Resolved — the two changes compose safely. No action needed on either PR.**

I checked a live, in-use profile (written today, currently serving a running MCP deployment). Cookie names and domains only, no values:

```
.google.com             n=17   OSID-ish: []
accounts.google.com     n= 5   OSID-ish: []
notebook.google.com     n= 2   OSID, __Secure-OSID   (__Secure-OSID expires 2028-08-02)
notebooklm.google.com   n= 2   OSID, __Secure-OSID   (__Secure-OSID expires 2028-07-17)
```

**Both hosts carry their own `OSID` / `__Secure-OSID`.** So a correctly domain-scoped jar targeting `notebooklm.google.com` sends that host's own binding cookie, and withholds the `notebook.google.com` one — which is the correct behaviour, not a loss. The concern that #2020's rebrand-host cookie would be the *only* one available does not hold.

The expiry dates are informative: the legacy-host `__Secure-OSID` was minted ~2026-07-17 and the rebrand-host one ~2026-08-02. The cutover minted a new host-scoped cookie alongside the existing one rather than replacing it.

**Residual unknown, worth stating rather than hiding:** this is one profile, and its legacy-host cookie predates the cutover. A brand-new login performed today might mint only on `notebook.google.com`. Closing that requires a fresh interactive login into an isolated storage path (`notebooklm login --storage <tmp>`) and inspecting the resulting domains — cheap, but it needs a human at a browser. Until someone does it, treat "a fresh post-rebrand login still yields a legacy-host OSID" as unverified.

That residual does not block either PR: if it ever bites, the symptom is a bare `OSID` absence, and #2044's jar-scope diagnostics now print `name@domain` precisely so it is visible instead of surfacing as a bogus "Authentication expired".

PR #2036, opened 2026-08-03
chore: release v0.8.0

teng-lin:
## Summary

Cut the final **v0.8.0** release (from `v0.8.0rc1`):

- Bump version `0.8.0rc1` → `0.8.0` in `pyproject.toml`, `uv.lock`, `desktop-extension/manifest.json`
- Fold the post-rc1 `[Unreleased]` fixes (#2026, #2011, #2013, #1994) into the `[0.8.0]` changelog section and date the heading (2026-08-03)
- Bump locked `mcp` 1.27.2 → 1.29.0 (transitive via `fastmcp==3.4.2`) to clear Dependabot alert #15 (high: WebSocket server transport missing Host/Origin validation, patched in 1.28.1)

> Note: the `uv.lock` diff also touches resolution markers for `aiofile` and `rpds-py` (Python 3.14 / Windows wheel-metadata branches). That's re-lock churn from uv picking up new upstream metadata during the `mcp` upgrade, not an intentional change.

## Release plan (post-merge)

1. `gh workflow run testpypi-publish.yml --ref main`
2. `gh workflow run verify-package.yml --ref main -f source=testpypi` (incl. E2E)
3. Tag `v0.8.0` on the merge commit → `publish.yml` → production PyPI

## Test plan

- [x] `ruff format --check .` / `ruff check .` — clean
- [x] `mypy src/notebooklm --ignore-missing-imports` — clean
- [x] `pytest` with browser/dev/markdown/impersonate/mcp/server extras — 12442 passed, 64 skipped, 1 xfailed
- [x] CI green (20/20 checks)
- [ ] TestPyPI staged verify (post-merge)

PR #2041, opened 2026-08-03
fix(download): audio overviews are .m4a/audio/mp4, not .mp3 (#2034) + REST spool-name honours output_format

### LINKED ISSUE #2034: `download audio` writes AAC/M4A content with a `.mp3` extension
**Version:** notebooklm-py 0.7.3 (pip, `[browser]` extra) · Python 3.12 · macOS

**What happens**

`notebooklm download audio` (audio overview) saves the file with a `.mp3` extension, but the content is actually AAC in an ISO Media (M4A/DASH) container:

```
$ file audio.mp3
audio.mp3: ISO Media, MPEG v4 system, Dynamic Adaptive Streaming over HTTP

$ ffprobe -v error -show_entries format=format_name:stream=codec_name,codec_type \
    -of default=noprint_wrappers=1 audio.mp3
codec_name=aac
codec_type=audio
format_name=mov,mp4,m4a,3gp,3g2,mj2
```

**Why it matters**

Anything that trusts the extension (players, transcoders, MIME-sniffing upload endpoints) mishandles the file. Renaming to `.m4a` fixes downstream handling — the extension is the only wrong part, which makes it easy to lose hours on.

**Suggestion**

Derive the extension from the response `Content-Type` (or magic bytes) before writing, or default audio overviews to `.m4a`.

teng-lin:
Fixes #2034

**Scope — this PR fixes two bugs of the same class, both in the diff:**

1. **#2034 (type axis):** `download audio` labelled AAC/MP4 bytes `.mp3`. Fixed to `.m4a` / `audio/mp4` on CLI, MCP and REST.
2. **Format axis (found while fixing the above):** the REST `/download` route named its spool file from the type's *default* extension regardless of `output_format`, then served that name and derived Content-Type from it — so `--format pptx` was served as `artifact.pdf` / `application/pdf`. Not scope creep: shipping the type-axis fix while leaving the format axis broken would be half a fix, and the same `resolve_extension` rule repairs both. Codex reached this one independently during review.

## The bug

`notebooklm download audio` wrote the audio-overview file with a `.mp3` extension, but the bytes are AAC in an ISO-BMFF/MP4 container. Anything that trusts the extension — players, transcoders, MIME-sniffing upload endpoints — mishandled the file, and because the extension is the *only* wrong part it is easy to lose hours on.

## The bytes were never MP3

Two independent lines of in-repo evidence, no live probe needed:

1. **Every audio artifact row on the wire advertises `audio/mp4`.** Across `tests/cassettes/*.yaml` there are **84 occurrences of `audio/mp4` and zero of `audio/mpeg`**. The rows carry `[url, code, mime]` media triples and the audio ones are uniformly `audio/mp4` — which is exactly the MIME type of `.m4a`.

2. **The codebase already knew.** `src/notebooklm/_row_adapters/artifacts.py:372` reads *"Audio Overview media URL, preferring the `audio/mp4` entry"* — the adapter explicitly selects the `audio/mp4` variant when picking the download URL, and then the spec labelled the result `.mp3`.

Both match the reporter's `ffprobe` (`codec_name=aac`, `format_name=mov,mp4,m4a,…`).

Video is unaffected (`video/mp4` → `.mp4`). Infographic/slide-deck rows carry no MIME at all, and report / mind-map / data-table / quiz / flashcards are formatted locally, so audio was the only type-axis mismatch.

## Why not derive the extension from `Content-Type`

The issue suggests deriving from the response header or sniffing magic bytes. Neither is structurally possible here:

- **The filename is chosen before the fetch.** `_app/download.py` resolves the output path in `build_download_plan` / `_execute_download_single`, and only then does `_artifact/downloads.py::download_url` issue the request. Header-derived naming would mean writing and then renaming, which breaks the "an explicit output path is honoured verbatim" contract.
- **The MCP broker never fetches at all.** `_broker_download` mints `filename` + `mime_type` into a *signed link* without touching the artifact — its own docstring says `size_bytes` is `None` because it "can't be known without eagerly fetching the artifact, which this must not do". There is no response to read a header from on that path.

So the extension stays a static spec field, corrected to `.m4a`. If per-artifact variance ever appears there is already a pre-fetch source of truth to switch to — the artifact row's own `audio/mp4` — but today it is invariant, so a constant is the honest encoding.

## Fixed on all three surfaces

The `DownloadTypeSpec` registry is deliberately rebuilt three times (layering forbids `server/` and `mcp/` from importing `cli/`), so the fix touches all three:

| Surface | File |
|---|---|
| CLI `download audio` | `cli/_download_specs.py` |
| MCP `studio_download` + `/files/dl` | `mcp/tools/_studio_download.py` |
| REST `POST .../artifacts/download` | `server/routes/artifacts.py` |

The extension→MIME table moves into the neutral `_app/download.py` (`EXTENSION_MIME_TYPES` / `mime_type_for_extension`) so all three advertise the same Content-Type from one place.

### Why the REST route now sets `media_type` explicitly

`.m4a` is **absent from Python's builtin `mimetypes` map**. Verified:

```python
>>> mimetypes.MimeTypes(filenames=(), strict=True).types_map[True].get('.m4a')
None      # builtin table — no .m4a row
>>> mimetypes.guess_type('x.m4a')
('audio/mp4', None)   # only because THIS host ships /etc/mime.types
```

Left to guess, `FileResponse` falls back to `guess_type(...)[0] or "application/octet-stream"` on any host without an `/etc/mime.types` — i.e. a slim deploy container. (**Correction to an earlier revision of this description**, which said `text/plain`: that is the base `Response` default, not `FileResponse`'s. `application/octet-stream` is still the wrong answer for audio — it defeats inline playback and any client dispatching on MIME — so the conclusion is unchanged, but the stated mechanism was wrong and is now verified against Starlette's source.) This is not hypothetical: inside the running `deploy-notebooklm-mcp-1` container `guess_type('x.m4a')` returns `(None, None)` (verified independently in that container; on this host it returns `('audio/mp4', None)` because this host ships a mime database). So the route passes `media_type` from the shared table instead of relying on the ambient mime database.

## Second bug, same class, found while fixing this

The REST route named its spool file `artifact{spec.extension}` — the type's **default** extension — regardless of `output_format`, then served that basename as the download filename and derived Content-Type from its suffix. So:

- `--format pptx` slide deck → served as `artifact.pdf` / `application/pdf`
- `--format markdown` quiz → served as `artifact.json` / `application/json`

Same defect as #2034 on the *format* axis instead of the *type* axis. The MCP `/files/dl` route already resolved both format-aware and was unaffected; this brings REST in line. The rule now lives in `_app/download.py::resolve_extension` and every spool site uses it, so the invariant greps clean.

`tests/server/test_artifacts.py::test_download_pptx_is_not_served_as_pdf` covers it — confirmed it fails without the fix with `attachment; filename="artifact.pdf"`.

## Guardrail

Nothing previously enforced that the three duplicated registries agree — `test_generation_kind_exhaustiveness.py` checks names/kinds/`download_attr` but never `extension`. That absence is exactly how a one-surface fix ships half-broken, so this adds `tests/_guardrails/test_download_spec_parity.py`: registries agree on every behavioural field, audio pinned to `.m4a`/`audio/mp4`, and the extension↔MIME table is total in both directions (no unmapped extension, no unreachable row — a stale `.mp3 → audio/mpeg` row is how a corrected mapping drifts back).

It immediately earned its keep by catching a real divergence: the CLI's slide-deck row keeps the legacy Click param name `format_param_name="slide_format"` while MCP/REST use `output_format`. That field is adapter-local, so it is excluded with a documented rationale rather than "fixed".

## ⚠ Migration

**Only *derived* filenames change.** An explicit output path is still honoured verbatim, so `notebooklm download audio out.mp3` keeps writing `out.mp3` (with the same, still-AAC bytes) — covered by a regression test.

What changes: the default name when you pass no path, the per-file names under `--all`, and the `filename` / `mime_type` the MCP and REST surfaces advertise.

- Scripts globbing `*.mp3` should glob `*.m4a` (or both during a transition).
- Re-running `download audio --all <dir>` after upgrading writes new `.m4a` files **alongside** any `.mp3` files a previous run left there — the names no longer collide, so nothing is overwritten or auto-renamed.

No deprecation runway: the file was never MP3, so there is no state in which the old name was correct. Per the repo's breaking-change policy, runways are for real contracts; a wrong file extension is a bug. CHANGELOG carries the migration note under `[Unreleased] → Fixed`.

## Verification

> **CI status.** The post-v0.8.0 allowlist prune landed in #2037, so `Code Quality` is green and the 15-job `Test` matrix (3 OS x 5 Python) is executing on this branch for the first time — rebased onto `27926aa4`. Per-leg results are reported in a comment below. The numbers in this block are the local run.

```
mypy src/notebooklm --ignore-missing-imports   Success: no issues found in 314 source files
pytest tests -q                                12459 passed, 64 skipped, 1 xfailed in 498.06s
ruff check . && ruff format --check .          All checks passed! / 951 files already formatted
```

Docs updated: `cli-reference` (the extension table now reads "AAC audio in an MP4 container"), `python-api`, `installation`, `mcp-guide`, `README`, `SKILL.md`, `examples/quickstart.py`.

https://claude.ai/code/session_01WddM9PnqkU8S3wmFYPorD5

## Summary by CodeRabbit

* **Bug Fixes**
  * Audio Overview downloads now use `.m4a` filenames and `audio/mp4` metadata across CLI, MCP, and REST.
  * REST artifact downloads now use the requested output format’s correct filename extension and MIME type, including non-default formats.
  * Explicit output paths remain unchanged.

* **Documentation**
  * Updated quick starts, guides, CLI references, and examples to show `.m4a` audio downloads and AAC audio in an MP4 container.

* **Tests**
  * Added coverage validating consistent download behavior and metadata across supported interfaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   > the default name when you pass no path, the per-file names under `--all`,
   > and the `filename` / `mime_type` the MCP and REST surfaces advertise.
   > Scripts that glob `*.mp3` in a download directory should glob `*.m4a`
-  > (or `*.m4a` and `*.mp3` during a transition).
+  > (or `*.m4a` and `*.mp3` during a transition). Note that re-running
+  > `download audio --all <dir>` after upgrading writes new `.m4a` files
+  > *alongside* any `.mp3` files a previous run left there — the names no
+  > longer collide, so nothing is overwritten or auto-renamed.
+
+- **REST `POST /v1/notebooks/{id}/artifacts/download` no longer mislabels a
+  non-default `output_format`.** The route named its spool file from the type's
+  *default* extension, then served that name as the download filename and
+  derived `Content-Type` from it — so a `pptx` slide deck arrived as
+  `artifact.pdf` / `application/pdf`, and a `markdown` quiz as `artifact.json`.
+  Same defect class as the audio fix above, on the format axis instead of the
+  type axis. The MCP `/files/dl` route already resolved both format-aware and
+  was unaffected
+  ([#2034](https://github.com/teng-lin/notebooklm-py/issues/2034)).
 
 ## [0.8.0] - 2026-08-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Audio overviews now download as `.m4a`, not `.mp3`.** The Audio Overview
+  bytes have always been AAC in an ISO-BMFF/MP4 container — the artifact
+  metadata itself advertises them as `audio/mp4` — so the `.mp3` name was a
+  mislabel that broke players, transcoders, and MIME-sniffing upload endpoints
+  in ways that were hard to trace back to the extension. Fixed on all three
+  surfaces (CLI `download audio`, MCP `studio_download`, REST
+  `POST /v1/notebooks/{id}/artifacts/download`), and the advertised MIME type is
+  now `audio/mp4` instead of `audio/mpeg`
+  ([#2034](https://github.com/teng-lin/notebooklm-py/issues/2034)).
+
+  > **⚠ Migration.** Only the *derived* filename changed — an explicit output
+  > path is still honoured verbatim, so `notebooklm download audio out.mp3`
+  > keeps writing `out.mp3` (with the same, still-AAC bytes). What changes is
+  > the default name when you pass no path, the per-file names under `--all`,
+  > and the `filename` / `mime_type` the MCP and REST surfaces advertise.
+  > Scripts that glob `*.mp3` in a download directory should glob `*.m4a`
+  > (or `*.m4a` and `*.mp3` during a transition).
+
 ## [0.8.0] - 2026-08-03
 
 The headline of 0.8.0 is **integrations**: NotebookLM is now reachable from AI
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ notebooklm generate mind-map                       # interactive studio map (def
 notebooklm generate data-table "compare key concepts"
 
 # 5. Download artifacts
-notebooklm download audio ./podcast.mp3
+notebooklm download audio ./podcast.m4a
 notebooklm download video ./overview.mp4
 notebooklm download cinematic-video ./documentary.mp4
 notebooklm download quiz --format markdown ./quiz.md
@@ -274,7 +274,7 @@ async def main():
         # Generate content (podcast, video, quiz, etc.)
         status = await client.artifacts.generate_audio(nb.id, instructions="make it fun")
         await client.artifacts.wait_for_completion(nb.id, status.task_id)
-        await client.artifacts.download_audio(nb.id, "podcast.mp3")
+        await client.artifacts.download_audio(nb.id, "podcast.m4a")
 
         # Generate quiz and download as JSON
         status = await client.artifacts.generate_quiz(nb.id)
--- a/SKILL.md
+++ b/SKILL.md
@@ -252,7 +252,7 @@ Before starting workflows, verify auth is in place. **Use `--test --json` (not b
 | Check artifact status | `notebooklm artifact list` |
 | Wait for completion | `notebooklm artifact wait <artifact_id>` |
 | Delete artifact | `notebooklm artifact delete <artifact_id> --yes` |
-| Download audio | `notebooklm download audio ./output.mp3` |
+| Download audio | `notebooklm download audio ./output.m4a` |
 | Download video | `notebooklm download video ./output.mp4` |
 | Download cinematic video | `notebooklm download cinematic-video ./cinematic.mp4` (alias for `download video`) |
 | Download infographic | `notebooklm download infographic ./infographic.png` |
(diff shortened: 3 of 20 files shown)


teng-lin:
@claude review

teng-lin:
### CI status — no test signal yet, and it is not this PR

`Code Quality` is red here, but **not because of this diff**:

- The audit's own verdict is *"Public API is compatible with v0.8.0 (**0** reviewed break(s) allowlisted), but the allowlist has stale entries."*
- This diff does not touch `scripts/api-compat-allowlist.json`.
- Running `audit_public_api_compat.py --check-stale` on a **pristine `origin/main` worktree** (8fb61cb1, none of this branch's code) reproduces it byte-identically.
- **Every open PR is red on the same step** — confirmed on #2039 and #2042.

Cause: the v0.8.0 tag rebased the API-compat baseline, making every allowlist entry stale. `docs/releasing.md:411` documents this prune as mandatory and states in as many words that CI goes red until the prune PR lands; #2036 shipped without it. **#2037 carries the prune.**

**Consequence worth flagging:** `test.yml` declares `needs: quality`, so the `Test` matrix here reports `skipping` — **there is currently no CI test signal on this PR at all**. The green numbers in the description are local only, and I am not treating them as CI-green. Once #2037 lands I will rebase, let the matrix actually execute, and re-verify.

Local, whole tree, on this branch:

```
pytest tests -q         12459 passed, 64 skipped, 1 xfailed in 498.06s
mypy src/notebooklm     Success: no issues found in 314 source files
ruff check .            All checks passed!
ruff format --check .   951 files already formatted
```

teng-lin:
Thanks @claude — replying to the two non-blocking observations, since one of them is stronger than stated.

**On `Path(served).suffix` coupling to `_resolve_conflict`'s ` (N){ext}` renaming format.** Fair flag, and you're right that the renaming inserts the counter *before* the extension so the suffix survives. It's actually safer than that: in the REST route that code path is **unreachable**. The route spools into a freshly-created, empty `tempfile.mkdtemp()` and passes an explicit `output_path`, so in `_execute_download_single` the conflict check hits `not path.exists()` on the first try and returns the path untouched — `force`/`no_clobber` both default to `False` and never come into play. So the Content-Type derivation depends on the renaming format only in a branch that cannot execute here. I've left the code as-is rather than adding a guard for an unreachable state, but noting the reasoning so it's on the record if that spool-into-a-private-empty-dir property ever changes — that, not the renaming format, is the invariant actually holding this up.

**On leftover `.mp3` literals in tests.** Agreed with your read, and I audited the same set. They fall into three groups, none of them assertions about the audio spec's extension:
- `artifact_title_to_filename` fixtures, where the extension is an arbitrary parameter to a generic helper (`test_download_helpers.py`);
- mocked `download_audio` return values and `GenerationState.url` fixtures, which are opaque strings;
- `tests/e2e/test_file_upload.py`, which uploads a real MP3 *source file* — unrelated to the Audio Overview download path.

The ones in `tests/unit/cli/test_download.py` that pass an explicit `out.mp3` output path I deliberately **kept**: they now double as back-compat coverage proving an explicit `.mp3` path is still honoured verbatim after this change.

One correction to your summary for the record: the `otmP3b` match in `rpc/types.py` isn't a coincidental `.mp3` hit — it's an obfuscated RPC method id (GeneratePromptSuggestions) and is unrelated to file extensions entirely.

Issue #2038, opened 2026-08-03
extract_csrf_from_html reports "Authentication expired" for any Google page — the body scan makes a wrong-page failure indistinguishable from a real expiry (#2019)


teng-lin:
## Summary

When `SNlM0e` (or `FdrFJe`) cannot be extracted from a page, `extract_csrf_from_html` /
`extract_session_id_from_html` report

```
Authentication expired or invalid. Run 'notebooklm login' to re-authenticate.
```

**even when the session is perfectly valid** and the real problem is that we fetched the
wrong page. The message also discards the single most useful piece of evidence: the final URL.

## Root cause

`src/notebooklm/_auth/extraction.py` (drift path):

```python
if is_google_auth_redirect(final_url) or contains_google_auth_redirect(html):
    raise ValueError(
        "Authentication expired or invalid. Run 'notebooklm login' to re-authenticate."
    )
```

`contains_google_auth_redirect` (`src/notebooklm/_url_utils.py:109-124`) regex-extracts
*every* URL from the body and returns `True` if **any** points at `accounts.google.com`:

```python
url_pattern = r'https?://[^\s"\'<>]+'
urls = re.findall(url_pattern, text)
return any(is_google_auth_redirect(url) for url in urls)
```

Effectively every Google-served page qualifies — sign-in links, account-chooser menus,
footers, privacy links. So a genuine "wrong page / structure changed" condition is
swallowed by the auth branch and misreported as expiry.

This is already documented as a known hazard elsewhere in the same package —
`_auth/account.py::_probe_authuser`:

> Only checks the *final* URL for an auth redirect. The page body is not scanned because a
> healthy NotebookLM page legitimately contains many `accounts.google.com` links (account
> chooser, manage-account menu) that would fool `contains_google_auth_redirect`.

The extraction path never got the same treatment.

## Real-world impact (#2019)

The scheduled `rpc-health.yml` workflow failed daily **2026-07-28 -> 2026-08-03** with exactly
this false "expired" message. The credentials were valid the whole time — proved by a passing
nightly `E2E Tests (windows-latest)` on the same `NOTEBOOKLM_AUTH_JSON`.

The actual cause was a cookie-domain-flattening bug that sent the redirect chain to
`accounts.google.com/CookieMismatch` -> `support.google.com/accounts/answer/32050` (HTTP 200) —
a **help article** whose body contains `accounts.google.com` links, which tripped the body scan.

[...]

PR #2044, opened 2026-08-03
fix(scripts): preserve cookie domains in the RPC-health / bundle-drift auth path (#2019)

### LINKED ISSUE #2019: RPC Health Check: Authentication Failure
ERROR: Authentication expired or invalid. Run 'notebooklm login' to re-authenticate.
RPC Health Check (FULL mode)
============================================================

Fetching auth tokens...

teng-lin:
Fixes #2019. Also fixes the #2018 bundle-drift failures — same root cause, same commit.

> **The short version, if you read nothing else:** flattening cookies to a `name -> value` map is **data loss**, not just an over-broad send. When the same cookie name exists on two hosts — which is exactly the case for `OSID` — `flatten_cookie_map` keeps one and **silently discards the other**. The bundle-drift monitor still does this on `main` today, and #2037 does **not** mask it. Proof is in [Relationship to #2037](#relationship-to-2037), reproduced offline against the merged file-backed profile.

## It is not expired credentials

The auto-filed issue says `ERROR: Authentication expired or invalid. Run 'notebooklm login' to re-authenticate.`, and three people have since piled onto #2019 assuming a rotated cookie snapshot. Two independent pieces of evidence say otherwise.

**1. The secret is valid right now.** Today's nightly `E2E Tests (main@2ba6c313/windows-latest)` **passed** using the same `NOTEBOOKLM_AUTH_JSON` secret (run [30792136468](https://github.com/teng-lin/notebooklm-py/actions/runs/30792136468)). The rpc-health workflow's own "Verify auth secret is present" preflight also never fired — the job reached `Fetching auth tokens...` before dying.

**2. No repo change caused it.** The last green and the first red ran on the *same commit*:

| date | commit | rpc-health |
|---|---|---|
| 2026-07-26 | `45fd4258` | success |
| **2026-07-27** | **`7d0aa42c`** | **success** |
| **2026-07-28** | **`7d0aa42c`** | **failure** ← first red |
| 2026-07-29 → today | `7d0aa42c` … `38d25949` | failure, every run |

Nothing in the tree moved between the last green and the first red. What moved was Google: the `notebook.google.com` cutover.

## Root cause: the two scripts flatten cookie domains; the CLI does not

- `scripts/check_rpc_health.py` called `load_auth_from_storage()`, which returns a flat `name -> value` dict with domains dropped (`src/notebooklm/_auth/tokens.py`).
- Under `NOTEBOOKLM_AUTH_JSON`, `resolve_storage_path()` returns `None`, so `fetch_tokens(cookies, storage_path=None)` rebuilt the jar through `build_cookie_jar` → `normalize_cookie_map`, which pins **every** cookie to `.google.com`:
  ```python
  name, domain, path = key, ".google.com", "/"   # src/notebooklm/_auth/cookies.py
  ```
- `scripts/capture_rpc_registry.py` handed the same flat dict straight to `httpx.get(cookies=...)`. A plain mapping carries no domain at all, so httpx attaches it to every host in the redirect chain.

Host-scoped cookies (`OSID` / `__Secure-OSID` on the NotebookLM host, `LSID` on `accounts.google.com`) were therefore broadcast to all `*.google.com`. Post-cutover, the NotebookLM host redirects through `accounts.google.com/ServiceLogin` → `notebook.google.com/accounts/SetOSID`, which mints an `OSID` for the **new** host; the stale broadcast copy collides with it and the chain dead-ends at `accounts.google.com/CookieMismatch` → `support.google.com/accounts/answer/32050`, served **HTTP 200**.

**Flattening also destroys data outright, which is the more serious half.** A flat `name -> value` map has exactly one slot per name. `OSID` legitimately exists on *both* the app host and `accounts.google.com` with *different* values, so `flatten_cookie_map` resolves the collision by domain priority — the app host wins — and the accounts-scoped `OSID` is **silently dropped before any request is made**. The result is not "the right cookie plus some extras": it is the *wrong* value sent to the sign-in host and the correct one gone. `LSID` leaks the opposite direction at the same time. That is why this is worth landing on its own merits and not only as #2019 insurance — see the wire trace under [Relationship to #2037](#relationship-to-2037), which reproduces it against the *merged* file-backed profile.

Because that is a 200 on a non-`accounts` host, `is_google_auth_redirect()` never fires, `SNlM0e` is absent, and `extract_csrf_from_html`'s body-scan fallback reports "expired". Note the issue text carries **no** `Redirected to: …` line — that is the tell that it came from the body scan, not from the redirect check.

The CLI and library were never affected: `src/notebooklm/cli/auth_runtime.py` special-cases env auth into the domain-preserving `build_httpx_cookies_from_storage(None)`. That is exactly why the Windows E2E job is green on the same secret.

Reproduced offline against the library, no credentials needed:

```
flat dict (what the old scripts passed): {'SID': …, '__Secure-1PSIDTS': …, 'OSID': …, 'LSID': …}
OLD path, storage_path=None  (env-var auth): [('LSID', '.google.com'), ('OSID', '.google.com'), …]
OLD path, storage_path=<file>              : [('LSID', 'accounts.google.com'), ('OSID', 'notebooklm.google.com'), …]
```

## The fix

- **`scripts/check_rpc_health.py`** — `load_auth()` now returns `await AuthTokens.from_storage(storage_path)`, the same env-var-aware, domain-preserving loader the CLI uses, replacing the `fetch_tokens(...)` + hand-built `AuthTokens(...)` pair. The `FileNotFoundError` / `ValueError` / `httpx.HTTPError` → `sys.exit(2)` handling is preserved and consolidated into one place; all three arms now go through `scrub_secrets`, and each names the auth source or the path it checked.
- **`scripts/check_rpc_health.py` diagnostics** — the report now prints the auth source, the base host, the account route, and the jar's `name@domain` scopes (names and domains only, never values). #2019 was a cookie-*scoping* failure that produced a generic "authentication expired" line, so the nightly log carried nothing to diagnose it with. Both open questions below would have been self-evident in the failing run had this been there.
- **`scripts/capture_rpc_registry.py`** — passes `build_httpx_cookies_from_storage()` instead of the flat dict; the `_fetch` annotation becomes `httpx.Cookies | None`. The gstatic chunk reads stay unauthenticated.
- **`tests/unit/test_scripts_auth_cookie_domains.py`** (new guardrail) — the live rpc-health workflow cannot gate a PR, so this pins the contract offline, for **both** scripts and for **both** the env-var and the file-backed (#2037) configuration. The fixture deliberately carries `OSID` on *two* hosts with *different values*, because same-name-different-domain is the entire reason domains matter: a name-only assertion (`"OSID" not in sent`) would be wrong once the sign-in host legitimately has an `OSID` of its own. Assertions are therefore on `(name, value)` pairs — which is also what lets the guardrail catch the **data loss**, not just the broadcast: `assert ("OSID", "v-osid-accounts") in pairs` fails when flattening discards the accounts-scoped value, and the file-backed test makes the same claim against the persisted `storage_state` (`("OSID", "v-osid-accounts", accounts.google.com) in persisted_triples`). The mocked chain also mints a fresh host-scoped `OSID` via `Set-Cookie` — the exact mechanism #2019 describes — so the jar assertions must survive a rotation rather than pinning a jar that never changes. **Verified to fail against the pre-fix code**, on the intended assertions.
- **`tests/unit/test_check_rpc_health.py`** — the `ValueError` and `httpx.HTTPError` arms of `load_auth` had no coverage at all; the `httpx` one carries the `scrub_secrets` call that keeps `f.sid=<session_id>` out of the Actions log and out of the auto-filed issue body. Both are now pinned. The missing-storage test no longer patches anything (empty `NOTEBOOKLM_HOME` → the real loader raises for itself), and the file's `importlib.util.spec_from_file_location` preamble is gone: it built a *second* module object for `scripts/check_rpc_health.py`, so a patch applied to one copy left the other untouched.

### Open question for #2020 (worth resolving before both land)

#2020 (`fix(auth): request + rank notebook.google.com binding cookies`) reports, from a live auth check, that Google now sets `OSID` + `__Secure-OSID` on the bare `notebook.google.com` host, **host-only, no leading dot**. Host-only binding cookies are exactly the population flattening destroys, so at first glance the two PRs are complementary — and they do touch disjoint code (`_auth/cookie_policy.py` vs `scripts/`; only `CHANGELOG.md` overlaps).

But there is a wrinkle I do **not** want to paper over. `get_base_url()` can only return `notebooklm.google.com` or `notebooklm.cloud.google.com` (`_ALLOWED_BASE_HOSTS`, `src/notebooklm/_env.py`) — never `notebook.google.com`. A *correct*, domain-preserving jar therefore **withholds** a `notebook.google.com` host-only cookie from a request to `notebooklm.google.com`; the old flat dict sent it. Probed here:

```
JAR: [('OSID','notebook.google.com'), ('SID','.google.com'), ('__Secure-1PSIDTS','.google.com'), …]
GET https://notebooklm.google.com/  →  cookies: ['SID', '__Secure-1PSIDTS']      # OSID withheld
```

So once #2020 starts populating those cookies, they are inert on the token-fetch path — and, since `_kernel.py` hands the same jar to the client, on the library path too. That is a #2020 question rather than a defect in this PR (this PR only makes the scripts behave the way the CLI already did), but it should be settled between the two rather than after both land, because if it bites the symptom is the *same* "Authentication expired or invalid" string and nothing in the log distinguishes it. The new diagnostics line above is deliberately aimed at that: the jar scopes are printed, so a withheld `OSID` is visible.

## Relationship to #2037

**Rebased onto `27926aa4` (#2037 merged).** Clean rebase, no conflict — #2037 added no `[Unreleased]` CHANGELOG entry. This branch touches **zero** workflow files. The overlap, now **re-verified against the merged code rather than the PR diff**:

**How the merged workflow feeds this script.** `rpc-health.yml` now drops `NOTEBOOKLM_AUTH_JSON` from the run step's `env:` and writes the secret to `~/.notebooklm/profiles/default/storage_state.json`. So `resolve_storage_path()` returns a real `Path` in CI instead of `None`. That changes the *value* passed to `AuthTokens.from_storage(storage_path)`, not the call site or its shape — no code change was needed. Confirmed the layouts actually agree, since a mismatch would be a silent `FileNotFoundError`:

```
NOTEBOOKLM_HOME    : /tmp/tmp_2e4avxp
get_storage_path() : /tmp/tmp_2e4avxp/profiles/default/storage_state.json
#2037 writes       : profiles/default/storage_state.json          → MATCH
```

- **#2037 does mask the health-check half** — re-confirmed. With a real file, `build_cookie_jar` short-circuits to `build_httpx_cookies_from_storage(storage_path)`, so the pre-fix flattening never fires on that path.
- **#2037 does *not* mask the bundle-drift half** — re-confirmed, and it is worse than "unmasked". Probed against the merged file-backed configuration, pre-fix vs post-fix:

```
PRE-FIX (main today), flat dict → httpx.get(cookies=…), FILE-backed profile:
  -> notebooklm.google.com   SID=…; __Secure-1PSIDTS=…; OSID=v-osid-app; LSID=v-lsid
  -> accounts.google.com     SID=…; __Secure-1PSIDTS=…; OSID=v-osid-app; LSID=v-lsid   ← app OSID broadcast
POST-FIX, domain-preserving jar, same file:
  -> notebooklm.google.com   SID=…; __Secure-1PSIDTS=…; OSID=v-osid-app
  -> accounts.google.com     SID=…; __Secure-1PSIDTS=…; OSID=v-osid-accounts; LSID=v-lsid
```

Read the second row pre-fix carefully — there are **three** distinct failures in it, not one:

1. the app-host `OSID=v-osid-app` is **sent to the sign-in host**, which is the collision that dead-ends the chain;
2. the accounts-scoped `OSID=v-osid-accounts` is **gone** — destroyed by `flatten_cookie_map`'s duplicate-name resolution before the request was even built, so the sign-in host receives a value that was never its own;
3. `LSID=v-lsid`, scoped to `accounts.google.com`, **leaks onto the app host**.

Only (1) is an over-broad send. (2) is data loss and (3) is a credential leak across hosts. **On `main` as it stands the drift monitor still does all three, so #2018 will keep firing until this lands** — no amount of workflow-side auth hardening reaches it, because the flattening happens after the credentials are loaded, regardless of where they were loaded from.

So: for **#2018 this is the fix, unconditionally.** For **#2019** it is defense-in-depth plus regression protection — the flattening remains live on every env-var auth path (local runs, forks, anyone who exports the secret), and the guardrail is what stops it coming back.

**The guardrail covers the path the merged workflows actually take.** `test_check_rpc_health_auth_preserves_domains_for_file_backed_profile` drives `resolve_storage_path()` under a temp `NOTEBOOKLM_HOME`, which produces exactly the `profiles/default/storage_state.json` layout above — the #2037 configuration, including the cookie-persistence side that only exists on that branch. The env-var test still covers local dev and forks. Both were re-confirmed **red against the pre-fix scripts** after the rebase.

## Deliberately not fixed here

- **The misdiagnosis itself.** `src/notebooklm/_auth/extraction.py` concludes "authentication expired" from a body scan alone, so **any** `SNlM0e` miss reports as expiry — every signed-in Google page contains an `accounts.google.com` link. That is why #2019 collected three wrong-theory reports. Tracked separately in **#2038**, owned by someone else; untouched here.
- **`scripts/diagnose_get_notebook.py`** carries the same flat-`load_auth_from_storage()` shape. Naming it so it isn't lost, but leaving it: it is a manual one-off diagnostic for issue #114, no workflow runs it, and converting it also means replacing its hand-rolled `"; ".join(auth.cookies.items())` cookie header (a `DomainCookieMap` has tuple keys) — a second unrelated change that the new guardrail would not cover.
- **The RPC phase of `check_rpc_health.py` still hand-builds `"Cookie": auth.cookie_header`** (three call sites). That is `flat_cookies`, whose own docstring says domain-aware HTTP operations should use `cookie_jar` instead — and production does exactly that (`_kernel.py` passes `auth.cookie_jar` to the client). It is *not* the #2019 failure: those are `client.post(...)` calls with httpx's default `follow_redirects=False`, so there is no chain to dead-end, and the behaviour predates this PR. It is left alone because switching to `httpx.AsyncClient(cookies=auth.cookie_jar)` changes which cookies reach batchexecute — and given the open question above about a jar legitimately *withholding* a host-only cookie, that is precisely the kind of change that should not be made blind on a canary I cannot run live. Worth a follow-up so a request-shape drift monitor sends requests the way production sends them.

The last item was flagged in review; downgrading it from "fix now" to "named and deferred" is a deliberate call, not an oversight.

## Behaviour notes for review

- Under `NOTEBOOKLM_AUTH_JSON` (`path is None`) `from_storage`'s `save_cookies_to_storage` is a documented no-op, so the canary still writes nothing. With a real profile path (i.e. after #2037) it now persists rotated cookies back — the same thing every CLI invocation does, and desirable in-run.
- Also under env auth, `from_storage` reads in-band account metadata out of the storage state, whereas the old code passed `storage_path=None` to `read_account_metadata` and always got `{}` → `authuser=0`, `email=None`. Net effect: the script now honours a recorded `authuser` / account email instead of silently defaulting to `0`, matching the library. Today's secret has no `notebooklm.account` namespace, so CI behaviour is unchanged — but the routing rides on every batchexecute call the canary makes, so if a *stale* in-band record were ever added the canary would flip from "auth expired" to wrong-account `NOT_FOUND`. The guardrail asserts `auth.authuser == 0` / `auth.account_email is None` for a record-free storage state, and the new report line prints `account route` so a change is visible rather than silent.

## Verification — read this before trusting the checks

> ✅ **Resolved.** The earlier repo-wide `Code Quality` failure (10 stale `[removed-export] notebooklm.rpc.*` allowlist entries, which made `test.yml`'s `needs: quality` skip the entire matrix) was fixed by #2037's prune. Re-run locally on the rebased branch: `scripts/audit_public_api_compat.py --check-stale` → `OK: public API is compatible with v0.8.0 (0 reviewed break(s) allowlisted)`. The Test matrix now executes for real on this PR.

Verified locally against the repo venv (whole tree, not just the diff):

- `ruff check .` → clean; `ruff format --check .` → 952 files already formatted
- `mypy src/notebooklm --ignore-missing-imports` → clean, 314 source files
- full `pytest` → **12447 passed, 64 skipped, 1 xfailed** (re-run on the rebased branch)
- `tests/_guardrails` → green (incl. the ADR-0007 monkeypatch lint and the string-patch ratchet)

The guardrail was confirmed **red against the pre-fix scripts**, on the intended assertions, before being trusted.

### DIFF
--- a/.github/workflows/rpc-health.yml
+++ b/.github/workflows/rpc-health.yml
@@ -235,8 +235,12 @@ jobs:
     # so a drift hit files an issue rather than red-failing the canary outright.
     - name: Run bundle drift monitor
       id: bundle
-      # Reads auth from the profile materialized above (load_auth_from_storage
-      # falls back to the storage_state.json file when the env var is absent).
+      # Reads auth from the profile materialized above
+      # (build_httpx_cookies_from_storage falls back to the storage_state.json
+      # file when the env var is absent). It must stay the DOMAIN-PRESERVING
+      # loader: a flat name->value mapping handed to httpx.get(cookies=...)
+      # carries no domain, so every cookie is sent to every host in the
+      # redirect chain regardless of which auth source fed it — issue #2019.
       continue-on-error: true
       shell: bash
       run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   type axis. The MCP `/files/dl` route already resolved both format-aware and
   was unaffected
   ([#2034](https://github.com/teng-lin/notebooklm-py/issues/2034)).
+- **Nightly RPC-health and bundle-drift checks no longer report "Authentication
+  expired" against valid credentials.** `scripts/check_rpc_health.py` and
+  `scripts/capture_rpc_registry.py` loaded auth through the flat
+  `load_auth_from_storage()` mapping, which drops every cookie's domain. Under
+  `NOTEBOOKLM_AUTH_JSON` there is no storage file to recover them from, so
+  host-scoped cookies (`OSID` on the NotebookLM host, `LSID` on
+  `accounts.google.com`) were broadcast to every `*.google.com` host. Once
+  Google's `notebook.google.com` cutover landed, the stale broadcast `OSID`
+  collided with the freshly minted host cookie and dead-ended the sign-in chain
+  on `accounts.google.com/CookieMismatch`. Both scripts now use the
+  domain-preserving loaders the CLI already used
+  (`AuthTokens.from_storage()` / `build_httpx_cookies_from_storage()`), and a
+  unit guardrail pins that the jar mirrors the `storage_state` domains
+  ([#2019](https://github.com/teng-lin/notebooklm-py/issues/2019),
+  [#2018](https://github.com/teng-lin/notebooklm-py/issues/2018)).
 
 ## [0.8.0] - 2026-08-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,9 +51,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   on `accounts.google.com/CookieMismatch`. Both scripts now use the
   domain-preserving loaders the CLI already used
   (`AuthTokens.from_storage()` / `build_httpx_cookies_from_storage()`), and a
-  unit guardrail pins that the jar mirrors the `storage_state` domains
+  unit guardrail pins that the jar mirrors the `storage_state` domains and that
+  a host-scoped cookie value never reaches a host it was not scoped to
   ([#2019](https://github.com/teng-lin/notebooklm-py/issues/2019),
   [#2018](https://github.com/teng-lin/notebooklm-py/issues/2018)).
+
+### Changed
+
+- **`scripts/check_rpc_health.py` reports what it authenticated with.** The
+  report now names the auth source (`NOTEBOOKLM_AUTH_JSON` vs the resolved
+  profile path), the base host, the account route, and the cookie jar's
+  `name@domain` scopes — names and domains only, never values. #2019 was a
+  cookie-*scoping* failure that surfaced as a generic "authentication expired"
+  line, leaving the nightly log with nothing to diagnose it from
+  ([#2019](https://github.com/teng-lin/notebooklm-py/issues/2019)).
 
 ## [0.8.0] - 2026-08-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,17 +42,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Nightly RPC-health and bundle-drift checks no longer report "Authentication
   expired" against valid credentials.** `scripts/check_rpc_health.py` and
   `scripts/capture_rpc_registry.py` loaded auth through the flat
-  `load_auth_from_storage()` mapping, which drops every cookie's domain. Under
-  `NOTEBOOKLM_AUTH_JSON` there is no storage file to recover them from, so
+  `load_auth_from_storage()` mapping, which drops every cookie's domain, so
   host-scoped cookies (`OSID` on the NotebookLM host, `LSID` on
   `accounts.google.com`) were broadcast to every `*.google.com` host. Once
   Google's `notebook.google.com` cutover landed, the stale broadcast `OSID`
   collided with the freshly minted host cookie and dead-ended the sign-in chain
-  on `accounts.google.com/CookieMismatch`. Both scripts now use the
-  domain-preserving loaders the CLI already used
+  on `accounts.google.com/CookieMismatch`. The health check lost the domains
+  only under `NOTEBOOKLM_AUTH_JSON`, where there is no storage file to rebuild
+  the jar from; the drift monitor lost them unconditionally, because a plain
+  mapping handed to `httpx.get(cookies=...)` carries no domain at all. Both now
+  use the domain-preserving loaders the CLI already used
   (`AuthTokens.from_storage()` / `build_httpx_cookies_from_storage()`), and a
-  unit guardrail pins that the jar mirrors the `storage_state` domains and that
-  a host-scoped cookie value never reaches a host it was not scoped to
+  unit guardrail pins — for both the env-var and the profile-file
+  configuration — that the jar mirrors the `storage_state` domains and that a
+  host-scoped cookie value never reaches a host it was not scoped to
   ([#2019](https://github.com/teng-lin/notebooklm-py/issues/2019),
   [#2018](https://github.com/teng-lin/notebooklm-py/issues/2018)).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,15 +47,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `accounts.google.com`) were broadcast to every `*.google.com` host. Once
   Google's `notebook.google.com` cutover landed, the stale broadcast `OSID`
   collided with the freshly minted host cookie and dead-ended the sign-in chain
-  on `accounts.google.com/CookieMismatch`. The health check lost the domains
-  only under `NOTEBOOKLM_AUTH_JSON`, where there is no storage file to rebuild
-  the jar from; the drift monitor lost them unconditionally, because a plain
-  mapping handed to `httpx.get(cookies=...)` carries no domain at all. Both now
-  use the domain-preserving loaders the CLI already used
+  on `accounts.google.com/CookieMismatch`. Flattening also **lost** data
+  outright, which is the more serious half: a flat map has one slot per name,
+  so where `OSID` legitimately existed on both the app host and
+  `accounts.google.com` with different values, the duplicate-name resolution
+  silently discarded one of them before any request was built — the sign-in
+  host then received a value that was never its own. The health check lost the
+  domains only under `NOTEBOOKLM_AUTH_JSON`, where there is no storage file to
+  rebuild the jar from; the drift monitor lost them unconditionally, because a
+  plain mapping handed to `httpx.get(cookies=...)` carries no domain at all.
+  Both now use the domain-preserving loaders the CLI already used
   (`AuthTokens.from_storage()` / `build_httpx_cookies_from_storage()`), and a
   unit guardrail pins — for both the env-var and the profile-file
-  configuration — that the jar mirrors the `storage_state` domains and that a
-  host-scoped cookie value never reaches a host it was not scoped to
+  configuration — that the jar mirrors the `storage_state` domains, that a
+  host-scoped cookie value never reaches a host it was not scoped to, and that
+  the same-named sibling on the other host survives
   ([#2019](https://github.com/teng-lin/notebooklm-py/issues/2019),
   [#2018](https://github.com/teng-lin/notebooklm-py/issues/2018)).
 
--- a/scripts/capture_rpc_registry.py
+++ b/scripts/capture_rpc_registry.py
@@ -410,12 +410,12 @@ def fetch_bundle() -> str:
     import httpx
 
     from notebooklm._env import get_base_url
-    from notebooklm.auth import authuser_query, load_auth_from_storage
+    from notebooklm.auth import authuser_query, build_httpx_cookies_from_storage
 
     def _fetch(
         url: str,
         *,
-        cookies: dict[str, str] | None = None,
+        cookies: httpx.Cookies | None = None,
         follow_redirects: bool = False,
         timeout: float = 60.0,
     ) -> httpx.Response:
@@ -429,7 +429,13 @@ def fetch_bundle() -> str:
         response.raise_for_status()
         return response
 
-    cookies = load_auth_from_storage()
+    # Domain-preserving jar, NOT the flat ``load_auth_from_storage()`` mapping:
+    # a plain ``dict`` handed to ``httpx.get(cookies=...)`` carries no domain, so
+    # every cookie is sent to every host in the redirect chain. Host-scoped
+    # cookies (``OSID`` on the NotebookLM host, ``LSID`` on
+    # ``accounts.google.com``) leaking across hosts dead-ends the post-cutover
+    # sign-in chain on ``accounts.google.com/CookieMismatch`` — see issue #2019.
+    cookies = build_httpx_cookies_from_storage()
     html = _fetch(
         f"{get_base_url()}/?{authuser_query(0)}",
         cookies=cookies,
--- a/scripts/check_rpc_health.py
+++ b/scripts/check_rpc_health.py
@@ -61,7 +61,7 @@ from notebooklm._chat.wire import (
     build_streaming_chat_request,
     parse_streaming_chat_response,
 )
-from notebooklm._env import get_default_language
+from notebooklm._env import get_base_url, get_default_language
 from notebooklm._logging import scrub_secrets
 from notebooklm._notebooks import build_create_notebook_params
 from notebooklm.auth import AuthTokens
@@ -226,6 +226,25 @@ def extract_id(data: Any, *indices: int) -> str | None:
         return None
 
 
+def describe_auth_source(storage_path: Path | None) -> str:
+    """Name where credentials were loaded from, for diagnostics."""
+    return "NOTEBOOKLM_AUTH_JSON env var" if storage_path is None else str(storage_path)
+
+
+def describe_cookie_scopes(auth: AuthTokens) -> str:
+    """Summarise the jar as ``name@domain`` pairs — names and domains only.
+
+    #2019 was a cookie-*scoping* failure that produced a generic "authentication
+    expired" line, so the nightly log carried nothing to diagnose it with. Cookie
+    names and domains are not secrets (values are never printed), and having them
+    in the report makes the next scoping regression readable from the log alone.
+    """
+    jar = auth.cookie_jar
+    if jar is None:
+        return "<no jar>"
+    return ", ".join(sorted(f"{cookie.name}@{cookie.domain}" for cookie in jar.jar))
+
+
 async def load_auth(storage_path: Path | None) -> AuthTokens:
     """Load auth from environment or storage file, preserving cookie domains.
 
@@ -248,18 +267,32 @@ async def load_auth(storage_path: Path | None) -> AuthTokens:
     that collides with the freshly minted host cookie, dead-ends the sign-in
     redirect chain on ``accounts.google.com/CookieMismatch``, and made this
     script report "Authentication expired" against perfectly valid credentials.
+
+    Only the three failure classes below are mapped to the AUTH exit code.
+    ``RuntimeError`` (a failed ``NOTEBOOKLM_REFRESH_CMD``) and non-missing-file
+    ``OSError`` still propagate as a traceback, exactly as they did through the
+    old ``fetch_tokens`` call.
     """
     try:
         return await AuthTokens.from_storage(storage_path)
-    except FileNotFoundError:
+    except FileNotFoundError as e:
+        # ``e`` names the storage path that was actually checked — under
+        # profiles that is the only way to see which one the script resolved.
         print(
-            "ERROR: No authentication found.\n"
+            f"ERROR: No authentication found: {scrub_secrets(e)}\n"
             "Set NOTEBOOKLM_AUTH_JSON env var or run 'notebooklm login'",
             file=sys.stderr,
         )
         sys.exit(2)
     except ValueError as e:
-        print(f"ERROR: {scrub_secrets(e)}", file=sys.stderr)
+        # Name the auth source. The file branch of ``_load_storage_state``
+        # re-raises a bare ``json`` error carrying no context at all, so a
+        # corrupt storage_state.json would otherwise print only
+        # "Expecting value: line 1 column 1 (char 0)".
+        print(
+            f"ERROR: {scrub_secrets(e)}\nAuth source: {describe_auth_source(storage_path)}",
+            file=sys.stderr,
+        )
         sys.exit(2)
     except httpx.HTTPError as e:
         # ``httpx`` exception strings can echo full request URLs including
@@ -1301,9 +1334,13 @@ async def run_health_check(full_mode: bool = False) -> tuple[list[CheckResult], 
     temp_resources = TempResources()
     cohort_status = CohortStatus.UNKNOWN
 
-    print("Fetching auth tokens...")
-    auth = await load_auth(resolve_storage_path())
+    storage_path = resolve_storage_path()
+    print(f"Fetching auth tokens... (source: {describe_auth_source(storage_path)})")
+    auth = await load_auth(storage_path)
     print(f"Auth OK (CSRF token length: {len(auth.csrf_token)})")
+    print(f"  base host:     {httpx.URL(get_base_url()).host}")
+    print(f"  account route: {auth.account_route}")
+    print(f"  cookie scopes: {describe_cookie_scopes(auth)}")
     print()
 
     async with httpx.AsyncClient(timeout=30.0) as client:
--- a/scripts/check_rpc_health.py
+++ b/scripts/check_rpc_health.py
@@ -64,13 +64,7 @@ from notebooklm._chat.wire import (
 from notebooklm._env import get_default_language
 from notebooklm._logging import scrub_secrets
 from notebooklm._notebooks import build_create_notebook_params
-from notebooklm.auth import (
-    AuthTokens,
-    fetch_tokens,
-    get_account_email_for_storage,
-    get_authuser_for_storage,
-    load_auth_from_storage,
-)
+from notebooklm.auth import AuthTokens
 from notebooklm.exceptions import ChatError, ChatResponseParseError, DecodingError
 from notebooklm.paths import get_storage_path
 from notebooklm.rpc import (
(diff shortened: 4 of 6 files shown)


teng-lin:
@claude review

teng-lin:
### Heads-up for reviewers: the checks on this PR carry no test signal

`Code Quality` is failing **repo-wide**, not because of this PR:

```
Public API is compatible with v0.8.0 (0 reviewed break(s) allowlisted), but the allowlist has stale entries.
Stale allowlist entries — they match no break against the v0.8.0 baseline:
  - [removed-export] notebooklm.rpc.AuthError
  - [removed-export] notebooklm.rpc.ClientError
  … 10 entries total
Prune them from scripts/api-compat-allowlist.json (see docs/releasing.md → prune-allowlist-at-release).
```

The mandatory post-v0.8.0 allowlist prune was skipped at release. Because `test.yml` declares `needs: quality`, the entire 15-job **Test matrix reports `skipping`** — same on #2039, #2041, #2042 and on `main`.

**#2037 already carries the prune** (its diff empties `allowed_breaks`), so I have deliberately *not* touched `scripts/api-compat-allowlist.json` here — a second prune would conflict on exactly that file. I'll rebase and re-run once #2037 lands so this PR gets a real matrix result.

Until then, treat the local run as the evidence: whole-tree `ruff check .` + `ruff format --check .` clean, `mypy src/notebooklm --ignore-missing-imports` clean (314 files), full `pytest` **12444 passed / 64 skipped / 1 xfailed**, `tests/_guardrails` green. The new guardrail was confirmed **red against the pre-fix scripts** on its intended assertions before being trusted.

teng-lin:
Pushed `81c34dda` — a second review round found a real hole in the guardrail, so the test surface changed materially since @claude[bot] and CodeRabbit last looked. Re-review requested below.

**The hole:** the wire assertions keyed on cookie *names*. The contract is "the app-host `OSID` never reaches the sign-in host", but what was written was "no cookie named `OSID` ever reaches it". Those coincide only because the fixture had no accounts-scoped `OSID`. Add one — which is realistic, Google sets `OSID` on both hosts — and the old assertion **fails against correct behaviour**. Same-name-different-domain is the entire reason domains matter, and it was the one case the test could not express.

Changes in `81c34dda`:

- Assertions compare `(name, value)` pairs; the fixture carries `OSID` on two hosts with different values, plus `__Secure-OSID`. Flattening now visibly *destroys* the accounts-scoped value rather than merely duplicating a name.
- The mocked chain mints a fresh host-scoped `OSID` via `Set-Cookie` — the exact mechanism #2019 describes — so the jar assertions must survive a rotation instead of pinning a jar that never changes.
- New test for the **file-backed profile**, the configuration #2037 moves CI to. It takes a different `from_storage` branch and actually persists, so it asserts the rotated cookie reaches disk still scoped to the app host.
- `capture_rpc_registry` is now asserted **on the wire** (`httpx_mock`, real redirect chain) instead of by inspecting the argument, so a regression that builds a correct jar and then flattens it before use is caught too.
- Sign-in assertions filter on `/ServiceLogin`, so the autouse `accounts.google.com/RotateCookies` keepalive mock can't satisfy them.
- `authuser` / `account_email` are asserted, since `from_storage` reads in-band account metadata under env auth where the old code always got `{}`.

Also in this push:

- The `ValueError` arm had lost its context — the file branch of `_load_storage_state` re-raises a bare `json` error, so a corrupt `storage_state.json` printed only `Expecting value: line 1 column 1 (char 0)`. It now names the auth source; the `FileNotFoundError` arm keeps the path it checked.
- The report prints the auth source, base host, account route and the jar's `name@domain` scopes (**names and domains only, never values**). #2019 was a cookie-*scoping* failure that surfaced as a generic "authentication expired" line — the nightly log had nothing to diagnose it with.
- The `ValueError` and `httpx.HTTPError` arms of `load_auth` had **no coverage at all**; the `httpx` one carries the `scrub_secrets` call that keeps `f.sid=<session_id>` out of the Actions log and the auto-filed issue body. Both are pinned now.
- Dropped the `importlib.util.spec_from_file_location` preamble in `tests/unit/test_check_rpc_health.py`: it built a **second** module object for the same file, so a patch applied to one copy left the other untouched. The missing-storage test patches nothing now — an empty `NOTEBOOKLM_HOME` makes the real loader raise for itself.

The PR body has also been corrected on two points I'd overstated: an **open question for #2020** (a domain-preserving jar legitimately *withholds* a `notebook.google.com` host-only cookie from `notebooklm.google.com`, because `_ALLOWED_BASE_HOSTS` can never yield the rebrand host), and the RPC phase of this script still hand-building `"Cookie": auth.cookie_header`, which is now explicitly named and deferred rather than silently left.

Re-verified whole-tree: `ruff check .` + `ruff format --check .` clean, `mypy src/notebooklm` clean (314 files), full `pytest` **12447 passed / 64 skipped / 1 xfailed**. The guardrail was re-confirmed **red against the pre-fix scripts** on its intended assertions after every one of these changes.

@claude review

teng-lin:
Addressed in `12db8db7`: the file-backed profile test reached for `_cookie_entry(...)["name"]` where the literal `"OSID"` was meant. It now compares the persisted `(name, value, domain)` triples directly — and additionally asserts the accounts-scoped `OSID` survived the save untouched, since `save_cookies_to_storage` writes *deltas*: "the rotation landed" and "the sibling was not clobbered" are two different claims and only the first was covered.

Thanks for tracing the module-identity point independently — that one was a genuine latent bug, not cleanup, and worth the confirmation.

Re-verified after the change: `ruff check .` + `ruff format --check .` clean, `mypy src/notebooklm` clean (314 files), and the guardrail + `tests/_guardrails` suite green (1194 passed).

teng-lin:
@claude review — final state is `12db8db7` (the assertion cleanup you flagged). Nothing else changed since your `81c34dda` pass.

teng-lin:
**Open question to settle between this PR and #2020 — before both land, not after.**

This PR makes the scripts' cookie jar domain-preserving, which is correct. But it interacts with #2020 in a way neither PR resolves alone:

- `_env.py` restricts `_ALLOWED_BASE_HOSTS` to `notebooklm.google.com` / `notebooklm.cloud.google.com`, so `get_base_url()` can **never** return `notebook.google.com`. Requests still go to the legacy host.
- `REQUIRED_COOKIE_DOMAINS` in `_auth/cookie_policy.py` currently lists `.google.com`, `google.com`, `.notebooklm.google.com`, `notebooklm.google.com` — **not** the rebrand host. #2020 adds it.
- Once #2020 starts capturing a **host-only** `OSID` / `__Secure-OSID` on `notebook.google.com`, a correctly domain-scoped jar will **withhold** that cookie from requests to `notebooklm.google.com`. The old flat dict sent it — wrongly, by pinning everything to `.google.com`, which is the bug this PR fixes.

So the two changes are individually right and jointly untested. The question is whether the legacy host still has its own `OSID` after the rebrand, or whether Google now only mints one on `notebook.google.com`. If the latter, a domain-preserving jar could withhold the only binding cookie that exists.

**The failure mode is what makes this worth pre-empting:** it would surface as the identical `"Authentication expired or invalid"` string this issue was about, sending the next person down the same wrong path. That is precisely why the jar-scope diagnostics in this PR print `name@domain` (never values) — but a diagnostic only helps someone who already knows to look.

Not blocking this PR. Flagging so it is settled deliberately rather than discovered as a second false expiry report. cc #2020, #2038.

teng-lin:
**Resolved — the two changes compose safely. No action needed on either PR.**

I checked a live, in-use profile (written today, currently serving a running MCP deployment). Cookie names and domains only, no values:

```
.google.com             n=17   OSID-ish: []
accounts.google.com     n= 5   OSID-ish: []
notebook.google.com     n= 2   OSID, __Secure-OSID   (__Secure-OSID expires 2028-08-02)
notebooklm.google.com   n= 2   OSID, __Secure-OSID   (__Secure-OSID expires 2028-07-17)
```

**Both hosts carry their own `OSID` / `__Secure-OSID`.** So a correctly domain-scoped jar targeting `notebooklm.google.com` sends that host's own binding cookie, and withholds the `notebook.google.com` one — which is the correct behaviour, not a loss. The concern that #2020's rebrand-host cookie would be the *only* one available does not hold.

The expiry dates are informative: the legacy-host `__Secure-OSID` was minted ~2026-07-17 and the rebrand-host one ~2026-08-02. The cutover minted a new host-scoped cookie alongside the existing one rather than replacing it.

**Residual unknown, worth stating rather than hiding:** this is one profile, and its legacy-host cookie predates the cutover. A brand-new login performed today might mint only on `notebook.google.com`. Closing that requires a fresh interactive login into an isolated storage path (`notebooklm login --storage <tmp>`) and inspecting the resulting domains — cheap, but it needs a human at a browser. Until someone does it, treat "a fresh post-rebrand login still yields a legacy-host OSID" as unverified.

That residual does not block either PR: if it ever bites, the symptom is a bare `OSID` absence, and #2044's jar-scope diagnostics now print `name@domain` precisely so it is visible instead of surfacing as a bogus "Authentication expired".

teng-lin:
Rebased onto `27926aa4` (#2037 merged) — head is now `1ce159e8`. Clean rebase, no conflict: #2037 added no `[Unreleased]` CHANGELOG entry. This branch touches **zero** workflow files.

**Code Quality is unblocked.** `scripts/audit_public_api_compat.py --check-stale` now returns `OK: public API is compatible with v0.8.0 (0 reviewed break(s) allowlisted)` on the rebased branch, so `test.yml`'s `needs: quality` gate opens and the matrix executes for real. The "no test signal" caveat in the PR body has been replaced accordingly.

### Re-verified the #2037 interaction against the *merged* code, not the diff

**1. How the script now receives credentials.** `rpc-health.yml` drops `NOTEBOOKLM_AUTH_JSON` from the run step's `env:` and writes the secret to `~/.notebooklm/profiles/default/storage_state.json`. So `resolve_storage_path()` returns a real `Path` in CI instead of `None`. That changes the *value* handed to `AuthTokens.from_storage(storage_path)` — not the call site or its shape. No code change was needed. I checked the layouts genuinely agree, since a mismatch would surface as a silent `FileNotFoundError`:

```
NOTEBOOKLM_HOME    : /tmp/tmp_2e4avxp
get_storage_path() : /tmp/tmp_2e4avxp/profiles/default/storage_state.json
#2037 writes       : profiles/default/storage_state.json          → MATCH
```

**2. My earlier "masks one half but not the other" finding holds — and the drift half is worse than I described.** Probed against the merged file-backed configuration:

```
PRE-FIX (main today), flat dict → httpx.get(cookies=…), FILE-backed profile:
  -> notebooklm.google.com   SID=…; __Secure-1PSIDTS=…; OSID=v-osid-app; LSID=v-lsid
  -> accounts.google.com     SID=…; __Secure-1PSIDTS=…; OSID=v-osid-app; LSID=v-lsid   ← app OSID broadcast
POST-FIX, domain-preserving jar, same file:
  -> notebooklm.google.com   SID=…; __Secure-1PSIDTS=…; OSID=v-osid-app
  -> accounts.google.com     SID=…; __Secure-1PSIDTS=…; OSID=v-osid-accounts; LSID=v-lsid
```

Look at the second row pre-fix: the app-host `OSID` is broadcast to the sign-in host **and** the accounts-scoped `OSID` is destroyed outright — `flatten_cookie_map` resolves the duplicate name by domain priority and the app host wins. `LSID` leaks the other way too. **On `main` as it stands the bundle-drift monitor is still broken, so #2018 will keep firing until this lands.** The health-check half is genuinely masked by #2037's file-backed profile, so for #2019 this is defense-in-depth plus regression protection — the flattening stays live on every env-var auth path (local runs, forks, anyone exporting the secret).

**3. The guardrail exercises the path the merged workflows take.** `test_check_rpc_health_auth_preserves_domains_for_file_backed_profile` — added in the review round, before #2037 landed — drives `resolve_storage_path()` under a temp `NOTEBOOKLM_HOME` and lands on exactly the `profiles/default/storage_state.json` layout above, including the cookie-persistence behaviour that only exists on the file-backed branch. The env-var test still covers local dev and forks. Both were re-confirmed **red against the pre-fix scripts** after the rebase.

`1ce159e8` also sharpens the CHANGELOG to state the scope split correctly: the health check lost domains only under `NOTEBOOKLM_AUTH_JSON`; the drift monitor lost them unconditionally.

The **#2020 cookie-scope question remains open and unanswered** — still stated as an open question in the PR body, not acted on.

Re-verified on the rebased branch: `ruff check .` + `ruff format --check .` clean, `mypy src/notebooklm` clean (314 files), full `pytest` **12447 passed / 64 skipped / 1 xfailed**, API-compat audit clean.

@claude review

teng-lin:
`580b1428` — both review points from the post-rebase round.

**CodeRabbit's duplication nitpick (taken).** The drift-monitor test re-implemented the three-response redirect chain inline. Both scripts follow redirects on their authenticated read and walk the *same* chain — only the final body differs (WIZ tokens vs a bundle URL) — so `_register_sign_in_chain` now takes `final_html` and both tests share it. Side benefit: the drift test now also exercises the `Set-Cookie` mint it previously skipped, so it covers the rotation hop too.

**@claude[bot]'s stale-comment observation (taken, narrowly).** The bundle-drift step in `rpc-health.yml` documented `load_auth_from_storage` as its loader — a comment *this PR invalidates* by replacing that call. Rather than leave knowingly-wrong documentation in a workflow, I re-pointed it at `build_httpx_cookies_from_storage` and recorded **why** it must stay the domain-preserving one:

```yaml
# Reads auth from the profile materialized above
# (build_httpx_cookies_from_storage falls back to the storage_state.json
# file when the env var is absent). It must stay the DOMAIN-PRESERVING
# loader: a flat name->value mapping handed to httpx.get(cookies=...)
# carries no domain, so every cookie is sent to every host in the
# redirect chain regardless of which auth source fed it — issue #2019.
```

This is the one place where the "touches zero workflow files" claim in the PR body no longer holds — **comment only, no workflow logic**, YAML re-validated. I checked no open PR touches `rpc-health.yml`, so there's no conflict surface. Happy to drop this hunk if you'd rather keep workflow edits strictly inside #2037's lineage.

Re-verified: `ruff check .` clean, `mypy src/notebooklm` clean (314 files), guardrail + `tests/_guardrails` green, and the guardrail re-confirmed **red against the pre-fix scripts** after the refactor (all three tests fail, on their intended assertions).

**Matrix result on the rebased branch — the point of the #2037 wait:** all **15/15** Test jobs pass (ubuntu/macos/windows × Python 3.10–3.14), plus Code Quality, CodeQL, and Analyze. That is the first real test signal this PR has had.

teng-lin:
@claude review — final state is `580b1428` (the shared sign-in-chain refactor plus the one workflow-comment fix). All 19 checks now pass, including the full 15-job Test matrix.

teng-lin:
`38882a2a` (CHANGELOG only) — sharpening the framing of what the flattening actually does, because "broadcast to every host" undersells it.

**Flattening is data loss, not just an over-broad send.** A flat `name -> value` map has exactly one slot per name. `OSID` legitimately exists on *both* the app host and `accounts.google.com` with *different* values, so `flatten_cookie_map` resolves the collision by domain priority — the app host wins — and the accounts-scoped `OSID` is **silently discarded before any request is built**. The sign-in host then receives a value that was never its own.

Re-reading the pre-fix wire trace with that in mind, there are **three** distinct failures in it, not one:

1. the app-host `OSID=v-osid-app` is **sent to the sign-in host** — the collision that dead-ends the chain;
2. the accounts-scoped `OSID=v-osid-accounts` is **gone**, destroyed before the request existed;
3. `LSID=v-lsid`, scoped to `accounts.google.com`, **leaks onto the app host**.

Only (1) is an over-broad send. (2) is data loss and (3) is a credential crossing a host boundary it was scoped away from. **The drift monitor still does all three on `main` today**, and no amount of workflow-side auth hardening reaches it — the flattening happens *after* credentials are loaded, regardless of where they were loaded from. So for #2018 this is the fix unconditionally, independent of #2037.

Worth noting the guardrail already pins this half — it wasn't only checking absence. Because the assertions are on `(name, value)` pairs, the sign-in hop asserts `("OSID", "v-osid-accounts")` **must be present**, which fails the moment flattening discards it; and the file-backed test makes the same claim against the persisted `storage_state` via `("OSID", "v-osid-accounts", accounts.google.com) in persisted_triples`. Confirmed at `tests/unit/test_scripts_auth_cookie_domains.py:193` and `:285`.

No code change — `git show --name-only 38882a2a` is `CHANGELOG.md` alone. The PR body's summary and root-cause sections carry the same correction.

Also confirming the `rpc-health.yml` comment hunk **stays** per maintainer review: it is comment-only, and the old text named `load_auth_from_storage` — pointing the next reader at the exact function that caused #2019.

Issue #2057, opened 2026-08-04
auth: PSIDTS recovery gate ranks a domain-blind global winner while the heal it gates is RFC-routed


teng-lin:
## Summary

`_psidts_needs_recovery` decides whether to fire the `RotateCookies` heal from a **domain-blind global winner** — the same flat projection tracked in #2054 — while the POST it gates is **RFC 6265 routed**. The gate and the action can therefore answer different questions about the same cookie set. The docstring that justifies the ranking is false.

No divergence is observed on the profiles measured today (see [Severity](#severity)). This is a latent correctness defect, not an outage.

## Mechanism

`_index_recovery_cookies` (`src/notebooklm/_auth/psidts_recovery.py:147-163`) builds a `name -> expiry` index by keeping the entry with the highest `_auth_domain_priority` tier, first-in-tier wins:

```python
priority = _auth_domain_priority(entry.get("domain", "") or "")
if name not in cookie_names or priority > name_priority[name]:
    cookie_names.add(name)
    cookie_expiry[name] = entry.get("expires")
    name_priority[name] = priority
```

`_psidts_needs_recovery` (`:112-123`) then gates recovery on *that one winner's* `expires`.

The tiers in `_auth_domain_priority` (`src/notebooklm/_auth/cookie_policy.py:534-563`) are **not** distinct:

| tier | domains |
|---|---|
| 4 | `.google.com` |
| 3 | `.notebooklm.google.com`, `.notebook.google.com`, `.notebooklm.cloud.google.com` |
| 2 | `notebooklm.google.com`, `notebook.google.com`, `notebooklm.cloud.google.com` |
| 1 | every regional ccTLD (`_is_google_domain`) |
| 0 | everything else allowlisted — **including `accounts.google.com`**, `drive.google.com`, `.googleusercontent.com`, `lh3.google.com` |

Within a tier the first entry in `storage_state` iteration order wins, so the winner is arbitrary and changes if the file is reordered.

Two things make this worse than a plain tie:

[...]

Issue #2054, opened 2026-08-04
auth: model cookies as (name, domain, path) end-to-end; delete the flat name->value map


teng-lin:
## Problem

Cookies are represented in two incompatible ways in this codebase: a domain-preserving jar, and a flat `name -> value` dict (`load_auth_from_storage()` → `_auth/tokens.py:247`, normalized through `flatten_cookie_map`).

The flat form **cannot represent the domain**, which makes a whole class of bug not merely possible but unavoidable wherever it is used. `#2019` / `#2018` were exactly that bug, and the damage was worse than "cookies sent too broadly":

`OSID` legitimately exists on the app host **and** on `accounts.google.com` with *different values*. A flat map has one slot per name, so `flatten_cookie_map`'s duplicate-name resolution **discarded one outright, before any request was built**. Three distinct failures from one modeling choice:

1. app-host `OSID` **sent to** the sign-in host — the collision that dead-ended the redirect chain on `/CookieMismatch`;
2. accounts-scoped `OSID` **destroyed** — the sign-in host received a value that was never its own;
3. `LSID` **leaked** the opposite direction, onto the app host.

Fixed at the two call sites in #2044, and a guardrail now pins domain preservation for those scripts — but **the flat type still exists and is still reachable.** `scripts/diagnose_get_notebook.py` carries the same shape today (deliberately deferred there, as a manual one-off), and nothing prevents the next caller from picking it up.

## Why a guardrail isn't enough

The #2044 guardrail asserts the *scripts'* jar preserves domains. It does not stop a new call site from using the flat loader. This is a case where the type should make the bug unwriteable rather than a test catching it after the fact — the cost of getting it wrong is silent credential misrouting that surfaces as a bogus "Authentication expired" (see #2038).

## Suggested direction

1. Make `(name, domain, path)` the only representation across `_auth/` — auth extraction, refresh, download paths, and scripts.
2. Remove `load_auth_from_storage()`'s flat return, or narrow it to a genuinely name-only consumer that cannot build a request (and say so in its name).
3. Delete `flatten_cookie_map`, or restrict it behind a type that cannot reach `httpx`.
4. Migrate `scripts/diagnose_get_notebook.py`.
5. Consider making a bare `dict[str, str]` unassignable to anything that feeds `httpx.Client(cookies=...)`, so the mistake fails at type-check time.

## Relevance to the live migration case

[...]

PR #2060, opened 2026-08-04
docs(auth): correct cookie-domain ranking documented as deterministic (#2054)

teng-lin:
Comments and tests only. **No behaviour change** — every added line under `src/` is a comment.

## Scope reduced after #2059 landed

This PR originally corrected three docstrings and one comment. **#2059 (implementing #2057) has since landed and fixed three of the four**, in equivalent or better form — `_auth_domain_priority`'s docstring, `flatten_cookie_map`'s, and the `psidts_recovery` index. Rather than force a stale branch through, I reset to `main` and reduced this to what is *still* wrong.

## What remains

**The tier comment added by #2020** (`_auth/cookie_policy.py`) still reads:

> …both sit below `.notebooklm.google.com` so a cookie on the canonical app host still wins when both hosts carry it.

It does not. `.notebooklm.google.com` and `.notebook.google.com` both return **3**; their bare variants both return **2**. Those pairs tie, so the winner falls to `storage_state` iteration order — and that is the state a live post-rebrand profile is actually in, carrying different `OSID` values on each host. The docstring directly above it now describes the real structure, so the file currently contradicts itself.

**Two misleading tests:**

- `test_priority_strict_ordering` asserted *"no ties between named tiers"* over a five-domain sample containing **exactly one domain per tier** — true of the sample, false as the guarantee its name claimed, and passing today while four tiers are shared. Renamed to `test_priority_ordering_within_this_sample`.
- `test_gemini_notebook_rebrand_host_matches_app_host_tiers` justified the tie as stopping either host "silently starving the other". Measurement shows the tie *is* the ambiguity; #2057 resolved it by removing the ranking from the PSIDTS gate, not by re-ordering tiers.

**Added** `test_named_tiers_are_shared_not_unique`, pinning the real structure — including that tier 0, the catch-all, holds `accounts.google.com`, the host the PSIDTS rotation POST actually targets, ranked *below* hosts it never reaches.

## Verification

Full suite 12806 passed / 64 skipped / 1 xfailed; mypy clean on 322 files; ruff and format clean whole-tree. Verified mechanically that every added line under `src/` is a comment, so this cannot change behaviour.

Refs #2054, #2057, #2020

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected documentation of cookie-domain ranking behavior when multiple domains share the same priority.
  * Clarified that tied domains are resolved according to cookie storage order, while URL-specific selection follows cookie-jar routing.

* **Tests**
  * Expanded coverage for shared priority tiers across NotebookLM, Notebook, and Google product domains.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Corrected the last cookie-domain tier claim that still said the ranking is
+  decisive.** #2057 rewrote the `_auth_domain_priority` docstring and its two
+  consumers, but the tier comment added by
+  [#2020](https://github.com/teng-lin/notebooklm-py/pull/2020) still said the
+  canonical app host "still wins when both hosts carry" a cookie name. It does
+  not: `.notebooklm.google.com` and `.notebook.google.com` are both tier 3, and
+  their bare variants both tier 2, so those pairs tie and the winner falls to
+  `storage_state` iteration order. Also renames `test_priority_strict_ordering`
+  — which asserted "no ties between named tiers" over a sample holding exactly
+  one domain per tier, and so passed while four tiers were shared — and adds a
+  test pinning the real tie structure. Comments and tests only; no behaviour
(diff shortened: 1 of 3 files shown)


teng-lin:
@claude review

PR #2085, opened 2026-08-05
fix(migration): make legacy-layout migration crash-safe via temp+rename

teng-lin:
## Summary

An interrupted legacy-layout migration could permanently destroy the user's only copy of `storage_state.json`. `shutil.copy2` interrupted mid-copy leaves a partial destination whose mtime is *newer* than the source (copystat never ran), so the retry's skip-if-newer guard kept the truncated file — and the migration then unconditionally deleted the intact legacy originals. Interrupted `copytree` had the same hole via the bare `dst.exists()` skip.

This change makes every migration copy land under a dot-prefixed temp name (`.<name>.migrating`) and renames it into place atomically, so a path at its *final* name is always a complete copy and both skip guards become trustworthy. Found during a cookie/auth-subsystem audit.

## Related Issue

N/A — found by internal audit, no tracking issue.

## Changes

- `src/notebooklm/migration.py`: file copies go through temp + `replace_file_atomically`; directory copies go through temp + `os.replace`; stale temps from prior interrupted runs are cleaned before each retry; originals are still only deleted after the copy loops complete.
- Migrated `storage_state.json` is tightened to `0o600` on POSIX instead of inheriting a loose legacy mode (e.g. `0644`); non-secret files keep the legacy mode as before. Windows behavior unchanged.
- Module docstring updated to state the temp+rename guarantee it now actually provides.
- `tests/unit/test_migration.py`: three regression tests — stale partial temp file (newer mtime) is discarded and the intact legacy content wins; stale partial temp dir is discarded and the full copytree completes; loose 0644 legacy mode is tightened to 0600 on the migrated copy (POSIX-only).

## Test Plan

## Notes

- `tests/unit/test_migration.py` is 18/18 green (15 prior + 3 new); the full unit suite at this commit passes (11647 passed, 85 skipped, 1 xfailed). No ADR needed — no architectural shape change.
- Deliberate behavior retention: the "profile copy is same age or newer" skip (a post-migration login writing fresh cookies to the legacy root must not be clobbered) is preserved — `copy2` + rename keep the source mtime, so the `>=` comparison still works across retries.
- A companion fix for a separate audit finding (write-time cookie-domain filtering on the browser-cookie login path) is in progress and will land on this branch as a follow-up commit unless a separate branch/PR is preferred.

https://claude.ai/code/session_017RMas6ZjhYgUQtqS7zgw2y

---
_Generated by [Claude Code](https://claude.ai/code/session_017RMas6ZjhYgUQtqS7zgw2y)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved migration reliability by preventing incomplete files or directories after interruptions.
  - Automatically cleans up leftover temporary migration data and safely retries interrupted operations.
  - Secured migrated `storage_state.json` files with restricted permissions on supported systems.
  - Preserved original permissions for other migrated files.

### DIFF
--- a/src/notebooklm/migration.py
+++ b/src/notebooklm/migration.py
@@ -6,7 +6,9 @@ Handles transparent migration of ~/.notebooklm/ files into
 The migration is:
 - Automatic: triggered by ensure_profiles_dir() on CLI startup
 - Idempotent: safe to run multiple times
-- Crash-safe: uses copy-then-delete with marker file
+- Crash-safe: uses copy-then-delete with marker file; each copy lands via
+  temp-name + atomic rename, so a crash mid-copy can never leave a torn
+  file/dir at its final destination name
 - Non-destructive: originals kept until all copies succeed
 - Single-writer: cross-process ``filelock`` serializes concurrent CLI
   invocations so a container start-up race cannot interleave copies.
@@ -14,6 +16,7 @@ The migration is:
 
 import json
 import logging
+import os
 import shutil
 import sys
 from pathlib import Path
@@ -21,7 +24,7 @@ from typing import Any
 
 from filelock import FileLock, Timeout
 
-from ._atomic_io import atomic_update_json
+from ._atomic_io import atomic_update_json, replace_file_atomically
 from .paths import get_config_path, get_home_dir
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


teng-lin:
@claude review
@codex review

teng-lin:
Re: the claude[bot] review (https://github.com/teng-lin/notebooklm-py/pull/2085#issuecomment-5188700162) — thanks for the trace-through. Responding to the three non-blocking observations, all deliberately left as-is for this PR:

1. **Directory replace without the Windows retry loop**: correct — `os.replace(tmp, dst)` for directories does not get `replace_file_atomically`'s transient-`PermissionError` retries. The migration runs under a single-writer `filelock`, the directory branch only executes when `dst` does not exist (no destination contention to collide with), and the prior code had no retry either. Noted as a future hardening rather than expanding this fix's scope.

2. **`rmtree(ignore_errors=True)` leaving debris → `copytree` raising `FileExistsError`**: acknowledged. That failure mode is loud (the migration raises and the intact legacy originals are preserved, since deletion only happens after all copies complete), which is the safe outcome for this crash-safety fix — no silent data loss is possible. Making that retry path cleaner is orthogonal to the torn-copy bug being fixed.

3. **No retroactive chmod of previously-migrated profiles**: intentionally out of scope, as the review notes; the tightening applies to copies this code creates.

CodeRabbit posted no actionable comments and Codex found no issues, so no code changes were needed for review feedback. The only CI failure (windows-latest / Python 3.12) was `tests/unit/test_atomic_update_json.py::test_many_concurrent_increments`, a pre-existing concurrency stress test untouched by this PR that passed on the other four Windows jobs — re-ran the failed job.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2088, opened 2026-08-05
refactor(download): derive adapters from one registry

### LINKED ISSUE #2056: architecture: derive the CLI/MCP/REST download registries from one table instead of triplicating them
## Problem

The download type/format table is declared **three times**, once per adapter:

- `src/notebooklm/cli/_download_specs.py`
- `src/notebooklm/mcp/tools/_studio_download.py`
- `src/notebooklm/server/routes/artifacts.py`

Each independently states the extension, MIME type, and format axis for every artifact type. Until #2041, **nothing enforced that they agree.**

This is the one place the otherwise-clean `cli/ → _app/ → client → rpc/` layering breaks down. The bottom of the stack is properly shared — `_app/download.py` owns the plan-building and now the extension→MIME table — but the top re-declares what should be derived from it.

## Evidence that the duplication is load-bearing

`#2034`: audio overviews were served as `.mp3` when the bytes are AAC/M4A. The wrong extension was present in **all three** registries, and had to be fixed in all three. A one-surface fix would have shipped a half-fix that looked complete.

The parity guardrail added in #2041 (`tests/_guardrails/test_download_spec_parity.py`) earned its keep immediately by catching a genuine divergence — the CLI's slide-deck row uses the legacy Click param name `format_param_name="slide_format"` while MCP/REST use `output_format`. That one is adapter-local and legitimately excluded, which is exactly the point: the registries differ in ways that *are* meaningful and ways that *aren't*, and today nothing distinguishes them structurally.

## Why a parity test is a patch, not a fix

The guardrail proves the three copies match. It does not stop a fourth adapter from being added with a fourth copy, and it does not stop someone editing one row and updating the test to match. The invariant "these are the same table" is better expressed by there being one table.

## Suggested direction

1. Promote a single registry into `_app/` — types, extensions, MIME, format axis, and which format values are legal per type.
2. Have each adapter derive its surface from it, contributing only genuinely adapter-local fields (Click param names, MCP schema text, REST route wiring).
3. Keep the parity guardrail only for the adapter-local residue, or delete it once there is nothing left to compare.

## Watch-outs

- MCP is subject to the ADR-0025 schema-character budget; derivation must not inflate tool schemas.
- `_app/download.py` is ~961 lines against the ADR-0008 1000-line cap, so the shared registry likely wants its own module rather than being folded in.
- The CLI's legacy `slide_format` param name is a public surface — the API-compat allowlist gate will catch a rename, and it should not be renamed casually.

## Acceptance criteria

- Adding a new artifact type or format requires editing exactly one table.
- No adapter restates an extension or MIME type.
- A new adapter cannot silently disagree with the others.

teng-lin:
## Summary
- add one immutable, transport-neutral download registry that keeps artifact behavior, extensions, and MIME types together
- derive CLI behavior and MCP schemas from the registry while preserving the legacy Click `slide_format` parameter
- make MCP and REST consume the exact shared projection, with guardrails for adapter identity, schema enums, MIME coverage, and one-row format additions

## Task
Closes #2056

## Test plan
- [x] `uv run ruff check .`
- [x] `uv run ruff format --check <changed Python files>`
- [x] `uv run mypy src/notebooklm`
- [x] focused CLI/MCP/REST/guardrail suite — 309 passed
- [x] `uv run pytest` — 13,122 passed, 77 skipped, 1 xfailed
- [x] `uv run pre-commit run --all-files`
- [x] post-rebase focused suite, Ruff, mypy, and pre-commit

## Deferred polish findings
- Retained `from .download_specs import FORMAT_EXTENSIONS as FORMAT_EXTENSIONS`: the explicit self-alias marks the compatibility symbol as an intentional public re-export for Ruff.

## Summary by CodeRabbit

* **New Features**
  * Added a shared registry for downloadable artifact types, formats, file extensions, and MIME types.
  * CLI, MCP, and REST downloads now use consistent supported types and format options.
  * MCP download schemas now dynamically advertise valid artifact types and formats.

* **Bug Fixes**
  * Improved validation for unsupported formats, conflicting mappings, and invalid file extensions.
  * Preserved legacy CLI slide-deck format compatibility.


teng-lin:
@claude review

teng-lin:
@codex review

PR #2094, opened 2026-08-06
feat(artifacts): add support for listing and downloading CUSTOM artifact types (type 10)

amaxcz:
Adds support for listing and downloading custom artifact types (type 10) in NotebookLM.

## Summary by CodeRabbit

- **New Features**
  - Added support for custom artifacts, including recognition, listing, filtering, and display.
  - Added custom artifact downloads with binary output and automatic handling of available content or URLs.
  - Added custom artifact metadata, generation prompts, and URL access.
  - Added the `custom` studio listing filter.

### DIFF
--- a/src/notebooklm/_app/download_specs.py
+++ b/src/notebooklm/_app/download_specs.py
@@ -202,6 +202,13 @@ DOWNLOAD_REGISTRY: tuple[DownloadRegistryEntry, ...] = (
         download_attr="download_report",
     ),
     DownloadRegistryEntry(
+        name="custom",
+        kind=ArtifactType.CUSTOM,
+        default_output=DownloadFormatSpec(".bin", "application/octet-stream"),
+        default_dir="./custom",
+        download_attr="download_custom",
+    ),
+    DownloadRegistryEntry(
         name="mind-map",
         kind=ArtifactType.MIND_MAP,
         default_output=DownloadFormatSpec(".json", "application/json"),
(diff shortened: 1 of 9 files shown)


teng-lin:
Thanks for the contribution, but closing this without merging.

The core problem is that there's no evidence Google's NotebookLM actually uses type code 10 as a "custom" artifact type. The RPC type codes in this project are all reverse-engineered from observed network traffic, and every mapped code (0–9) is backed by real captured payloads. This PR instead speculates about an unobserved code, and the implementation shows it:

- `_CUSTOM_METADATA_POS = 24` is a guessed row position with no captured payload confirming that shape.
- `artifact_url()` for CUSTOM falls back through `custom_url or infographic_url or slide_deck_pdf_url or audio_url or video_url` — trying every other type's URL slot in turn is a sign we don't know what the payload actually looks like.
- The download path defaults to writing whatever it finds as `.bin` / `application/octet-stream`, which isn't a meaningful user-facing artifact.

Mapping an unverified code is worse than leaving it unmapped: today, an unrecognized type code safely surfaces as `UNKNOWN`, whereas this change would confidently misclassify whatever Google eventually ships (or repurposes) as type 10, and mis-parse its payload. Given that RPC shape drift is already the #1 breakage class for this project, we only add type mappings backed by captured traffic showing the real payload structure.

If you have actual captured `batchexecute` responses showing type 10 artifacts in the wild (what feature produces them, and the real row shape), please open an issue with those captures and we can revisit.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2099, opened 2026-08-06
test(guardrails): pin the drained ADR-0007 baselines with stays-empty guards

teng-lin:
## Problem

ADR-0007's monkeypatch policy is enforced by two gates with **four** exemption structures between them. All four have drained to zero:

| Structure | Gate | Pinned at empty? |
|---|---|---|
| `_ALLOWLIST` (patterns a–e, h) | `test_no_forbidden_monkeypatches.py` | ✅ `test_allowlist_stays_empty` |
| `_DEEP_LEAF_ALLOWLIST` (pattern f) | `test_no_forbidden_monkeypatches.py` | ❌ |
| `_PATCH_OBJECT_PRIVATE_ATTR_ALLOWLIST` (pattern g) | `test_no_forbidden_monkeypatches.py` | ❌ |
| `STRING_PATCH_CEILINGS` | `test_string_patch_ratchet.py` | ❌ |

Only the first was locked in. The other three are **not self-locking**, so a contributor blocked by one of these gates could re-baseline their own file in the same PR and keep CI green:

- `_assert_tier_clean` *skips* every file listed in its allowlist and only reports entries that have **stopped** offending (the stale check). A genuinely-offending file therefore satisfies both halves the moment it is re-added.
- A ceiling re-added at a file's current site count reads as "not grown" (`count == ceiling`), "not slack" (`count == ceiling`), and "not stale" (path exists) — all three ratchet checks pass simultaneously.

That silently re-opens the escape hatch the drain closed. I confirmed it by driving the real gate functions against a synthetic offender:

```
_DEEP_LEAF_ALLOWLIST                   empty (today)     -> fails
_DEEP_LEAF_ALLOWLIST                   entry RE-ADDED    -> PASSES   ← green CI
_PATCH_OBJECT_PRIVATE_ATTR_ALLOWLIST   entry RE-ADDED    -> PASSES
STRING_PATCH_CEILINGS                  ceiling RE-ADDED  -> PASSES
```

## Change

**1. Two new guards**, mirroring the existing `_ALLOWLIST == frozenset()` assertion:

- `test_baselined_allowlists_stay_empty` — pins both pattern (f)/(g) allowlists. Checks the immutable **type** as well as the empty value, so a refactor to a mutable `set()` cannot slip past a mere falsiness check (same reasoning the `_ALLOWLIST` guard already documents).
- `test_ceilings_stay_empty` — pins `STRING_PATCH_CEILINGS` at `{}`.

**2. Prose corrections.** Five places described state that no longer exists:
- the two baselines were documented as *"start populated and drain opportunistically — there is no stays-empty guard"*
- pattern (f)'s comment called `_DEEP_LEAF_ALLOWLIST` *"baselined, drains opportunistically"*
- the module header said matches are checked against *"the shrinking allowlist"*
- the ratchet advertised a *"2026-06-10 baseline: 52 files, 768 sites"* directly above an empty dict
- ADR-0007 lines 13/62/63 carried the same claims

Anyone reading these would conclude debt remained. ADR-0007's `_ALLOWLIST` paragraphs were updated when that list drained; these were missed.

## Why now

The drain is finished — this is not an eager burndown. It converts "drained" into "cannot re-grow" at zero ongoing cost, per the repo's enforce-don't-document principle: ship the fix *with* its gate, and don't leave reclaimed ground un-pinned.

Worth noting the string-patch ratchet is no longer really a ratchet: with the map empty, the unbaselined-file budget of zero applies to every file, so it has become a flat ban on string-target `patch("notebooklm…")` anywhere under `tests/`. The docs now say so.

## Verification

- **Mutation-tested** — each guard was confirmed to *fail* when its invariant is broken (re-add a deep-leaf entry; swap `frozenset()`→`set()`; re-add a ceiling), then restored to green. A guard that cannot fail is worthless, and this file's own philosophy is to self-test against vacuity.
- Independently re-measured all 8 forbidden patterns + the ratchet pattern across the 567 scanned test files: **0 live sites**. The only hits anywhere are self-test string literals in `tests/_guardrails/` and docstring prose in `tests/_fixtures/` — both deliberately skipped subtrees, and I verified those hits are comments, not code.
- `ruff check` + `ruff format --check` clean on changed files
- `mypy src/notebooklm --ignore-missing-imports` — 332 files, no issues
- `tests/_guardrails/` — 1194 passed, 1 xfailed

No behavior change to the gates today; every population is already at zero.

### Unrelated pre-existing issue

`ruff format --check .` over the whole tree flags `tests/integration/mcp_vcr/test_notebooks.py`, which is **not** in this diff — it is pre-existing on `main`, most likely the known `uv.lock` vs `pyproject` ruff pin divergence. Left alone rather than folded into this change.

## Summary by CodeRabbit

* **Tests**
  * Strengthened safeguards to ensure prohibited patching patterns remain fully eliminated.
  * Added checks preventing previously cleared exception lists and patching budgets from being repopulated.
  * Updated enforcement rules to require approved dependency-injection or public-interface approaches for future changes.

* **Documentation**
  * Clarified the policy that all patching exemptions are globally enforced and must remain empty.


teng-lin:
@claude review

Issue #2103, opened 2026-08-06
login --master-token ignores --storage for master_token.json, breaking the sibling invariant


teng-lin:
## Summary

`notebooklm login --master-token --storage <path>` resolves `storage_state.json` and `master_token.json` **independently**. The storage path honors `--storage`; the master-token path does not — it is always the active profile dir. Everything at or below the public line *derives* the token path from the storage path as a sibling, so under `--storage` the CLI writes the token to a location no reader ever looks at.

`--storage` is a documented option on `login` (`src/notebooklm/cli/session_cmd.py:169-173`) and is forwarded verbatim into `run_master_token_login` (`session_cmd.py:355-368`), so this is user-reachable, not theoretical.

## Evidence

| Site | Code | Policy |
|---|---|---|
| `src/notebooklm/cli/master_token_login.py:35` | `storage_path = Path(storage) if storage else get_storage_path(profile=profile)` | honors `--storage` |
| `src/notebooklm/cli/master_token_login.py:36` | `master_token_path = get_master_token_path(profile)` | **ignores `--storage`** |
| `src/notebooklm/paths.py:340-352` | `get_master_token_path()` — docstring: *"Sibling of storage_state.json."*; body: `get_profile_dir(resolve_profile(profile)) / "master_token.json"` | profile-only; takes no `storage_path` |
| `src/notebooklm/_auth/recovery.py:223` | `master_token_path = canonical_path.parent / "master_token.json"` | derived from storage |
| `src/notebooklm/_app/auth_check.py:231` | `path = plan.storage_path.with_name("master_token.json")` | derived from storage |
| `src/notebooklm/cli/services/auth_refresh.py:62` | `master_token_path = storage_path.parent / "master_token.json"` | derived from storage |

`get_master_token_path` has exactly one caller in `src/` — the buggy site above.

`auth_check.py:220-224` states the intended contract outright: *"resolve it relative to the actual storage path — this also honors a `--storage` override."* The login writer is the only place that doesn't.

Note the sibling helper `get_context_path(profile, *, storage_path=...)` (`paths.py:355+`) already takes a `storage_path` override and returns a sibling for it. `get_master_token_path` has no such parameter.

## Reproduction

```python
from pathlib import Path
from notebooklm.paths import get_storage_path, get_master_token_path
storage = "/tmp/probe/foo.json"
storage_path = Path(storage)                            # what --storage produces
master_token_path = get_master_token_path(None)         # what the CLI writes to
[...]

PR #2105, opened 2026-08-06
fix(cli): canonicalize --storage before deriving the master-token sibling

teng-lin:
Follow-up to the #2104 review (P2 finding, [comment 3731587648](https://github.com/teng-lin/notebooklm-py/pull/2104#discussion_r3731587648)). Refs #2103.

## Why

#2104 established the sibling invariant — `master_token.json` lives beside the storage file, because **every reader derives it that way** — but the writer used the raw `--storage` value.

The L4 master-token recovery rung canonicalizes *first*. `try_master_token_reauth` runs the path through `canonical_storage_key` (which is `expanduser().resolve()`) and only then takes the parent:

```python
canonical_path = canonical_storage_key(storage_path)
master_token_path = canonical_path.parent / "master_token.json"
```
<sub>`src/notebooklm/_auth/recovery.py:221-223`</sub>

## The alias failure mode

Any alias for the same storage file — a relative path, a `~` prefix, or a symlinked directory — splits the writer and the reader apart:

| | path used | token location |
|---|---|---|
| `login --master-token --storage ~/link/auth.json` | verbatim | `~/link/master_token.json` |
| L4 recovery rung | resolved | `/real/dir/master_token.json` |

Login reports success, the token is on disk, and the recovery ladder still reports "no token" — the same silent failure #2103 fixed, reached through a different spelling of the path. `auth check` reports `present: False` for the same reason.

## The fix

Canonicalize an explicit `--storage` in the login driver, exactly as the auth-source resolver already does for the very same option:

```python
storage_override = Path(raw_storage).expanduser().resolve()
```
<sub>`src/notebooklm/cli/services/auth_source.py:111`</sub>

Every syntactic spelling of one storage file now collapses to one token location, and the writer agrees with the reader by construction rather than by coincidence. Profile-derived paths are already absolute, so the no-`--storage` branch is byte-for-byte unchanged.

## Test coverage

- New `test_master_token_refresh_storage_symlink_selects_target_sibling` — creates a symlinked directory alias and asserts the token lands beside the **target**. Verified to fail without the source fix (it writes `alias/master_token.json` instead of `real/`), and skipped where the platform cannot create directory symlinks.
- The two #2103 regression tests now assert the resolved paths.
- Full suite green: 13332 passed, 64 skipped, 1 xfailed. `ruff check` / `ruff format --check` whole tree, `mypy src/notebooklm` clean.

## Not in scope

The P1 finding on #2104 ([comment 3731587652](https://github.com/teng-lin/notebooklm-py/pull/2104#discussion_r3731587652)) — two custom storage files in one directory sharing a single `master_token.json`, unlike `get_context_path`'s per-storage namespacing — is a real cross-account hazard on the L4 read side, but fixing it changes the ADR-0023 on-disk layout contract and every reader. It is deferred to the planned master-token structural-move PR tracked from #2103.

## Summary by CodeRabbit

* **Bug Fixes**
  * Explicit storage paths are now consistently resolved before locating or creating the master token.
  * Relative paths, home-directory shortcuts, and symlinked storage paths now correctly map to the same token location used during recovery.

* **Documentation**
  * Added an Unreleased changelog entry describing the storage path resolution improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -162,6 +162,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   account-ownership guard now checks the token that actually sits beside the
   target storage instead of the active profile's.
 
+- **A symlinked or relative `--storage` alias now selects the same
+  `master_token.json` the recovery ladder reads (#2104 review).** `--storage`
+  was used verbatim when deriving the token sibling, so an alias — a relative
+  path, a `~` prefix, or a symlinked directory — wrote the token beside the
+  *alias* while the L4 master-token recovery rung, which canonicalizes through
+  `canonical_storage_key` (`expanduser().resolve()`) before taking the parent
+  directory, looked beside the resolved target and found nothing. The login
+  driver now canonicalizes an explicit `--storage` the same way the
+  `cli.services.auth_source` resolver already does, so every syntactic spelling
+  of one storage file collapses to one token location. Profile-derived paths
+  were already absolute and are unaffected.
+
 - **`NOTEBOOKLM_AUTH_JSON` now beats a profile everywhere, as documented.** The
   precedence `--storage` > `NOTEBOOKLM_AUTH_JSON` > profile file is stated in
   `docs/configuration.md`, drawn in `docs/architecture.md`, and implemented by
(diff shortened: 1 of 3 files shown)


teng-lin:
@claude review

PR #2108, opened 2026-08-07
refactor(auth): master-token relocation — chokepoint, transaction move, minting denylist

teng-lin:
## Summary

Structural follow-up to #2103, in four stages (each implemented on its own branch, reviewed by a native multi-agent team, then merged into this integration branch — see individual merge commits for full review history):

1. **PR-0 — legacy `context.json` account migrate-then-drop.** Extends startup migration to promote a legacy sibling `context.json`'s `account` key in-band (via the canonical storage writer) before dropping the standing read fallback, closing a wrong-account-routing hazard for pre-v0.5.0 profiles. Fixed 4 review-round bugs: promotion-failure collapsing to `{}` instead of a sanitized legacy fallback, an unlocked check-then-act race, a lock-deadline mismatch now that promotion runs on the request path, and a phantom-file-creation bug in `promote_legacy_account`. A `notebooklm auth import-cookies` bug that permanently destroyed legacy account bindings was found and fixed during review.

2. **PR-1 — master-token path chokepoint.** Converges four independent `master_token.json` sibling-derivation sites (`_auth/recovery.py`'s L4 rung, `_app/auth_check.py`, `cli/services/auth_refresh.py`, `cli/master_token_login.py` — previously three different canonicalization policies for a symlinked/relative `--storage`) onto `notebooklm.paths.master_token_path_for`, guarded by an AST lint against a second derivation site reappearing. Review found and fixed a real crash: `Path.resolve()` raises `RuntimeError` (not `OSError`) on a symlink loop on Python 3.10–3.12 (a CPython pathlib behavior fixed in 3.13) — verified against real 3.10/3.11/3.12 interpreters, not just this repo's own dev interpreter.

3. **PR-2 — move the master-token transaction into `_auth`.** Relocates bootstrap/re-mint/ownership-guard from `cli/services/login/master_token.py` and `cli/services/auth_refresh.py` into `_auth/master_token.py`; the CLI now calls whole audited transactions (`master_token_bootstrap` / `master_token_remint` / `master_token_bootstrap_storage` / `assert_account_writable`) via the `notebooklm.auth` facade instead of assembling minting primitives itself. New: `remint_from_stored_token` (the single kernel both the L4 recovery rung and the operator refresh path now share, previously duplicated with diverging error handling), `BootstrapOutcome` (a four-state result replacing a bool that conflated distinct cold-start cases), and the authoritative account-ownership guard on `persist_minted_jar` — closing the #2104 review thread (`discussion_r3731673393`) that bound ADR-0023 to fixing the L4 read-side cross-account re-mint. Review found and fixed one MAJOR security bug (a refused bootstrap could still leave a durable master token for the wrong account on disk — `write_master_token` ran before the authoritative gate) and one MAJOR silent-failure bug (a corrupted `master_token.json` was silently treated as absent, generating a fresh `android_id` instead of raising).

4. **PR-3 — minting denylist + ADR amendments.** Locks in PR-2's invariant with a new AST guardrail: no module under `cli/` may import or attribute-access `exchange_master_token` / `mint_cookies` / `persist_minted_jar` / `write_master_token` from the `notebooklm.auth` facade, by name or via a module-alias attribute call. Amends ADR-0021 with the relocation as a further data point for its "would a headless server call this?" placement test.

## Design

Design decisions D1-D8 were tracked in a working planning doc during this session (not committed to the repo). `notebooklm.auth`'s cross-boundary ledger (`AUTH_CROSS_BOUNDARY_NAMES` / `_AUTH_DEBLESSED_KEEP_IMPORTABLE` in `tests/_guardrails/test_public_surface.py`) moved net-zero on the cross-boundary side (−5 old minting primitives, +5 new coarse ops/types — the plan's −5+3 estimate didn't account for `BootstrapOutcome` and the bootstrap-storage op also needing to cross the CLI boundary through the public facade) and +5 on the de-blessed side (26 → 31), as planned.

## Review discipline

Every stage: implement on its own branch → full test suite → native multi-agent review team (code quality, security, reliability/concurrency, silent-failures, test-coverage, sized to the stage's risk) → every finding addressed with a verified fix (revert-confirm-restore: temporarily undo the fix, confirm the regression test fails, restore, confirm it passes again) → merge. Two reviewers across the four stages never returned a report within a bounded wait; each time, an independent self-review pass over the full diff compensated (and in two cases — PR-2's ephemeral-reviewer-name cleanup, PR-3's package-qualified-import gap — caught a real, separate issue).

## Test plan

- [x] Full unit test suite green on every stage and on the final merged branch (only pre-existing, environment-caused `playwright`-missing failures remain, unrelated to this work — confirmed identical with the diff stashed)
- [x] `ruff check` / `ruff format --check` / `mypy src/notebooklm --ignore-missing-imports` clean (aside from 3 pre-existing errors in `_source/markdown.py`, confirmed present on the base commit too)
- [x] All guardrail tests pass, including the two new AST lints added by this PR
- [x] Live-verify: `login --master-token-refresh` (kernel path) and `auth refresh --verify` with storage deleted (bootstrap path) against a throwaway copy of a real profile — both succeeded, and a subsequent `list` call against the re-minted session returned real notebooks in each case.

https://claude.ai/code/session_01QaeZfcxWvV3AuogrsiswrP

## Summary by CodeRabbit

* **New Features**
  * Added safer master-token bootstrap and recovery with concurrent-start handling, explicit outcomes, and account-ownership protection.
  * Added automatic promotion of legacy account metadata into unified storage.
  * Added canonical master-token path handling, including symlink and fallback scenarios.

* **Deprecations**
  * Deprecated the pre-profiles storage layout in v0.9.0, with removal planned for v1.0.

* **Documentation**
  * Updated authentication, CLI, architecture, migration, and deprecation guidance for the new storage and recovery workflows.


teng-lin:
@claude review

teng-lin:
Addressed in e7d7155d (and 09d678d3, 6dbd9895):

**Fixed:**
- `_auth/master_token.py:541-544` (`_bootstrap_lock_path` missing `Path.resolve()` error handling) — fixed with the same degrade-gracefully pattern PR-1 already applied to `master_token_path_for`; added both a real symlink-loop test and a deterministic monkeypatch-based regression test, verified via revert-confirm-restore.
- `tests/unit/test_paths.py:381-385` (missing `try/except(OSError, NotImplementedError)` + skip guard) — fixed, matches the sibling test's pattern now.
- `tests/unit/test_auth_account_coverage.py:330-388` (`_PROMOTION_WARNED_PATHS` not cleared between tests) — added the autouse fixture exactly as proposed.
- `src/notebooklm/cli/services/login/master_token.py:4-11` (docstring inaccuracy) — corrected; the module no longer reaches the transaction ops itself, `cli/master_token_login.py` does.
- `MASTER-TOKEN-MOVE-PLAN.md` removal-version references — fixed for consistency (moot now: the plan doc itself has been removed from the repo per maintainer request, it was a working session doc not meant for permanent inclusion).

**Also fixed (not flagged by CodeRabbit, but this round of review surfaced it):** `tests/unit/test_paths.py::TestMasterTokenPathFor::test_tilde_path_expands` hardcoded a literal `/tmp/...` prefix, which broke on macOS (`/tmp` → `/private/tmp` symlink) and Windows CI — this was the actual CI failure on this PR. Now uses `tmp_path` and compares against its resolved form.

**Verified, no action needed:**
- `src/notebooklm/auth.py:52-79` — confirmed `__all__` is unchanged (still exactly the 6 documented names) and none of the new transaction ops/`BootstrapOutcome` appear in `docs/stability.md`. Nothing was accidentally published.

**Declined, with reasoning:**
- `CHANGELOG.md:268` (duplicate `### Changed` heading) — this is pre-existing, established practice in this file: `main` already has a second `### Changed` header later in the same `[Unreleased]` section (and several more throughout the file's history), from prior PRs each appending their own category header rather than merging into an earlier one. Renaming to a unique heading would be *more* inconsistent with the file's own convention, not less, and restructuring the whole file is out of scope here.
- `src/notebooklm/_auth/account.py:333-355` (`promote_legacy_account` blocking the event loop) — verified this is currently unreachable in practice: grepped the full `src/` tree (including `mcp/` and `server/`) and confirmed no async caller of `read_account_metadata`/`promote_legacy_account` exists anywhere today; all 5 real call sites are synchronous CLI/`_app` code with no running event loop. The promotion path is also one-shot and self-healing (bounded to a 2s lock deadline, and never re-triggers after the first successful promotion for a given profile). Wrapping a fundamentally synchronous, filesystem-only primitive in `asyncio.to_thread` speculatively, for a caller that doesn't exist, would be adding complexity ahead of need — happy to revisit if/when an async caller (e.g. a future MCP/REST profile-listing route) actually appears.

Full test suite, ruff, and mypy all green on the final commit (aside from 3 pre-existing, unrelated mypy errors in `_source/markdown.py`, confirmed present on `main` too).

Issue #2134, opened 2026-08-07
failed_error_text reads two phantom fields (index 3 is sources, index 5 is isPubliclyReadable) — dead code, and its unit tests encode a schema that does not exist


teng-lin:
## Summary

`_row_adapters/artifacts.py` documents artifact row index 3 as "failed-artifact plain error text" and index 5 as "failed-artifact nested error payload". **Neither field exists.** Index 3 is `Artifact.sources` and index 5 is `isPubliclyReadable`. `ArtifactRow.failed_error_text` is dead code that can never fire, and it is wired into the polling path — so every failed generation surfaces `None` where the API promises an explanation.

Notably, **its unit tests pass by constructing a schema that does not exist.**

## Evidence (live-verified, not inferred)

The two constants and their docstrings live at `src/notebooklm/_row_adapters/artifacts.py:117,119` (`_ERROR_TEXT_POS = 3`, `_ERROR_PAYLOAD_POS = 5`). The mobile proto and gRPC both contradict them directly:

```
index 3  ->  Artifact.sources          (tag 4, repeated ArtifactSource)
index 5  ->  Artifact.isPubliclyReadable (tag 6)
```

gRPC on the same artifact returns `tag 4: <42 sources>`, each shaped `{1: {1: '<uuid>'}}`.

Run against **three real failed artifacts**, `ArtifactRow.failed_error_text` (`src/notebooklm/_row_adapters/artifacts.py:553`) returned `None` every time, with `raw[3]` holding the source-id list and `raw[5]` null. It is saved from emitting a source UUID *as* an error message only because index 3 happens to be a list rather than a string.

It is consumed at `src/notebooklm/_artifact/polling.py:490`, so the dead path is on the live failure-reporting route.

### The tests encode the fiction

`tests/unit/test_row_adapters.py:393-404` constructs synthetic rows with **strings at index 3** — a shape the backend never produces. The tests pass, and they pass precisely because they assert against an invented schema rather than a captured one. They are currently evidence *for* the wrong model.

## Why this matters

- Every failed artifact generation reports no reason at all, while the code and its docstrings claim it reports one. Anyone debugging a failure trusts a field that is structurally incapable of being populated.
- The docstrings mislabel two fields that *are* real and useful (`sources`, `isPubliclyReadable`), so anyone reading the adapter to find them is sent the wrong way.
[...]

PR #2144, opened 2026-08-07
test(guardrails): validate positional wire constants against the recovered mobile schema

teng-lin:
## Why

The `batchexecute` client reads responses by **hardcoded array index**. The wire carries no field names, so a wrong index is silent — it returns a plausible value of the wrong thing. Worse, a unit test written from the same misunderstanding *confirms* the mistake rather than catching it.

There are **112 such constants** across the row adapters, and until now nothing validated a single one. Given that RPC drift is this project's documented #1 breakage class, that's a gap worth closing structurally rather than one bug at a time.

An audit against the schema recovered from the official Android app found real defects of exactly this shape — an inverted ownership flag, transposed generation options, a swapped status enum, and two constants naming fields that have **never existed**, the last with *passing* tests built around the imaginary shape.

## What this adds

A guardrail that checks our constants against an independent source of truth. Mobile gRPC carries the same backend messages tag-addressed, and the two line up exactly:

```
JSON index i  ==  protobuf tag (i + 1)
```

Verified independently, e.g. `Source.settings` is tag 4 read at index 3, whose `SourceSettings.status` is tag 2 read at index 1 — matching this client's documented `source[3][1]` descent.

**Three checks:**

| check | what it prevents |
|---|---|
| Declared mappings hold (`constant == tag - 1`) | a constant silently pointing at the wrong field |
| No constant escapes review | a new positional read landing with nobody recording what it points at |
| Enum members match the backend | inventing values, or missing real ones |

**Three registry categories**, reflecting how much is actually known:

- **`Mapping`** — schema-validated against a named proto field
- **`Pinned`** — meaning established by live evidence only (`addUnused()` slots the proto can't confirm); a change-detector, explicitly documented as weaker
- **`UNMAPPED`** — honest "not checked", reason on the record, including a `_NO_SUCH_FIELD` category for indexes the message has **no field for**

The design bias is deliberate: *an honest `UNMAPPED` entry is better than a guessed `Mapping`*, because a wrong mapping lends false confidence to the exact defect class this exists to catch.

Two known-wrong constants are `xfail` with an issue reference (#2134); the fixing PR must remove the marker.

## Also: `docs/mobile/` consolidation

Six top-level RE files in `docs/` with no entry point. Grouped under `docs/mobile/` with a README, plus the three scripts that generate them.

Grouped rather than merged, deliberately — `schema.proto` and `enums.txt` are **generated artifacts CI parses**, and the blutter patch must stay `git apply`-able. The README front-loads caveats that cost real time to learn (`fieldType` is a parse failure not a field name; several messages exist twice with different tags; `addUnused()` describes what the *client* ignores, not what the backend omits).

## Test plan

- [x] `pytest tests/_guardrails -m repo_lint` — 598 passed, 3 xfailed
- [x] Full unit suite — 11,091 passed, 62 skipped
- [x] `ruff format --check .` / `ruff check .` — clean (1013 files)
- [x] `mypy src/notebooklm` — clean (333 files)
- [x] **Mutation-tested the guardrail** rather than assuming it works:
  - breaking a verified constant (`_STATUS_BLOCK_POS` 3→2) → mapping check fails
  - adding an undeclared constant → coverage check fails
- [x] All 8 relative links in moved docs verified to resolve (the docs-refs gate caught one the move broke)
- [x] Staged content scanned for credentials and personal emails — clean
- [ ] CI green

## Notes for review

**Known limitation, documented in the module docstring:** this validates *constant → tag*, so it is blind to a fixture with the **correct index and the wrong type** — which is a real, separately-identified defect class. Closing that needs a fixture-vs-wire type check derived from cassette histograms; it is not in this PR.

The enum dump **must** be the merged one (74 enums / 273 values from the object pool alone, vs 77 / ~1900 merged). Auditing against the pool alone manufactures false "we invented this value" findings and hides real members — `ARTIFACT_PENDING_REVIEW` was missed exactly that way.

No `src/` behaviour changes. Test infrastructure, reference data, and docs only.

## Summary by CodeRabbit

- **Documentation**
  - Added guides for mobile API traffic capture, authentication, troubleshooting, endpoint behavior, schema recovery, and development workflows.
  - Documented recovered API schemas, enums, field mappings, safety considerations, limitations, and Dart tooling compatibility.

- **New Tools**
  - Added utilities to capture, decode, inspect, and reconstruct mobile API traffic and schemas.
  - Added support for analyzing newer Dart runtime artifacts.

- **Tests**
  - Added wire-contract checks for field mappings, enum values, reference data, and recovered schema compatibility.


teng-lin:
@claude review

Specific things worth a critical eye:

1. **Is the index↔tag premise sound?** Everything rests on `JSON index i == protobuf tag (i+1)`. It's verified on several messages, but if it's ever false the guardrail would assert wrong things confidently.

2. **Are the `Pinned` entries honest?** Those read `addUnused()` slots the proto can't confirm, so the check is only a change-detector. The `_DATA_TABLE_PAYLOAD_POS` entry in particular rests on a **single observation** — I flagged it as WEAK in the note, but push back if that's still too thin to encode.

3. **Windows path handling** — the two reference-data paths resolve at import time.

4. **Does the `UNMAPPED` escape hatch weaken this too much?** A contributor could silence the coverage check by dumping a constant into `UNMAPPED` with a hand-wave. The bias is intentional (an honest 'unknown' beats a guessed mapping) but the incentive is worth scrutinising.

teng-lin:
@codex review

teng-lin:
Addressed in 3fa34238 (and ade419ad / 9ac5b2b5 for the earlier ones). All ten findings were verified against the code before acting — summary of what was real and what I changed:

**Valid, fixed:**
- `development.md` + the `_wire_contract` module docstring both documented only `MAPPINGS`/`UNMAPPED`. `PINNED` was added later and never propagated, so a contributor would have filed an `addUnused()` slot as unknown when live evidence could pin it. Now a table ranking all three by strength.
- **14 broken links** from my move — more than the 5 flagged. My bare-filename rewrite produced `mobile/schema.proto` *inside* `docs/mobile/`, and one `adr/…` link needed `../adr/` at the new depth. My own link check had reported "8 checked, 0 broken" because it only matched dot-prefixed targets and silently skipped these. Re-verified properly: **32 targets, 0 broken**.
- `capture.md` scope said "Read-only" while a section drives create/delete/share RPCs; and the loopback DNAT fallback could never fire (the primary rule is appended to the same `OUTPUT` chain and also matches `--dport 8000`), with cleanup leaving the loopback rule, `adb reverse` tunnel and `route_localnet=1` behind. Both fixed, plus verification commands.
- `test_declared_enum_gaps_still_exist` indexed `ENUM_BINDINGS` directly while parametrizing over `ENUM_GAPS` — a gap-only enum would raise `KeyError`. Verified by injecting a synthetic gap-only enum; now asserts with a message naming the fix.
- `decode_mobile_grpc.py` never recorded fixed64.

**Two notes where I diverged from the suggestion:**

1. **The fixed64 patch would have caused a regression.** The proposed diff *replaced* the `wt == 5` branch with `wt == 1`, trading the fixed32 gap for a fixed64 one. Added alongside instead; both wire types now recorded.

2. **The method count: both documented numbers were wrong.** The suggestion was to reconcile 20 against 22. Counting the sections gives 22 `###` headings, of which 3 aren't RPCs (UUID, Timestamp, `RequestContext`) and 2 document two methods each — so **21**. All three sites now say 21.

**On line length**, worth recording for future reviews: `line-length = 100` is configured but `E501` is not in ruff's `select`, so over-length lines don't fail `ruff check` or CI (there are 11 pre-existing instances under `tests/_guardrails/`). Mine is wrapped regardless; I only touched my own.

teng-lin:
Thanks — the dead-link finding was correct and more serious than it looked from outside, because the file was not merely missing: **it was one I had been explicitly told not to commit, and I had committed and pushed it anyway.**

Cause: in the previous commit I ran `git add docs/mobile/` on the whole directory rather than naming files, and my link-repair pass had touched `auth-research.md`, so it was swept in. Your finding is what surfaced it.

Fixed in `9c82299e` by rewriting the commit rather than deleting in a follow-up, so the content is not recoverable from the PR's history:

```
auth-research.md on remote branch : removed
occurrences in branch history      : 0
```

The README row linking to it is gone, which is the dead link you flagged.

**Why my own checks kept missing this, since it is the actual lesson:** my link verifier resolved targets against the **working tree**, where the file exists locally. It never asked whether the target was *in the repo*. So it reported "0 broken" three times while a dead link sat in a committed file. Now verified against `git ls-files` — **26 targets, 0 dead**.

That is the third false negative from my own tooling in this PR (the first skipped non-dot-prefixed targets; the second counted nothing and reported success). A verifier that cannot distinguish "all clear" from "found nothing to inspect" is worse than none, and I should have caught the pattern earlier.

On your other four points — no action taken, and I agree with each:

1. **Index↔tag premise** — appreciated that you checked the non-consecutive gaps (`googleDocsMetadata=1, originalSourceContentType=5, webpageMetadata=8, googleDriveSourceMetadata=10` against `0/4/7/9`) rather than sampling consecutive tags where the relationship would hold trivially. That is the right way to test it.

2. **`_DATA_TABLE_PAYLOAD_POS`** — agreed it is disclosed and low blast radius. I will file the corroboration tracking issue you suggest rather than block on it.

3. **Windows** — your point that the reads happen lazily inside `lru_cache` rather than at import is a better characterisation than mine; a missing file surfaces as a normal collection-time error, not an import crash.

4. **`UNMAPPED` incentives** — agreed it cannot be enforced mechanically, and that forcing a guessed `Mapping` would be strictly worse. Your framing that an `UNMAPPED`-adding PR deserves the same scrutiny as a `Mapping`-adding one is the right place to put it; I will add that to the contributor docs.

Also noting for the record that you flagged working without a live test run. That was the correct call to disclose, and the finding you did land was one four passing CI runs and my own checks had all missed.

PR #2147, opened 2026-08-07
test(auth): pin the mint leg's rejected-rotation contract

teng-lin:
Closes a test gap `claude[bot]` identified while reviewing #2146. It raced that PR's merge rather than being an afterthought: the commit was pushed while #2146 was still open, but the squash merge landed ~4 minutes earlier and captured only up to `2c2f4da0`, so the test never made it to `main`. This is that commit, rebuilt on `main`.

## The gap

#2146 consolidated the quadruple-implemented `RotateCookies` POST into one wire contract. Its **only** production behavior delta is in the mint leg: `master_token.mint_cookies` now routes through `keepalive._rotate_post`, which adds the `raise_for_status` the old inline POST lacked. A 429/5xx that previously passed silently as a 200-shaped no-op is now an `httpx.HTTPStatusError` that gets logged and skipped.

Nothing pinned that. The adjacent `test_remint_from_stored_token_wraps_reload_failure_when_psidts_withheld` covers the *neighbouring* case — a 200 with no `Set-Cookie` — not a rejected status.

## What the test pins

`test_mint_cookies_survives_a_rejected_rotation`, parameterized over 429 and 503, asserts the contract the #2146 description claims:

- the mint still **succeeds** — the rotation is best-effort, so the inner handler logs and skips, and the failure must not escalate through the *outer* `except httpx.HTTPError` that wraps the whole mint;
- the jar comes back without `__Secure-1PSIDTS`, leaving the standard inline recovery to mint it on first load;
- the debug line actually fires.

## Why the log assertion is load-bearing

It is doing the real work here, and it is the reason this test is worth having rather than merely present. Under the *old* code a silently-swallowed 429 **also** yields a jar with no `__Secure-1PSIDTS` — so the cookie-set assertions alone would pass against the bug the PR fixed.

Verified by mutation rather than by watching it go green: deleting `raise_for_status` from `keepalive._rotate_post` makes both parameterizations fail, and fail specifically on the missing log line. That confirms the test pins the delta and not ambient behavior.

## Verification

Test-only — no production code touched (single file, +36 lines).

- `tests/unit/test_auth_master_token.py` + `tests/_guardrails/test_rotate_wire_contract.py`: 40 passed, run against the **merged** `main`, not the old branch.
- `ruff check` and `ruff format --check` clean.

https://claude.ai/code/session_01UC4GLbcdJx6UTb5mjsp6HY

## Summary by CodeRabbit

* **Tests**
  * Added coverage for cookie rotation failures caused by HTTP 429 and 503 responses.
  * Verified cookie minting continues successfully, the appropriate cookie is omitted, failures are logged, and the rotation request is attempted.

### DIFF
--- a/tests/unit/test_auth_master_token.py
+++ b/tests/unit/test_auth_master_token.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import logging
 import re
 import sys
 import types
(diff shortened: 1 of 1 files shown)


teng-lin:
@claude review

PR #2150, opened 2026-08-07
fix(auth): preserve sameSite through CookieJar + close ratchet bypasses

teng-lin:
## Summary

Two review fixes for #2148 (ADR-0031 Stage 1) that were **lost to a merge race** — #2148 was merged at its old head `caf1f4fa` while these were still in CI, so main currently carries both defects. Same race that cost #2146 a test (rescued as #2147).

Neither is speculative; both were found by review on #2148 and verified by injection.

### 1. `CookieJar` silently dropped `sameSite` (data loss)

`Cookie` had no `same_site` field, so `to_storage_state()` erased it:

```python
convert_rookiepy_cookies_to_storage_state(rows)      # {..., 'sameSite': 'Strict'}
CookieJar.from_rookiepy(rows).to_storage_state()     # sameSite gone
```

The sanitized row *does* carry `sameSite` (`validate_cookie_shape` does `dict(entry)`); it was dropped specifically by `_cookie_from_entry`. This is precisely the regression `storage._preserved_same_site` exists to prevent, and the converter documents preserving it. Latent today because no production caller uses `to_storage_state()` — but this type is slated to become canonical in ADR-0031 Stage 4, and the existing test already pinned `httpOnly`/`secure`/`expires`, so `sameSite` was the one flag with a dedicated protection mechanism that went unpinned.

`to_storage_state()` emits `sameSite` **only when not None**, so it never invents a value — an invented one would read as a real downgrade to the next `_preserved_same_site` merge. Now `CookieJar.from_rookiepy(rows).to_storage_state() == convert_rookiepy_cookies_to_storage_state(rows)` exactly, pinned by a jar-vs-converter equality test.

### 2. The conversion ratchet was bypassable two ways

- **Rename-on-import**: the AST scan matched `ast.Name.id`/`ast.Attribute.attr` against the literal name, so `from .cookies import normalize_cookie_map as _n; _n(x)` was invisible. Not theoretical — this package already uses that idiom (`storage_writer.py`, `browser_capture.py`). Renamed imports are now resolved back to the canonical name.
- **Per-module pair keying**: `_GRANDFATHERED` held `(module, function)` pairs, so a listed module could grow a second bespoke call for free. Also not theoretical — `browser_cookie_recovery.py` already had **two** `convert_rookiepy_cookies_to_storage_state` calls, and the entry's own comment said "converts rookiepy rows twice" while the gate pinned nothing. It now pins **measured counts**, shrink-only in both directions (counts may never rise; a pin above reality must be lowered or deleted), mirroring `test_string_patch_ratchet.py`'s ceiling/anti-slack/anti-stale shape.

## Test plan

- [x] Ratchet + cookie_types tests: 35 passed against main
- [x] `sameSite` preserved through the jar and equal to the legacy converter
- [x] Injection-tested both bypasses: an extra `normalize_cookie_map` call in `tokens.py` reds with `2 call(s), pinned at 1`; an aliased import in a new module reds with `1 call(s), pinned at 0`
- [x] Full suite / `tests/_guardrails` / mypy / ruff clean (verified on the source branch before the race)

https://claude.ai/code/session_01QaeZfcxWvV3AuogrsiswrP

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved cookie `SameSite` settings during authentication state storage and restoration.
  * Continued omitting `SameSite` when no value is provided, preventing unintended cookie changes.
  * Improved compatibility when reading cookie data that uses different field naming conventions.
  * Ensured cookie conversions consistently retain existing values across supported authentication flows.

### DIFF
--- a/docs/adr/0031-credential-tier-auth-model.md
+++ b/docs/adr/0031-credential-tier-auth-model.md
@@ -85,7 +85,9 @@ introduce its objects and operations in independently shippable stages:
   the free function that owns it, pinned by equivalence tests. A shrink-only
   ratchet (`tests/_guardrails/test_cookie_conversion_ratchet.py`) blocks *new*
   bespoke conversion call sites; the five existing ones are grandfathered with
-  the stage that retires each, and the allowlist may only shrink.
+  the stage that retires each, pinned at their measured call counts so a module
+  already on the list cannot grow a second one for free, and the allowlist may
+  only shrink.
 
   **The flat `name -> value` shape is deliberately not on the type.** It
   collapses the path component (#369) and picks an arbitrary winner among
--- a/src/notebooklm/_auth/cookie_types.py
+++ b/src/notebooklm/_auth/cookie_types.py
@@ -54,6 +54,10 @@ class Cookie:
     distinct domains or paths are independent entries, never one shadowing the
     other (issue #369). ``expires`` is the Playwright epoch-seconds convention
     where ``-1`` means a session cookie.
+
+    ``same_site`` is carried rather than defaulted: dropping a stored value is
+    the exact regression ``storage._preserved_same_site`` exists to prevent, so
+    ``None`` here means "the row carried none", never "assume ``None``".
     """
 
     name: str
@@ -63,6 +67,7 @@ class Cookie:
     expires: float | int | None = None
     http_only: bool = False
     secure: bool = False
+    same_site: str | None = None
 
     @property
     def identity(self) -> tuple[str, str, str]:
@@ -165,21 +170,23 @@ class CookieJar:
         Reproduces the jar's *filtered* view, not any original file — see the
         module docstring. ``origins`` is empty: this type models cookies only.
         """
-        return {
-            "cookies": [
-                {
-                    "name": c.name,
-                    "value": c.value,
-                    "domain": c.domain,
-                    "path": c.path,
-                    "expires": -1 if c.expires is None else c.expires,
-                    "httpOnly": c.http_only,
-                    "secure": c.secure,
-                }
-                for c in self._cookies
-            ],
-            "origins": [],
-        }
+        rows: list[dict[str, Any]] = []
+        for c in self._cookies:
+            row: dict[str, Any] = {
+                "name": c.name,
+                "value": c.value,
+                "domain": c.domain,
+                "path": c.path,
+                "expires": -1 if c.expires is None else c.expires,
+                "httpOnly": c.http_only,
+                "secure": c.secure,
+            }
+            # Emitted only when the source row carried one, so this never
+            # invents a value the way a bare default would.
+            if c.same_site is not None:
+                row["sameSite"] = c.same_site
+            rows.append(row)
+        return {"cookies": rows, "origins": []}
 
     # -- queries -----------------------------------------------------------
 
@@ -243,6 +250,7 @@ def _cookie_from_entry(entry: Mapping[str, Any]) -> Cookie:
     ``_sanitized_auth_entries`` / ``sanitize_cookie_entry``: identity and value
     fields are trusted here, and ``path`` has already been defaulted to ``/``.
     """
+    same_site = entry.get("sameSite", entry.get("same_site"))
     return Cookie(
         name=entry["name"],
         domain=entry["domain"],
@@ -251,4 +259,5 @@ def _cookie_from_entry(entry: Mapping[str, Any]) -> Cookie:
         expires=entry.get("expires"),
         http_only=bool(entry.get("httpOnly", entry.get("http_only", False))),
         secure=bool(entry.get("secure", False)),
+        same_site=same_site if isinstance(same_site, str) else None,
     )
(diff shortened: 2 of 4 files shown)


teng-lin:
Context for reviewers: these two commits were originally reviewed on #2148 and pushed while it was in CI, but #2148 was squash-merged at the previous SHA before they landed, so they were orphaned and recovered here. The review discussion is stranded on that closed PR — reposting the substance so it is discoverable from the PR that actually carries the change.

**`d35def07` — sameSite was being erased.** `Cookie` carried no `sameSite` field, so the jar silently dropped it:

```
convert_rookiepy_cookies_to_storage_state(rows)  -> {..., "sameSite": "Strict"}
CookieJar.from_rookiepy(rows).to_storage_state() -> {...}   # gone
```

The sanitized row does carry it (`validate_cookie_shape` returns `dict(entry)`), and the converter documents preserving a browser-shaped row's value, so the loss was specific to `_cookie_from_entry`. Dropping a stored `sameSite` is the exact regression `storage._preserved_same_site` exists to prevent, and the existing test pinned `httpOnly`/`secure`/`expires` — leaving the one flag with a dedicated protection mechanism as the only unpinned field.

`sameSite` is emitted **only** when the source row carried one. Inventing a default would be worse than dropping it: `_preserved_same_site` treats a missing value as "nothing to keep", so a fabricated `"None"` would read as a real downgrade on the next merge rather than as absence. The load-bearing assertion is the jar-vs-converter equality, which is what would have caught this when the type was written.

**`8c537a66` — two ratchet holes, both found in review and both confirmed by injection.**

Renamed imports bypassed the gate: it matched literal names, so `from .cookies import normalize_cookie_map as _n; _n(x)` was invisible. Not a theoretical shape here — the package already uses rename-on-import in `_auth/storage_writer.py` and `_auth/browser_capture.py`, so a contributor reaching for the local idiom would have routed around it unintentionally.

Grandfathered modules could also grow new calls for free, since the allowlist was a set of `(module, function)` pairs. `_auth/browser_cookie_recovery.py` already holds **two** calls to `convert_rookiepy_cookies_to_storage_state` — its own annotation says "converts rookiepy rows twice" — so the code documented a fact the guardrail did not enforce, and a third would have passed silently. Each site is now pinned at its measured count, mirroring `test_string_patch_ratchet.py`'s existing ceiling/anti-slack/anti-stale shape. Shrink-only now holds in both directions, so slack cannot quietly become headroom.

Verified red by injection: an extra `normalize_cookie_map` in `tokens.py` fails with `2 call(s), pinned at 1`; an aliased import in a new module fails with `1 call(s), pinned at 0`.

Local verification on this exact tree (8c537a66): `tests/_guardrails` + the cookie_types unit file = 1282 passed, 3 skipped, 3 xfailed.

teng-lin:
**Please do not merge this until CodeRabbit and claude[bot] have both reviewed** — the reason this PR exists is that a merge outran its review cycle.

These two commits were review fixes on #2148. They were pushed while that PR was in CI, but #2148 was squash-merged at the previous SHA before they landed, so they were orphaned and recovered here. Merging this one on green checks alone, before the bots have posted, would repeat exactly the sequence that created it.

Two things worth knowing for anyone checking status on this PR, because they actively mislead:

- `gh pr view --json headRefOid` reported the OLD sha for more than five minutes after a successful push, and `gh pr checks` kept showing the previous run as all-pass during that window. Polling either would have said CI was green on commits that were not in the PR at all. The reliable tell was `git ls-remote` disagreeing with `headRefOid`.
- A `claude` check showing `skipping` is the normal non-review skip, not a completed review. claude[bot] posts a sticky summary comment plus inline threads and never appears in `reviewDecision`, so a green check list does not imply it has reviewed.

Also note codex is out of review quota on this repo right now and has posted a usage-limit notice rather than a review, so it should not be waited on as a gate.

teng-lin:
@claude review

teng-lin:
@claude review

teng-lin:
On the outside-diff finding (`cookie_types.py:253-262`, validate `sameSite` against `{"Strict","Lax","None"}`) — **declining, with the cases actually run rather than reasoned about.**

Half of it is already implemented. The existing `isinstance(same_site, str)` guard rejects non-strings today:

| input `sameSite` | resulting `Cookie.same_site` | key emitted by `to_storage_state()` |
|---|---|---|
| `123` | `None` | no |
| `{"a": 1}` | `None` | no |
| `None` | `None` | no |
| `"Bogus"` | `"Bogus"` | yes |

So "reject non-string values" and "map absent values to None" already hold. The only real gap is a malformed *string* on the `from_storage_state` path — the rookiepy path never reaches it, since `convert_rookiepy_cookies_to_storage_state` (`cookies.py:222-224`) normalizes to the three-value enum before the row is ever seen.

That gap is not currently reachable in production: `CookieJar` has no production callers at all, so nothing can write a bogus value to disk through `to_storage_state()`. The matches for `CookieJar` under `src/` are `http.cookiejar.CookieJar`, an unrelated stdlib class.

The stronger reason to decline is design, not reachability. This type is deliberately a delegating wrapper — every policy decision routes to the free function that owns it, which is the property the conversion ratchet and ADR-0031 exist to protect. Enum-restricting here would make the wrapper a second, independent owner of `sameSite` policy, reintroducing exactly the scattered-policy problem the type was introduced to retire. It would also be a behavior change against the stated preserve-don`t-invent contract, which `test_to_storage_state_omits_same_site_when_the_row_had_none` pins.

If we do want that validation, the right home is `cookie_semantics.sanitize_cookie_entry`, so every reader benefits rather than only jar users, and `storage._preserved_same_site` keeps a single source of truth. Happy to file that as a follow-up if wanted — flagging that it has a wider blast radius than it looks, since it would change what the storage-writer merge paths accept.

Both `claude[bot]` review runs reached the same conclusion independently.

Issue #1159, opened 2026-05-29
`sources.list()` silently returns `[]` on a malformed/error response (default `strict=False`)


teng-lin:
## Summary
`sources.list()` defaults to `strict=False` and converts a malformed/error-shaped notebook response into an empty list, so callers see "0 sources" instead of an error.

## Severity
**High** — silent failure (data-loss illusion). Confidence: high.

## Evidence
- `src/notebooklm/_source_listing.py` — `_handle_malformed_list_response` logs a WARNING and returns `None` (→ `[]`) for unexpected shapes (empty/non-list root, short `nb_info`, non-list `sources_list`). It uses hand-rolled `isinstance` checks, so it bypasses `NOTEBOOKLM_STRICT_DECODE`.
- `src/notebooklm/_sources.py` (`list()`/`get()`) and the `src/notebooklm/_source_upload.py` baseline probe all use the silent `strict=False` default.

## Failure mode
A drifted or error-enveloped `GET_NOTEBOOK` response is reported as "0 sources." A sync script may conclude the sources vanished and re-add everything.

## Suggested fix
Make malformed-response detection raise by default (or at least honor strict-decode here); reserve the empty list for a genuinely empty notebook.

---
_Identified during a runtime-correctness architecture review (round 2, 2026-05-29) via automated multi-lens audit; the top findings were verified against the code. Please confirm before scheduling._

PR #2159, opened 2026-08-09
fix(notebooks): stop reporting an empty notebook as schema drift (#2131)

### LINKED ISSUE #2131: get_source_ids logs a false 'schema drift?' warning on every empty notebook — desensitizes operators to real RPC drift
## Summary

`NotebooksAPI.get_source_ids` (`_notebooks.py:328-333`) logs a `schema drift?` warning every single time it is called on an empty notebook. An empty notebook legitimately **elides** the sources slot (`None`, not `[]`), which is a valid empty state, not a malformed response — and the sibling code path already knows this and carves it out. `get_source_ids` lacks the carve-out.

## Evidence (live-verified, not inferred)

Observed live on a freshly created, genuinely empty notebook:

```
WARNING get_source_ids: notebook_info[1] not list for b6e4e936-… (schema drift?). len=11
```

The `len=11` in that same line shows the envelope is perfectly healthy — nothing drifted.

The code that emits it (`_notebooks.py:328-333`) treats "not a list" as one undifferentiated failure:

```python
if not isinstance(sources, list):
    logger.warning(
        "get_source_ids: notebook_info[1] not list for %s (schema drift?). len=%d",
        notebook_id,
        len(notebook_info),
    )
    return source_ids
```

The sibling path over the *same slot of the same response* handles it correctly (`_source/listing.py:134-140`):

```python
if sources_list is None:
    # A genuinely empty notebook elides the sources slot (``None``
    # instead of an empty list). This is a valid empty state, NOT a
    # malformed response, so return ``[]`` without raising even under
    # strict-decode — issue #1159 reserves the empty list for the
    # genuinely-empty case (see tests/cassettes/notebook_zero_sources.yaml).
    return []
```

## Why this matters

This is a false positive on the one signal that must stay trustworthy. Obfuscated RPC method IDs and positional payload shapes changing under us is this project's #1 breakage class, and `schema drift?` in the logs is how an operator finds out. A warning that fires routinely — on every empty notebook, in normal healthy operation — trains operators to filter out that exact log string, so the real drift event arrives already silenced.

## Suggested fix

Give `get_source_ids` the same `sources is None → return []` carve-out as `_source/listing.py:134-140`, and keep the `schema drift?` warning for the genuinely-unexpected case (a non-list, non-`None` value).

---

Full context: `docs/notes/web-rpc-vs-mobile-grpc-audit-2026-08-07.md`, §1.19.

audichuang:
## Summary

`NotebooksAPI.get_source_ids` logs `schema drift?` at WARNING on **every** call against a genuinely empty notebook. Per the live-verified evidence in #2131, an empty notebook returns a healthy envelope (the reported `len=11`) whose sources slot is present but explicitly null — the backend sends `None` there, not `[]` — and the existing guard tested only `isinstance(sources, list)`, so a valid empty state landed in the branch reserved for a malformed response.

This matters more than the noise itself. Obfuscated method ids and positional payload shapes changing underneath us is this project's #1 breakage class, and `schema drift?` in the logs is how an operator finds out. A warning that fires on every empty notebook trains that operator to scroll past the line that will one day be real.

## Related Issue

Closes #2131

## Root cause

One `isinstance` test was standing in for three distinct shapes, only one of which is drift:

| `notebook_info` shape | Meaning | Old behaviour | New behaviour |
|---|---|---|---|
| too short to carry slot 1 | truncated envelope | warns | **warns** (unchanged) |
| slot present, explicitly `None` | genuinely empty notebook | warns ❌ | **returns quietly** |
| slot present, wrong type | shape changed under us | warns | **warns** (unchanged) |

The sibling walk over this same slot of this same response already makes exactly this split: `_source/listing.py` rejects a short `nb_info` first, and only then accepts a present `None` without raising, even under strict-decode (issue #1159). `get_source_ids` collapsed the first two cases together.

## Changes

- `src/notebooklm/_notebooks.py` — split the single `not isinstance(sources, list)` guard into the three cases above. The short-envelope branch keeps warning (with a message naming the actual condition rather than reusing the slot-type wording); the present-and-null branch returns quietly; the wrong-type branch is untouched.
- `tests/unit/test_swallow_observability.py` — two tests alongside the existing drift cases:
  - `test_get_source_ids_empty_notebook_emits_no_drift_warning` — the fix itself (`[[None] * 11]`, mirroring the reported envelope).
  - `test_get_source_ids_warns_when_the_sources_slot_is_absent` — the boundary (`[[None]]`), pinning that the carve-out did not widen into "return quietly whenever the slot expression is `None`".

No public API, exception type, or return type changes; no architectural shape change, so no ADR.

## Test Plan

- [x] Tests pass (`uv run pytest --cov=src/notebooklm --cov-report=term-missing --cov-fail-under=90`)
- [x] If this PR changes architectural shape, an ADR has been added or updated — N/A, no shape change

Run against the canonical contributor install plus the extras the CI test matrix adds (`--extra browser --extra dev --extra markdown --extra mcp --extra server --extra impersonate`):

```console
$ uv run mypy src/notebooklm --ignore-missing-imports
Success: no issues found in 334 source files

$ uv run pre-commit run --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed

$ uv run pytest --cov=src/notebooklm --cov-report=term-missing --cov-fail-under=90 -q
TOTAL                                                    28159    740   7634    450    97%
Required test coverage of 90% reached. Total coverage: 96.51%
13749 passed, 64 skipped, 3 xfailed, 18 warnings in 1601.22s (0:26:41)
```

Both new tests were confirmed non-vacuous by reverting only the source change and re-running: the empty-notebook test fails with the exact line from the issue —

```text
get_source_ids: notebook_info[1] not list for nb_empty (schema drift?). len=11
```

— and the absent-slot test fails against the naive carve-out described in Notes.

## Notes

**The obvious fix is wrong, and this is the second attempt.** The first version simply returned quietly whenever the slot expression evaluated to `None`. That expression also yields `None` when `len(notebook_info) <= 1`, so a truncated envelope — which *did* warn before — silently became "empty notebook", defeating the observability rationale that motivates this issue in the first place. `["title-only"]` went from a drift warning to `{"result": [], "warnings": []}`. The absent-slot test exists specifically to keep that regression from coming back.

**Scope of my verification.** Everything here was verified offline: unit tests, plus `tests/cassettes/notebook_zero_sources.yaml`, which carries the same present-null slot. The claim about *current* live backend behaviour is the maintainer's from #2131, not an independent live run — I do not have an account positioned to re-verify it.

**Not verified locally:** the Python 3.10–3.14 matrix. The change uses no version-sensitive syntax, but I can only exercise one interpreter here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty notebooks when retrieving source IDs.
  * Prevented unnecessary schema-drift warnings for valid notebooks with no sources.
  * Continued reporting truncated notebook responses where source information is missing.
  * Preserved warnings for notebook responses with invalid source data.

* **Tests**
  * Added coverage for valid empty notebooks, incomplete responses, and invalid source data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -229,6 +229,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **An empty notebook no longer logs `schema drift?` on every
+  `get_source_ids` call.** A genuinely empty notebook returns a healthy
+  envelope whose sources slot is present but explicitly null — the backend
+  sends `None` there, not `[]` — and the guard tested only
+  `isinstance(sources, list)`, so a valid empty state landed in the branch
+  reserved for a malformed response. Obfuscated method ids and positional
+  payload shapes changing underneath us is this project's #1 breakage class
+  and `schema drift?` is how an operator finds out, so a warning that fires on
+  every empty notebook erodes the one signal that has to stay trustworthy. The
+  walk now separates the three shapes the single `isinstance` test conflated:
+  an envelope too short to carry a sources slot still warns (a truncated shape,
+  as before), a present-and-null slot returns quietly, and a present-but-wrong
+  type still warns. That is the same split the sibling walk over this slot
+  already makes in `_source/listing.py`
+  ([#2131](https://github.com/teng-lin/notebooklm-py/issues/2131)).
 - **Chat turn numbers now come from server history instead of the client-local
   cache.** Stateless remote MCP requests create a fresh client for each call, so
   a real continuation could previously report the contradictory pair
--- a/src/notebooklm/_notebooks.py
+++ b/src/notebooklm/_notebooks.py
@@ -319,13 +319,32 @@ class NotebooksAPI:
                 )
                 return source_ids
 
-            sources = (
-                safe_index(
-                    notebook_info, 1, method_id=method_id, source="NotebooksAPI.get_source_ids"
+            if len(notebook_info) <= 1:
+                # The sources slot is *absent*, which is not the same thing as
+                # present-and-null below: a healthy envelope carries the slot
+                # (the #2131 report shows ``len=11`` on an empty notebook), so a
+                # response too short to hold one is a truncated shape worth
+                # surfacing.
+                logger.warning(
+                    "get_source_ids: notebook_info has no sources slot for %s "
+                    "(schema drift?). len=%d",
+                    notebook_id,
+                    len(notebook_info),
                 )
-                if len(notebook_info) > 1
-                else None
+                return source_ids
+
+            sources = safe_index(
+                notebook_info, 1, method_id=method_id, source="NotebooksAPI.get_source_ids"
             )
+            if sources is None:
+                # Slot present, explicitly null: a genuinely empty notebook
+                # elides its sources as ``None`` rather than ``[]``. A valid
+                # empty state, not a malformed response, so it must not reach
+                # the drift warning below (#2131). This is the same split the
+                # sibling walk over this slot already makes — reject the short
+                # envelope first, then accept a present ``None``
+                # (``_source/listing.py``, issue #1159).
+                return source_ids
             if not isinstance(sources, list):
                 logger.warning(
                     "get_source_ids: notebook_info[1] not list for %s (schema drift?). len=%d",
--- a/tests/unit/test_swallow_observability.py
+++ b/tests/unit/test_swallow_observability.py
@@ -77,6 +77,52 @@ async def test_get_source_ids_happy_path_no_warning(caplog):
     assert warnings == []
 
 
+@pytest.mark.asyncio
+async def test_get_source_ids_empty_notebook_emits_no_drift_warning(caplog):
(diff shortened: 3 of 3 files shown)


audichuang:
Pushed a follow-up: added the `[Unreleased] → Fixed` CHANGELOG entry I had missed on the first pass.

**Cross-reference with #2160.** Both PRs now insert an entry at the top of the same `### Fixed` block, so whichever lands second will need a trivial CHANGELOG rebase — the code changes themselves are disjoint (this one touches the `get_source_ids` envelope walk, #2160 touches `get()`'s error path), so there is no semantic interaction, only the textual one. Happy to rebase whenever the first one merges, or to drop my entry if you would rather write the release note yourself.

teng-lin:
@codex review
@claude review

teng-lin (approved):
Thanks, @audichuang—this is a careful, well-bounded fix. I especially appreciate the regression test that distinguishes an absent sources slot from a present-and-null one; that preserves the real schema-drift signal while eliminating the false warning. The implementation matches the sibling source-listing semantics, and the full check matrix is green. Approved.

Issue #2118, opened 2026-08-07
Every no-answer chat response is misclassified and emits a false "API format may have changed" warning


teng-lin:
## Summary

The backend's **normal** no-answer response — the canned prose apology it returns when the sources don't support the question — carries **no `TailwindDoc`**. `AnswerRow.is_answer` therefore evaluates `False`, the parser falls out of its marked-answer path into the emergency fallback, and emits a WARNING claiming the API format may have changed. The format has not changed: this is the backend's standard no-answer shape, and it occurs on every single no-answer ask.

## Evidence (live-verified, not inferred)

Observed live against the real backend on a no-answer ask:

```
AnswerRow.is_answer -> False
WARNING "No marked answer found; falling back to longest unmarked text (159 chars).
         The API response format may have changed."
```

The warning text is `src/notebooklm/_chat/wire.py:233-234`. The flag it depends on is `AnswerRow.is_answer` (`src/notebooklm/_row_adapters/chat.py:739-751`), which requires the type block to be present and marked; the no-answer response has no `TailwindDoc` at all, so the marker can never be found.

## Why this matters

Two distinct problems, neither of which is data loss — callers still receive the apology text, not an empty string:

1. **Misclassification.** A perfectly well-formed, expected backend response is routed through the parser's emergency path. Any future logic that keys off "did we take the fallback?" (metrics, error handling, retries) will treat routine no-answers as anomalies.
2. **A misleading diagnostic that erodes a real signal.** RPC/wire-format drift is documented in `CLAUDE.md` as this project's #1 breakage class. A warning that says "the API response format may have changed" and fires routinely on the most ordinary of responses trains operators to ignore that exact log string — so when the format genuinely does change, the warning is noise.

## Suggested fix

Recognise the no-answer shape explicitly (no `TailwindDoc` present at all, as distinct from a `TailwindDoc` present but unmarked) and route it through a normal, non-warning path. Reserve the "format may have changed" WARNING for the case where the answer document *is* present but the marker is missing — the only case that actually indicates drift.

---

Source: `docs/notes/web-rpc-vs-mobile-grpc-audit-2026-08-07.md`, §1.13.

PR #2166, opened 2026-08-09
fix(chat): pin the answer marker wire position

### LINKED ISSUE #2121: is_answer reads a negative index into a positional wire message — a new TailwindDoc tag silently breaks every chunk
## Summary

`AnswerRow.is_answer` decides whether a streamed chunk is the answer by reading a **negative index** into a positional protobuf-derived message. It works today only because the field it wants happens to be the last one currently registered. The moment the backend adds one more field, the read silently lands on the wrong value and every chunk is classified as "not an answer".

## Evidence (verified against the disassembled schema)

`src/notebooklm/_row_adapters/chat.py:682` sets the marker position to `-1`:

```python
_ANSWER_MARKER_POS: ClassVar[int] = -1
```

and `is_answer` (`src/notebooklm/_row_adapters/chat.py:739-751`) reads `type_block[_ANSWER_MARKER_POS]`, i.e. `first[4][-1]`.

`first[4]` is `AnswerResponse.responseDoc` = `TailwindDoc`, whose `BuilderInfo` registers:

```
body    = 1
(2, 3)  = unused
objects = 4
type    = 5      <-- what [-1] resolves to
```

So `[-1]` reaches `TailwindDoc.type` **only because tag 5 is the highest registered tag today.** Nothing about the read expresses "I want tag 5"; it expresses "I want whatever is last".

## Why this matters

`CLAUDE.md` names undocumented RPC/wire drift as this project's **#1 breakage class** — the obfuscated method IDs and positional shapes are explicitly called out as the things that break when Google changes them. A negative index into a positional wire message is exactly the latent landmine that class of breakage produces:

- The backend adds `TailwindDoc` tag 6 (a routine, backward-compatible protobuf change that breaks nothing for a tag-addressed client).
- `[-1]` now reads the new field instead of `type`.
- `is_answer` goes `False` for **every** chunk.
- The parser falls through to its "No marked answer found" path on every single ask.

There is no error, no exception, and no obviously wrong output — just a silent, total misclassification, surfaced only by a warning that (per the related no-answer misclassification issue) already fires routinely for benign reasons. This is a one-line change to make the failure mode impossible.

## Suggested fix

Replace `_ANSWER_MARKER_POS = -1` with an explicit `4` (the position for tag 5) plus a length guard, so that an unexpected trailing field is ignored rather than silently read as the marker. If a future field genuinely relocates the marker, an explicit index fails visibly instead of quietly inverting the classification.

---

Source: `docs/notes/web-rpc-vs-mobile-grpc-audit-2026-08-07.md`, §1.18.

audichuang:
## Summary

- read the TailwindDoc answer marker from explicit position 4 instead of the last field
- keep malformed and short blocks safely classified as non-answers
- promote the recovered TailwindDoc.type position into the wire-contract guardrail
- prove trailing backend fields cannot change answer classification

Closes #2121

## Test plan

- focused trailing-field regression (red before fix, 2 passed after)
- row adapter and wire contract suites (189 passed, 2 xfailed)
- broad chat suite (490 passed, 1 skipped)
- full guardrails (1276 passed, 4 skipped, 3 xfailed)
- Ruff check and format check (passed)
- `git diff --check` (passed)

## Notes

This PR intentionally does not alter the no-answer fallback or warning behavior from #2118. Mypy is currently blocked by three base-existing unused-ignore diagnostics in `_source/markdown.py`; that file is unchanged by this branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved answer detection by using the correct fixed record position.
  * Prevented trailing data from incorrectly changing answer recognition.

* **Tests**
  * Added regression coverage to verify answer detection remains stable when extra trailing fields are present.

### DIFF
--- a/src/notebooklm/_row_adapters/chat.py
+++ b/src/notebooklm/_row_adapters/chat.py
@@ -660,7 +660,7 @@ class AnswerRow:
     whose ``inner_data`` is a populated list (heartbeats decode to ``[]``
     and never reach this adapter). Position knowledge is centralised here;
     consumer sites should NEVER open-code ``first[0]`` / ``first[2][0]`` /
-    ``first[4][-1]`` / ``first[4][3]``.
+    ``first[4][4]`` / ``first[4][3]``.
 
     The dataclass is frozen so the wrapped row can't be mutated through the
     adapter; the adapter never copies the raw row, so it is cheap to build.
@@ -679,7 +679,7 @@ class AnswerRow:
     _TEXT_POS: ClassVar[int] = 0
     _CONV_BLOCK_POS: ClassVar[int] = 2
     _TYPE_BLOCK_POS: ClassVar[int] = 4
-    _ANSWER_MARKER_POS: ClassVar[int] = -1
+    _ANSWER_MARKER_POS: ClassVar[int] = 4
     _CITATIONS_POS: ClassVar[int] = 3
     _ANSWER_MARKER_VALUE: ClassVar[int] = 1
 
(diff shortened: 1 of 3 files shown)


audichuang:

PR #2167, opened 2026-08-10
fix(sources): fail closed on unknown statuses

### LINKED ISSUE #2124: An unmapped SourceStatus falls back to READY, so unknown backend states are asserted healthy (a source pending deletion reads as ready)
## Summary

`SourceRow.status` (`src/notebooklm/_row_adapters/sources.py:662-673`) catches `ValueError` from `SourceStatus(code)` and returns **`SourceStatus.READY`**. An unrecognised status code therefore does not degrade to "unknown" — it asserts that the source is *healthy and usable*.

This is worth naming precisely, because it is a different failure mode from the more familiar "our enum is missing a value". The bug is not that a value is absent from `SourceStatus`; it is that **an unmapped value silently lands in a specific *wrong* branch instead of an explicit unknown state**. A missing enum member is a data gap; this is a wrong answer. The distinction matters because the mitigation is different: adding the currently-known missing members would fix today's symptoms while leaving the mechanism — and every future backend state — exactly as broken.

## Evidence (live-verified, not inferred)

Verified by executing this client's own adapter directly against synthetic status blocks:

```
raw status 0  -> SourceStatus.READY
raw status 4  -> SourceStatus.READY
raw status 99 -> SourceStatus.READY
```

All three return `is_ready=True`. Code `99` is not a real backend code — it is included to show the fallback is unconditional: *any* integer the enum does not know becomes `READY`.

The mapped enum today (`src/notebooklm/rpc/types.py:437-440`):

```python
PROCESSING = 1  # Source is being processed (indexing content)
READY = 2       # Source is ready for use
ERROR = 3       # Source processing failed
PREPARING = 5   # Source is being prepared/uploaded (pre-processing stage)
```

`0 = UNSPECIFIED` and `4 = PENDING_DELETION` are absent (the sibling gap noted alongside #2111), so both currently resolve to `READY`. **A source queued for deletion reads as ready to use.**

The fallback is deliberate and documented. The docstring at `src/notebooklm/_row_adapters/sources.py:657-660` defends it as forward-compatible:

> The membership check uses `SourceStatus(...)` directly (catching `ValueError`) rather than an explicit member tuple so the adapter automatically accepts any new values added to `SourceStatus` without a parallel update here.

The forward-compatibility goal is sound; the chosen default is not. Automatically accepting new values is fine — resolving all of them to `READY` is what makes it unsafe.

## Why this matters

`READY` is not a neutral placeholder. It is the one value in the enum that makes a positive claim about the source: `is_ready` returns `True`, `wait_until_ready` returns immediately, `source list` renders "ready", and downstream callers proceed to chat/generate against it.

So the fallback asserts health for **every backend state we have not yet mapped, including states that specifically represent danger** — `PENDING_DELETION(4)` is the concrete example available today: a source scheduled for removal is reported as fully usable. Any future state Google adds (quarantined, quota-blocked, re-indexing, access-revoked, …) inherits the same treatment on the day it ships, with no error, no warning, and no log line. The failure is silent by construction.

The safe default is the opposite: an unrecognised code should surface as an explicit unknown that callers can branch on, and `is_ready` should be `False` for it — "we do not know" must not be spelled "yes".

## Caveat

`PENDING_DELETION` was not observable live on this transport: the delete path transitions `status 2 → GONE` inside a 0.6s poll interval, so there is no client-visible window in which the status is 4. The degrade behaviour itself is proven regardless — it was confirmed by direct execution of the adapter (above) and independently re-verified — and it is the part that matters. The specific set of unmapped codes will change over time; the wrong-branch mechanism will not.

## Suggested fix

1. Add an explicit `UNKNOWN` member (or return a sentinel / `None`) for unrecognised codes instead of `READY`, and make `is_ready` `False` for it.
2. Apply the same treatment to the structural fallbacks at `src/notebooklm/_row_adapters/sources.py:662-667` (position 3 absent / non-list / too short), which currently also return `READY`.
3. Log at warning level on an unmapped code so enum drift becomes visible rather than silently absorbed — this is the actual forward-compatibility mechanism the docstring is reaching for.
4. Separately, add the known-missing members `UNSPECIFIED = 0` and `PENDING_DELETION = 4` to `SourceStatus`. This is a follow-on data fix, not the fix for this issue.

---

Ref: `docs/notes/web-rpc-vs-mobile-grpc-audit-2026-08-07.md`, §1.24.

audichuang:
## Summary

`SourceRow.status` fell back to `SourceStatus.READY` for **every** status it could not read — an absent
status block, a malformed one, and an integer code the enum does not know. So an unrecognised backend
state was asserted healthy: per #2124, a source pending deletion reads as ready, and `is_ready` returns
`True` for a row whose real state we never decoded.

This is the wrong direction for a client whose #1 breakage class is undocumented wire change. An
unknown code means "we do not know", and the honest answer is to say so — not to pick the one value
that makes callers proceed. This PR introduces `SourceStatus.UNKNOWN` and fails closed onto it, and
emits a warning for unmapped **integer** codes so enum drift becomes visible instead of silent.

## Related Issue

Closes #2124

## Root cause

One `except ValueError: return READY` was standing in for three different situations, only one of which
is "healthy":

| `_raw[3]` shape | Meaning | Old `status` | New `status` | Warns? |
|---|---|---|---|---|
| absent / non-list / too short | response shape that carries no status block | `READY` ❌ | `UNKNOWN` | no |
| present, code is a known enum value | normal | that value | that value (unchanged) | no |
| present, code is an **unmapped int** | backend enum drift | `READY` ❌ | `UNKNOWN` | **yes** |
| present, code is a non-int (`str`, `None`, `bool`, …) | malformed | `READY` ❌ | `UNKNOWN` | no |

The split between the warning and the silent arms is deliberate: several legitimate non-listing
response shapes omit the status block entirely, so warning on a structurally-absent block would
reproduce exactly the every-call noise that #2131 just fixed for `get_source_ids`. An unmapped
*integer*, by contrast, is a code the backend actually sent and we do not model — that is the drift
worth a line in the log.

## Changes

- `src/notebooklm/rpc/types.py` — `SourceStatus.UNKNOWN = -1` plus its `_SOURCE_STATUS_MAP` entry
  (`"unknown"`). `source_status_to_str` already returned `"unknown"` for unrecognised codes, so the
  string surface is unchanged; the enum can now *represent* the state it was already naming.
- `src/notebooklm/_row_adapters/sources.py` — `SourceRow.status` returns `UNKNOWN` on all three
  unreadable shapes, rejects `bool` explicitly (`True` would otherwise decode as `PROCESSING`), and
  logs `"Unknown source status code %r from RPC %s"` **once per code** for an unmapped int. The
  docstring now states the fail-closed contract and why the malformed arm is silent.
- `src/notebooklm/mcp/tools/sources.py` — `source_list`'s `status` filter `Literal` gains `"unknown"`,
  so an operator can actually find the rows this change surfaces.
- `src/notebooklm/_app/views.py`, `src/notebooklm/_types/sources.py` — docstrings corrected: they
  claimed the label set excluded `unknown` and that the fallback was `READY`.
- `docs/python-api.md` — the `Source.status` field is documented as `SourceStatus` (it was annotated
  `int` with a `"defaults to READY"` comment), and `SourceStatus` gains its `UNKNOWN` member.
- `docs/mcp-guide.md` — the `source_list` status-filter value list.
- Guardrails/fixtures: `tests/_guardrails/_wire_contract.py`,
  `tests/_guardrails/test_studio_enum_manifest.py`, `tests/fixtures/rpc_golden/ADD_SOURCE.json`.
- Tests: `tests/unit/test_row_adapters.py` (the four rows of the table above, plus the warn-once
  behaviour and an autouse reset of the module-level warned-code set),
  `tests/unit/test_rpc_types.py`, `tests/unit/test_types.py`,
  `tests/unit/test_source_listing_service.py`, `tests/unit/test_rpc_golden_payloads.py`,
  `tests/integration/test_sources_integration.py`.

## Test Plan

- [x] Tests pass (`uv run pytest --cov=src/notebooklm --cov-report=term-missing --cov-fail-under=90`)
- [x] If this PR changes architectural shape, an ADR has been added or updated — N/A, no shape change

Run against the canonical contributor install **plus the extras the CI test matrix adds**
(`--extra browser --extra dev --extra markdown --extra mcp --extra server --extra impersonate`).
The lean `browser+dev+markdown` install alone skips ~1,500 tests and lands at 83% coverage, i.e.
under the gate — so the suite below is the full-extras run that matches CI:

```console
$ uv run mypy src/notebooklm --ignore-missing-imports
Success: no issues found in 351 source files

$ uv run pre-commit run --all-files
ruff.....................................................................Passed
ruff-format..............................................................Passed

$ uv run pytest --cov=src/notebooklm --cov-report=term-missing --cov-fail-under=90 -q
TOTAL                                                    30941    826   8128    507    96%
Required test coverage of 90% reached. Total coverage: 96.41%
15225 passed, 65 skipped, 3 xfailed, 116 warnings in 580.23s (0:09:40)
```

I also ran every remaining gate from the `Code Quality` job of `.github/workflows/test.yml`, all green:
`check_workflow_permissions`, `check_workflow_secret_gates`, `check_action_pinning`,
`check_deprecation_targets`, `check_coverage_thresholds`, `check_ci_install_parity`,
`check_claude_md_freshness`, `check_docs_module_refs`,
`audit_public_api_compat.py --check-stale`, `tests/scripts/check_method_coverage.py`,
`pytest tests/e2e --collect-only`, and the lean-install core-import assertion
(`import notebooklm, notebooklm.notebooklm_cli, notebooklm.client, notebooklm.artifacts`).

## Notes

**No recorded live path flips from ready to unknown.** In principle any response shape that omits the
status block moves `Source.status` `READY -> UNKNOWN` and `Source.is_ready` `True -> False`. I went
looking for a real one and did not find it: every recorded `ADD_SOURCE` / `UPDATE_SOURCE` reply in the
cassettes carries the `[null, 2]` status block — `sources_add_url.yaml:1589`,
`sources_add_text.yaml:1589`, `sources_add_drive.yaml:105`, `sources_rename.yaml:1691` — as does every
listing row. So the shapes that now read `UNKNOWN` are the malformed and unmapped ones, which is the
point of the issue, and a single `### Fixed` entry is proportionate rather than a `### Changed` with a
migration note.

One caveat on reading the diff: `tests/fixtures/rpc_golden/ADD_SOURCE.json` now carries
`"status": -1`, but that fixture's payload is a synthetic id-only row, **not** the shape the live
endpoint returns. It should not be read as "real adds now decode UNKNOWN".

**`UNKNOWN = -1`, not `0`.** `0` is a value the wire could plausibly start sending for a real state;
`-1` cannot collide with a protobuf enum ordinal, so a future backend code can be mapped without
having to relocate `UNKNOWN` first.

**`bool` is rejected before the enum lookup.** `bool` is a subclass of `int`, so `SourceStatus(True)`
would resolve to `PROCESSING`. The explicit `isinstance(status_code, bool)` rejection keeps a wire
`true` from decoding as a real status.

**Two things this branch got wrong on its first pass, both fixed here.**

*Dead code.* The earlier version caught `(TypeError, ValueError)` and re-tested
`isinstance(status_code, int)` inside the handler. Neither is reachable: the `isinstance` guard three
lines above has already established the value is a non-`bool` `int`, and `SourceStatus(<plain int>)`
raises only `ValueError` — verified across `-1, 0, 4, 6, 99, -2, 2**63`, every one of which raises
`ValueError`, never `TypeError`. Now a plain `except ValueError`, same behaviour, fewer lines.

*The warning had no dedupe.* `SourceRow.status` is re-decoded on every row of every poll, and
`_source/polling.py` runs ~17 polls per `wait_until_ready` (timeout 120s, 1.0s initial, 1.5× backoff,
10s cap). A single source carrying status `0` or `4` — **both declared real backend members** in the
`ENUM_GAPS` block of `tests/_guardrails/_wire_contract.py` — would therefore have emitted the identical
WARNING ~17 times per wait. It now warns once per code, using the same `_warned_status_codes` set idiom
as the sibling unmapped-*type* path (`_types/sources.py::_warned_source_types`).
`test_unknown_status_warns_once_per_code` pins it, and because that set is module-level an autouse
fixture clears it around the class so the drift assertions cannot depend on test order — the same
hazard `tests/unit/test_types.py` already handles for `_warned_source_types`.

**Scope of verification.** Offline: unit tests, the golden payload fixture, and the integration
cassette path. The specific live case in #2124 (a source pending deletion reporting ready) is the
maintainer's observation, not an independent run — I have no account positioned to reproduce a
pending-deletion row. What I verified is that each of the four shapes in the table now resolves as
stated, and that the tests fail on the old fallback.

**Not verified locally:** the Python 3.10–3.14 matrix. No version-sensitive syntax is used.

## Summary by CodeRabbit

* **New Features**
  * Added an `UNKNOWN` source status for absent, malformed, or unrecognized status values.
  * Added `unknown` as a supported filter for source listings.
  * Unrecognized integer statuses now generate a warning.

* **Bug Fixes**
  * Sources with unknown or missing statuses are no longer incorrectly marked as ready.
  * Updated status documentation and API references to reflect the new behavior.

* **Documentation**
  * Added release notes describing unknown-status handling and warnings.


audichuang:

teng-lin:
@codex review
@claude review

Issue #2141, opened 2026-08-07
Deep-research citation ordinals (src[8]) are discarded, leaving 111 [cite: N] markers unresolvable


teng-lin:
## Summary

The deep report we hand callers contains **111 `[cite: N]` markers**, and the integers that resolve them are in the same response — at `src[8]`, which `ResearchResultRow` has no reader for.

## Evidence (live)

From a completed deep run (35,773-char report, 63 source rows):

- **111** `[cite: N]` occurrences in the report text, `N` spanning **1–41**
- exactly **41** rows carry an integer ordinal at `src[8]`, range **1–41**
- those are precisely the 41 rows with `src[6][1] == 1`

The mapping is one-to-one and already parsed into memory. `_row_adapters/research.py:239-242` declares positions up to 6, so `src[8]` is never read.

## Why this matters

Callers receive a report peppered with `[cite: 12]` and no way to resolve any marker to a source — while the resolution table sits in the response we already decoded.

It is also the practical reason `select_cited_sources` (`research.py:97`) has to regex-scrape URLs out of the markdown. That mostly works (the report carries 82 bare URLs and 41 markdown links), but it is a heuristic standing in for an authoritative linkage we discarded.

Deep-only: fast rows are 4 slots with no ordinals and no report.

## Related, same row

`src[6][5]` on the report row is a structured document tree of `[start, end, [...]]` spans over the report text — the formatted/annotated rendering:

```
[[[[0, 95, [[[0, 95, ["Biophysical and Molecular…"]]], [null, 4]]], [95, 172, …
```

Also unread. Flagging it as present rather than proposing we consume it.

## Same defect class as the chat path

This is the research-path instance of #2120 — an authoritative citation→source linkage present on the wire and dropped, leaving downstream code to reconstruct it heuristically.

---

Evidence: one completed live deep run, read-only analysis of the captured payload.

PR #2169, opened 2026-08-10
fix(research): identify report rows by content kind

### LINKED ISSUE #2140: Deep-research report extraction is order-dependent: src[6] is a typed content block, not report chunks
## Summary

`ResearchResultRow._LEGACY_CHUNKS_POS = 6` treats `src[6]` as "legacy deep-research report chunks". It isn't. On deep runs `src[6]` is a **structured content block populated on every row**, and the report is extracted correctly today only because the report row happens to arrive first.

**Severity: latent, not currently broken.** Confirmed live: on the payload as served, parsing is correct. But 62 of 62 web rows carry a populated `src[6]`, so every one is defused solely by arrival order — which is not a documented property and not ours to rely on.

## The real shape (recovered from two live captures, 14 months apart)

```
src[6] = [ text|None, kind:int, text|None, None, int|None, structured_doc|None ]

report row : ["# Biophysical and Molecular…", 3, None, None, None, [doc tree]]   len 6
web row    : [None,                           1, "Tardigrades can survive…", None, 13]  len 5
```

`src[6][1]` is the discriminator — **3 = report, 1/2 = web snippet**. Live distribution across 63 rows: `{1: 41, 2: 21, 3: 1}`. The 2026-06 cassette's report row has the byte-identical 6-slot shape, so this is stable.

`docs/rpc-reference.md:2168-2172` documents `src[6]` as `[chunk1, chunk2, ...]`, a list of report chunks. That is wrong and should be corrected alongside the code.

## Why it survives today — two stacked accidents

`extract_legacy_report_chunks` (`_research_task_parser.py:52-59`) never reads chunks. It harvests **every string** in the block and joins them, filtering on `isinstance(chunk, str)`. That works only because each row type happens to contain exactly one string — the report's at `[0]`, the snippet's at `[2]`.

Then `report_found` shields subsequent rows, so the *first* row carrying a string wins. Ordering is the second accident stacked on the first.

## Reproduction

Same 63 live rows, report row moved from first to last, nothing else changed:

- `task.report` becomes **4,020 chars** of a tardigrade blurb from the first web result
- the genuine **35,773-char report is silently dropped**
- one web source is handed a fabricated `report_markdown`
- no error, no warning

A deep run that returns **no** report row is worse: it does not degrade to "no report", it fabricates one from a web snippet and feeds it to `select_cited_sources` (`research.py:97`), which finds no URLs and falls back to importing everything — so `--cited-only` silently stops meaning anything.

## Suggested fix

Gate on `src[6][1] == 3` and read `src[6][0]`, instead of "join whatever strings are in slot 6, for the first row that has any". The discriminator is present on every row, so this is a small, principled change.

Note `ResearchResultRow.deep_payload` (`_row_adapters/research.py:249-251`, parser branch `_research_task_parser.py:229-248`) expects `src[1] = [title, report_md]`. **That shape has never appeared in any captured payload** — both real deep captures use `src[1] = title` plus the `src[6]` block, and every test of the `[title, report_md]` branch is hand-written. It may be dead code; worth deciding deliberately while fixing this.

---

Evidence: 2 live deep runs (374 s and 358 s) plus 2 cassettes. Index 6 is recorded with this shape in `tests/_guardrails/_wire_contract.py`.

audichuang:
## Summary

- model `src[6]` as a typed content block instead of a list of report chunks
- extract report markdown only from exact integer kind 3 at position 1 and text at position 0
- prevent web snippets, malformed kinds, and row ordering from fabricating or suppressing reports
- preserve the existing `src[1]` compatibility path and leave citation ordinals for #2141

Closes #2140

## Test plan

- focused parser, adapter, and wire-contract suites (203 passed, 2 xfailed)
- broader research, CLI, MCP, server, and deep cassette suites (497 passed, 2 xfailed)
- full pytest (13718 passed, 77 skipped, 3 xfailed)
- post-review malformed-kind focused suite (263 passed)
- mypy over 334 source files (passed)
- Ruff changed-file check and format check (passed)
- `git diff --check` (passed)

## Evidence scope

The discriminator is backed by the two captures described in #2140. This fixes a latent order dependency; it does not claim row reordering has been observed. No `src[8]` citation work is included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deep-research report extraction from current captured response formats.
  * Ensured report content is selected correctly when web results appear first.
  * Prevented non-report content or malformed data from being treated as reports.
  * Improved handling when no valid report is available.

* **Documentation**
  * Updated deep-research source documentation to reflect current report and web-content formats.

* **Tests**
  * Added coverage for report identification, markdown extraction, malformed content, and fallback behavior.

### DIFF
--- a/docs/rpc-reference.md
+++ b/docs/rpc-reference.md
@@ -2165,11 +2165,15 @@ await rpc_call(
 # Fast research web source:
 # [url, title, desc, type, ...]
 #
-# Deep research report source (current shape):
-# [None, [title, report_markdown], None, type, ...]
+# Deep research report source (captured shape):
+# [None, title, None, type, None, None,
+#  [report_markdown, 3, None, None, None, structured_document]]
 #
-# Deep research report source (legacy shape):
-# [None, title, None, type, None, None, [chunk1, chunk2, ...]]
+# Deep research web source content blocks use kind 1 or 2 and carry their
+# snippet at position 2; they are not report rows.
+#
+# Compatibility shape accepted by the parser (not seen in captures):
+# [None, [title, report_markdown], None, type, ...]
 #
 # Notes:
 # - The RPC returns all research tasks for the notebook, not just the latest one.
--- a/src/notebooklm/_research_task_parser.py
+++ b/src/notebooklm/_research_task_parser.py
@@ -49,14 +49,9 @@ _POLL_SOURCE = "_research.poll"
 _POLL_METHOD_ID = RPCMethod.POLL_RESEARCH.value
 
 
-def extract_legacy_report_chunks(src: list[Any]) -> str:
-    """Join legacy deep-research report chunks stored in ``src[6]``."""
-    chunks = [
-        chunk
-        for chunk in ResearchResultRow(src).legacy_report_chunks
-        if isinstance(chunk, str) and chunk
-    ]
-    return "\n\n".join(chunks)
+def extract_report_markdown(src: list[Any]) -> str:
+    """Return markdown from the kind-3 content block stored in ``src[6]``."""
+    return ResearchResultRow(src).report_markdown
 
 
 def _extract_task_id(task_data: Any) -> str | None:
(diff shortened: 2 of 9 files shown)


audichuang:

PR #2170, opened 2026-08-10
fix(sources): surface partial file upload failures

### LINKED ISSUE #2138: Orphaned sources: a post-transfer processing failure surfaces nothing, and a retained PREPARING row is invisible to source_list
> **Re-scoped 2026-08-13.** #2179 closed the `type_code = 0` spurious-warning half and made the *upload-rejection* route recoverable. This issue now tracks only what is left. See the [re-scope comment](https://github.com/teng-lin/notebooklm-py/issues/2138#issuecomment-5279313545) for exactly what shipped.

## Summary

Two gaps remain from the original orphaned-source report:

1. **The successful-transfer / processing-error route surfaces nothing.** When bytes transfer fine and processing *later* fails, `add_file` returns normally — no exception is raised, so #2179's recovery context (`source_id` / `stage`, attached only on a post-registration *upload* failure) never attaches. The caller is told the add succeeded.
2. **A retained orphan in `PREPARING` is invisible to the listing a caller would check.** #2179 deliberately does not delete the registered row — the right call, it preserves evidence — but the row sits in `PREPARING`, not `ERROR`, so `source_list(status="error")` does not surface it. There is no reconciliation path from "an add failed" to "here are the rows it left behind".

## Evidence (live-verified, carried over from the original report)

The `.wav` route, which is gap 1:

```
.wav upload -> bytes transferred successfully
            -> processing ended at status = ERROR, type_code = 0
```

(WAV is not a supported audio container; `.mp3` and `.m4a` both are.) `add_file` had already returned by then.

The `.xlsx` route, which is gap 2 — two attempts against a scratch notebook, the second supplying `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet` explicitly, so this is a genuine server-side format rejection rather than a MIME-inference miss:

```
ValidationError: NotebookLM rejected the upload ... (HTTP 400)
```

`GET_NOTEBOOK` afterwards still lists the row:

```
title     = "probe_excel.xlsx"
type_code = 0            -> SourceType.UNKNOWN   (now silent, per #2179)
status    = PREPARING    -> never resolves
```

The row counts toward the notebook's source quota. As of #2179 a caller *can* name it — `getattr(exc, "source_id", None)` on the raised exception — but only on that route, and only if they catch and act at the call site. Nothing reconciles it afterwards.

## Suggested scope

- **Gap 1:** decide where a post-transfer processing failure is reported. `add_file` cannot raise for it (it has returned), so this is either a documented "you must poll" contract, or surfacing it through the existing registration/readiness poll (`sources.wait_until_registered` / `wait_until_ready` in `src/notebooklm/_sources.py`), or both. Today the failure is silent at the API boundary.
- **Gap 2:** give a caller a way to find rows left behind — e.g. let `source_list` reach `PREPARING`-stuck rows for reconciliation, or document the `getattr(exc, "source_id", None)` recovery as the only supported route and say plainly that no listing filter will show the orphan. The opt-in cleanup posture discussed in #2110 still applies: do not auto-delete.

## Caveat

Both orphans were observed on a scratch notebook that was subsequently deleted, so the rows are gone. The reproduction is deterministic (two independent `.xlsx` attempts, plus the `.wav` variant), but no long-lived orphan was left in place to inspect further.

---

Source: `docs/notes/web-rpc-vs-mobile-grpc-audit-2026-08-07.md`. Related: #2111, #2110 (ghost rows via the RPC path). #2137 (the code-6 sibling of the type-map gap) is closed. Partially addressed by #2179.

audichuang:
## Summary

- add public `SourceAddPartialError` with the retained `source_id`, failure `stage`, and chained original cause after registration succeeds
- preserve cancellation and file-descriptor ownership semantics, and never auto-delete the registered row
- classify typed infrastructure causes consistently through app, MCP, and REST adapters
- treat wire source type code `0` as expected `UNKNOWN` without an upgrade warning
- document explicit recovery and cover session-start, upload/finalize, registration, cancellation, FD, export, and warning behavior

Fixes #2138

## Checks

- `uv run pytest -n 4 --dist=worksteal -q` — 13,733 passed, 77 skipped, 3 xfailed
- focused source/adapters — 857 passed
- guard/public surface — 1,668 passed, 3 xfailed
- `uv run mypy src/notebooklm` — success (334 files)
- `uv run ruff check .` — passed
- changed Python files: `uv run ruff format --check ...` — passed
- `git diff --check` — passed

## Summary by CodeRabbit

- **New Features**
  - Added `SourceAddPartialError` for upload failures occurring after source registration.
  - Errors now include the retained source ID, failure stage, filename, and original cause.
  - Exposed the new exception through the public package interfaces.
  - Added handling for unknown source types.

- **Bug Fixes**
  - Preserved accurate error categories, retryability, and cancellation behavior for partial uploads.
  - Clarified that retained source records are not deleted automatically.

- **Documentation**
  - Documented partial upload recovery and source reconciliation guidance.


audichuang:

Issue #239, opened 2026-04-02
Silently failing when reached your daily Cinematics limit


TimPickup:
Hi,

I presume this is the same on all of the daily limits, but currently tasks silently fail when you reach your daily Cinematics limit. It just keeps checking if the task was completed until the timeout.

It would be great if it could capture the fact you hit a limit - or at least recognise it has failed.

Cinematics limit is very low so an easy one to test. Generate a video of format cinematic.

PR #2190, opened 2026-08-12
fix(artifacts): delete the phantom failure-reason read on indices 3/5 (#2134)

audichuang:
## Summary

Delete the artifact row's failure-reason read. `_ERROR_TEXT_POS = 3` and `_ERROR_PAYLOAD_POS = 5` named fields that do not exist, so `failed_error_text` could not answer for any schema-conformant row while its docstring promised a human-readable explanation.

Refs #2134.

## Evidence

| Claim | Provenance |
|---|---|
| Index 3 is `Artifact.sources` (repeated message), index 5 is `isPubliclyReadable` (bool) | **Recovered schema** (`docs/mobile/schema.proto`), cross-checked in #2134 against gRPC on the same artifact |
| `failed_error_text` returned `None` on real failed artifacts | **Live, reported in #2134**: 3/3 |
| The pre-removal poll path really did read those slots | **Measured on this branch**: replayed against `e55de0af`, the same row reports `read_indices [0, 4, 2, 3, 5]` |
| `GenerationStatus.error` has other producers | **In-source**: `wait_for_completion`'s `REMOVED` message and the retry-callback status in `artifacts.py` — both client-authored, neither reads the wire |

## What changed

Removed: the property, both constants, the module-level `_extract_artifact_error`, its `ArtifactsAPI` static hop, the injected `extract_artifact_error` seam on the *private* `ArtifactPollingService.poll_status` (`ArtifactsAPI.poll_status`'s own signature is unchanged), the two `known_bad` wire-contract entries that recorded the gap, and the row-index table lines that mislabelled 3/5.

**Why `Changed` and not `Fixed`:** for conforming rows the output is identical, but a *nonconforming* row carrying a string in slot 3 or 5 no longer populates the public `GenerationStatus.error`. That is a behaviour change, small as it is, so it is priced into the CHANGELOG as one.

## On the tests

The unit tests that kept the wrong model green are **gone rather than adjusted** — they constructed rows with error text at index 3/5, a shape the backend has never emitted, and would have failed a correct implementation.

Their replacement deliberately does **not** just assert `error is None`: the deleted read also returned `None` for a list-at-3 / bool-at-5 row, so that assertion would have been a tautology dressed as evidence. Instead the row records which top-level indices the poll path reads and the test pins that 3 and 5 are not among them — which does fail against the pre-removal code (`read_indices [0, 4, 2, 3, 5]`).

## Scope boundary / Not done

- **No provider failure reason is captured.** A real one would have to come from the `CreateArtifact` RPC status at generation time (`QuotaFailure` / `ErrorInfo` / `PreconditionFailure` are delivered on the RPC, not persisted on the artifact resource). That is a separate change, and this PR supplies no captured status detail for it.
- `docs/notes/web-rpc-vs-mobile-grpc-audit-2026-08-07.md` §1.17 still carries the original recommendation; I left the audit note as the dated record it is.

## Verification

```
uv run ruff format --check <changed files>   # clean
uv run ruff check .                          # clean
uv run mypy src/notebooklm --ignore-missing-imports   # clean, 351 files
uv run pytest                                # 15413 passed, 64 skipped, 1 xfailed
```

Wire contract: mappings 38 → 36, with zero `known_bad` entries remaining.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected failed-artifact status handling so nonexistent artifact fields no longer produce misleading error messages.
  * Preserved client-provided and retry-generated errors.
  * Failed artifacts now correctly report their status without incorrectly reading unrelated artifact metadata.
  * Artifact source and visibility information is no longer mistaken for an error message.
  * Status and artifact URL handling remain unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and `artifact_status_to_str` helpers.
   ([#2125](https://github.com/teng-lin/notebooklm-py/issues/2125))
 
+### Changed
+
+- **Failed artifact polling no longer derives `GenerationStatus.error` from
+  nonexistent artifact-row fields.** `poll_status()` treated row index 3 as
+  "failure text" and index 5 as a "nested error payload"; those slots are
+  `Artifact.sources` and `Artifact.isPubliclyReadable`, so the read could only
+  ever produce a wrong answer. For schema-conformant rows the observed output is
+  unchanged — #2134 reports `error=None` for 3/3 real failed artifacts — but a
+  nonconforming row carrying a string in either slot no longer populates the
+  public field. `wait_for_completion()` still sets its own client-authored
+  message for `REMOVED`, and `generate_*(retry=...)` still synthesizes
+  `error=str(exc)` for retry callbacks; neither reads the wire for it. Capturing
+  a genuine provider failure reason (delivered on the `CreateArtifact` RPC
+  status, not persisted on the artifact) is not implemented.
+  ([#2134](https://github.com/teng-lin/notebooklm-py/issues/2134))
+
 ### Fixed
 
 - **`Notebook.is_owner` no longer inverts on shared notebooks, and the notebook's
--- a/src/notebooklm/_artifact/polling.py
+++ b/src/notebooklm/_artifact/polling.py
@@ -38,7 +38,6 @@ ListRawCallback = Callable[[str], Awaitable[builtins.list[Any]]]
 PollStatusCallback = Callable[[str, str], Awaitable[GenerationStatus]]
 MediaReadyCallback = Callable[[builtins.list[Any], int], bool]
 ArtifactTypeNameCallback = Callable[[int], str]
-ArtifactErrorCallback = Callable[[builtins.list[Any]], str | None]
 StatusChangeCallback = Callable[[GenerationStatus], object]
 
 
@@ -103,7 +102,6 @@ class ArtifactPollingService:
         list_raw: ListRawCallback,
         is_media_ready: MediaReadyCallback,
         get_artifact_type_name: ArtifactTypeNameCallback,
-        extract_artifact_error: ArtifactErrorCallback,
     ) -> GenerationStatus:
         """Poll the status of a generation task."""
         # List all artifacts and find by ID (no poll-by-ID RPC exists).
@@ -136,18 +134,12 @@ class ArtifactPollingService:
                     # Downgrade to PROCESSING to continue polling.
                     status_code = ArtifactStatus.PROCESSING
 
-            status = artifact_status_to_str(status_code)
-
-            error_msg: str | None = None
-            if status == "failed":
-                error_msg = extract_artifact_error(row.raw)
             url = row.artifact_url(artifact_type, suppress_drift=True)
 
             return GenerationStatus(
                 task_id=task_id,
                 status=_status_from_code(status_code),
                 url=url,
-                error=error_msg,
                 metadata=metadata,
             )
 
@@ -480,22 +472,6 @@ def _artifact_timeout_error(
         status_history=history,
         status_transitions=transitions,
     )
-
-
-def _extract_artifact_error(art: builtins.list[Any]) -> str | None:
-    """Try to extract a human-readable error from a failed artifact."""
-    try:
-        if not isinstance(art, list):
-            return None
-        return ArtifactRow(art).failed_error_text
-    except Exception:
-        preview = art[:6] if isinstance(art, list) else art
-        logger.warning(
-            "Failed to extract error from artifact data: %r",
-            preview,
-            exc_info=True,
-        )
-        return None
 
 
 def _get_artifact_type_name(artifact_type: int) -> str:
--- a/src/notebooklm/_artifacts.py
+++ b/src/notebooklm/_artifacts.py
@@ -734,7 +734,6 @@ class ArtifactsAPI:
             list_raw=self._list_raw,
             is_media_ready=self._is_media_ready,
             get_artifact_type_name=self._get_artifact_type_name,
-            extract_artifact_error=self._extract_artifact_error,
         )
 
     async def wait_for_completion(
@@ -969,17 +968,6 @@ class ArtifactsAPI:
         the implementation lives on :class:`ArtifactGenerationService`.
         """
         return self._generation._parse_generation_result(result, method_id=method_id, source=source)
-
-    @staticmethod
-    def _extract_artifact_error(art: builtins.list[Any]) -> str | None:
-        """Extract a human-readable error from a failed artifact, or ``None``.
-
-        Google's batchexecute responses embed error info in varying positions;
-        this walks known (reverse-engineered) locations and returns the first
-        non-empty string. ``art[3]`` sometimes holds an error reason string;
-        ``art[5]`` may hold a nested UserDisplayableError-style payload.
-        """
-        return _artifact_polling._extract_artifact_error(art)
 
     def _get_artifact_type_name(self, artifact_type: int) -> str:
         """Human-readable name for an ``ArtifactTypeCode``, else the raw int as str."""
--- a/src/notebooklm/_row_adapters/artifacts.py
+++ b/src/notebooklm/_row_adapters/artifacts.py
@@ -114,9 +114,9 @@ class ArtifactRow:
     0      artifact id (str)
     1      artifact title (str)
     2      type code (int — see :class:`notebooklm.rpc.ArtifactTypeCode`)
-    3      failed-artifact plain error text (when present)
+    3      sources (repeated message — unread; see #2134)
     4      processing status (int — see :class:`notebooklm.rpc.ArtifactStatus`)
-    5      failed-artifact nested error payload (when present)
+    5      isPubliclyReadable (bool — unread; see #2134)
     6      audio metadata; ``[6][5]`` is the audio media list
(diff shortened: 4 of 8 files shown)


teng-lin (commented):
Reviewed the full diff against `e55de0af`, and re-derived the evidence independently rather than taking the issue's numbers on trust. **This is a good removal and I think it should merge.** The findings below are one docstring correction I'd like to see in this PR, two cheap fixture improvements, and two things that are out of scope but which this PR makes provable — plus receipts for the "was anything load-bearing?" question, so a future reader doesn't have to re-derive them.

## What's strong

`test_poll_status_never_reads_the_non_error_slots` (`tests/unit/test_generation_state.py:305-347`) is the right test and better than what it replaces. Recording the indices the poll path actually touches, instead of asserting `error is None`, is exactly the distinction that matters here — as your PR description says, `error is None` would have passed against the pre-removal code too, so it would have been a tautology dressed as evidence. This one genuinely fails against `e55de0af`.

The docstring is the part I'd most like to see copied elsewhere in the repo:

> Evidence for "the resource carries no reason" is external: live, reported in #2134 as 3/3 failed artifacts returning `None`. This row is constructed.

Separating "what the wire does" from "what this fixture asserts" is precisely the discipline whose absence caused #2134. Likewise, classifying the CHANGELOG entry as **Changed** rather than **Fixed**, and pricing in the nonconforming-row behaviour difference, is more honest than the change strictly required.

## Receipts for "was anything load-bearing?" — no, and here's why

Verified against your diff, not inferred:

- `isinstance(direct, str)` at index 3 could never fire on a conforming row. Sweeping every `gArtLc` response in `tests/cassettes/` (33 files, **301 rows**): index 3 is a `list` **301/301**, index 5 is `None` **301/301**. Statuses across all 301 are only `{1: 25, 3: 276}` — **no FAILED row exists in any cassette at all**, which is worth knowing on its own.
- `docs/mobile/schema.proto:152-167` — `Artifact` has *no* failure or error field anywhere in the message, not merely "not at 3 or 5". So the deleted read had no correct version.
- The `except Exception` you removed from `_extract_artifact_error` was unreachable. By the time it ran, `find_artifact_row_by_id` had already constructed the same `ArtifactRow` and `poll_status` had already read `row.status` / `row.type_code` off it, so a construction failure raised earlier, outside the catch; `failed_error_text` itself was pure `len` / index / `isinstance` over a list. Deleting it removes a would-be silent-failure site rather than losing behaviour.
- No other caller: `failed_error_text` had exactly one consumer (`_extract_artifact_error`), which had exactly one (`poll_status`).

## 1. `tests/unit/test_quota_failure_detection.py:6-13` — stale claim left behind

This is my one substantive ask, and it's cheap. The module docstring is untouched by the diff and still reads:

```
Root causes:
...
3. Failed artifacts had no error message surfaced to the caller.
```

After this PR that file no longer tests root cause 3 — `TestPollStatusErrorExtraction` and `TestExtractArtifactError` are both gone — and the sentence implies a fix that, per the proto, cannot exist on the artifact resource. Someone auditing #239 coverage will land on this docstring and conclude the guarantee is covered.

That is the same defect class this PR is fixing: a test artifact asserting something the code does not support. Suggest rewriting item 3 along the lines of:

```
3. A synchronous quota rejection was not surfaced with a helpful message. It
   arrives on the CreateArtifact RPC status, not on the artifact resource — the
   listing row has no failure field (#2134), so poll_status reports
   status="failed" with error=None.
```

## 2. `tests/_guardrails/_wire_contract.py:182` — the registry loses the negative knowledge

Deleting both `Mapping` entries is correct: the constants are gone, so leaving them would trip `test_no_stale_registry_entries`. **And I verified check B still holds** — `test_every_adapter_constant_is_declared` scans `_*_POS` constants in `src/notebooklm/_row_adapters/` and requires each in `MAPPINGS ∪ UNMAPPED ∪ PINNED`, so a new `_ERROR_TEXT_POS = 3` cannot land unregistered. That property is intact.

What's lost is documentation, not enforcement: this file is the one place the repo keeps "what does index N point at" on the record, and after this PR it says nothing about 3 or 5. The next person reading `ArtifactRow` for `sources` gets the adapter docstring (good) but the registry — the authority — is silent. A comment where the mappings were would cost nothing:

```python
# Artifact indices 3 (`sources`, tag 4) and 5 (`isPubliclyReadable`, tag 6) have
# no constant here on purpose: `ArtifactRow` no longer reads them. They were
# `_ERROR_TEXT_POS` / `_ERROR_PAYLOAD_POS`, two known_bad mappings naming fields
# that have never existed; both were removed in #2134. A failed artifact carries
# no reason on the listing row.
```

Related and pre-existing, so entirely your call whether to touch it: the module docstring at `_wire_contract.py:9-13` points at "two that named fields which have never existed (see ``ARTIFACT_ROW`` below)". There is no `ARTIFACT_ROW` symbol in the file today either, so the pointer was already stale — but after this PR there is nothing below at all, which makes it more confusing than it was.

## 3. Fixtures still default index 3 to `None` where the wire says `list`

Minor, but it's the PR's own lesson applied one step further:

- `tests/unit/test_quota_failure_detection.py:60-62` — `_art(..., sources: list | None = None)` now names the field correctly, but every row it builds still carries `None` at index 3. On the wire that slot is a list 301/301.
- `tests/unit/test_generation_state.py:272-273` — the comment was updated to `sources` but the value stays `None`.

Defaulting to `[[["src-uuid"]]]` in both would make every row these helpers produce the shape the backend actually emits, at the cost of one literal. Your new `_RecordingRow` test already does this correctly.

## 4. Out of scope, but this PR makes them provable — flagging so they aren't lost

Neither belongs in this PR; I've written both up in **#2193** so they have a home.

**a. `notebooklm artifact wait` now provably exits 0 on a terminal FAILED artifact.** `src/notebooklm/cli/artifact_cmd.py:610-618`:

```python
elif status.error:
    console.print(f"[red]✗ Generation failed:[/red] {status.error}")
    exit_with_code(1)
else:
    console.print(f"[yellow]Status:[/yellow] {status.status}")   # ← no exit_with_code
```

Once `poll_status` provably never sets `.error`, that `elif` is dead for `status="failed"`: the failure prints in *yellow* and the command returns 0, so `notebooklm artifact wait <id> && ./publish.sh` publishes a failed generation. The sibling already handles it — `artifact retry --wait` (`artifact_cmd.py:735-743`) has the identical `else` with `exit_with_code(1)` and a comment naming this exact case ("`failed` with no extractable error") — and `artifact wait`'s own `--json` branch exits 1 via `if not status.is_complete`. Pre-existing in real-world behaviour, since `error` was already always `None` live; this PR is what turns it from latent into structural.

**b. `src/notebooklm/mcp/tools/studio.py:513-515`** still advertises *"Returns `status` / `url` / `error` / `is_complete` / `media_ready`"*. An agent will read `error`, find `null`, and relay a bare "generation failed" with no next step. Worth a sentence pointing at `artifact retry`, in the same spirit as item 1.

## On the tests, overall

The remaining assertions are hand-built rather than captured — unavoidably so, since there is no captured FAILED row anywhere in the suite to assert against. What makes that acceptable here is that the new test says so in its own docstring and pins a *structural* property (which indices are read) rather than a value the fixture itself supplies. That's the right pattern under the constraint.

The one thing genuinely left uncovered is #239 itself: the guarantee is upheld by the synchronous `CreateArtifact` rejection path (`extract_rpc_result` → `RateLimitError`, `src/notebooklm/rpc/decoder.py:629-641`), and nothing tests that it reaches a `generate_*` caller. Your "Scope boundary / Not done" section is right that capturing a *provider* reason is separate work — I'd only add that re-guarding the path that already works is not, and it's a dozen lines. That's #2193, with a working starting point in it; happy for it to be a follow-up rather than a condition on this PR.

Nice work — the `_RecordingRow` approach in particular is a better test than the one I'd have written.

teng-lin:
Thank you for this one, @audichuang — of the three, this is the best-evidenced, and I merged nothing on faith. I live-probed it before signing off.

**Live validation (42 artifact rows across 17 notebooks, real account):**

```
status codes (row[4]) : {3: 39, 4: 3}
type at index 3       : {'list': 42}       non-empty STR at [3]: 0
type at index 5       : {'NoneType': 42}   non-empty STR at [5]: 0
```

The sweep happened to catch **3 genuinely FAILED artifacts** (status `4`) — exactly the case the deleted read claimed to serve. On those rows index 3 holds `[[['61257d3c-…']], [['e96c30a6-…']]]`: nested source-id lists, i.e. `Artifact.sources`, precisely as you documented. Never a string, in any row.

Then `poll_status()` against one of those real failed artifacts, run under both `main` and this branch:

```
status: 'failed'   error: None   error_code: None   is_rate_limited: False
```

Byte-identical. Re-running the same analysis over the cassette corpus (301 rows) agrees. So the `Changed`-not-`Fixed` framing in the CHANGELOG is right in principle, and in practice the observable behaviour is unchanged.

Two things I want to call out as good practice, because they're the reason this reviewed quickly:

- **Refusing the tautological test.** Asserting `error is None` would have passed against the pre-removal code too. Recording `read_indices` and pinning that 3 and 5 are absent is a test that actually fails on the old implementation — I confirmed it reports `[0, 4, 2, 3, 5]` against `e55de0af`. That distinction is easy to skip and you didn't.
- **Deleting the tests instead of adjusting them.** They encoded a row shape the backend has never emitted; keeping them green would have meant preserving the bug.

Verified separately: no leftover references to the removed symbols, no orphaned imports, and `is_rate_limited`'s error-substring fallback was already dead for wire-sourced errors, so nothing regresses there.

No changes requested. Ready to merge.

teng-lin:
@claude review

Issue #2127, opened 2026-08-07
ArtifactStatus codes 1 and 2 are swapped (is_pending/is_processing inverted), and codes 0/5/6 are missing — incl. an unknown ARTIFACT_PENDING_REVIEW state


teng-lin:
## Summary

Our `ArtifactStatus` enum has codes 1 and 2 inverted relative to the backend, and omits three values entirely (0, 5, 6) — including `ARTIFACT_PENDING_REVIEW = 6`, a state that was not previously known to exist. The two terminal states (3 READY, 4 FAILED) are correct, which is why the inversion went unnoticed: it only affects the transitional window.

## Evidence (live-verified, not inferred)

Backend enum (merged from the Android app's `pp.txt` + `objs.txt` symbol dumps) versus ours (`src/notebooklm/rpc/types.py:204-213`):

| code | backend | ours |
|---|---|---|
| 0 | `ARTIFACT_STATUS_UNKNOWN` | **absent** → `"unknown"` |
| 1 | `ARTIFACT_STATUS_INITIALIZED` | `PROCESSING` → `"in_progress"` ❌ |
| 2 | `ARTIFACT_STATUS_PROCESSING` | `PENDING` → `"pending"` ❌ |
| 3 | `ARTIFACT_STATUS_READY` | `COMPLETED` ✓ |
| 4 | `ARTIFACT_STATUS_FAILED` | `FAILED` ✓ |
| 5 | `ARTIFACT_STATUS_SUGGESTED` | **absent** → `"unknown"` |
| 6 | `ARTIFACT_PENDING_REVIEW` | **absent** → `"unknown"` |

**The two transitional states are inverted.** A live trace of a fresh artifact showed **t+4s → code 2**, **t+20s → code 3** — so code 2 is what an artifact reports *while generating*. We report `"in_progress"` for a merely-initialized artifact and `"pending"` for one that is actively generating; `Artifact.is_processing` and `is_pending` are backwards with them. Code 2 = actively generating is CONFIRMED-LIVE twice, independently (two separate traces, both `2 → 3`).

**The mislabel is visible at our own public API.** A live generation returned `GenerationStatus(status='pending')` while the backend artifact was at `PROCESSING` 180 ms later. A caller polling "queued vs running" gets the wrong answer from the first response onward.

[...]

Issue #2117, opened 2026-08-07
QuizQuantity.MORE is an alias for STANDARD, but the backend accepts cardQuantity=3 — '--quantity more' silently no-ops


teng-lin:
## Summary

`QuizQuantity.MORE` is defined as an alias for `STANDARD = 2`, with a docstring asserting that Google's API cannot distinguish them. Live testing refutes that claim: the backend accepts and persists `cardQuantity = 3`. Because our enum can never emit 3, `--quantity more` silently produces a standard-sized set.

## Evidence (live-verified, not inferred)

`src/notebooklm/rpc/types.py:285-290`:

```python
class QuizQuantity(int, Enum):
    """Quiz/Flashcards quantity options.

    Note: Google's API only distinguishes between FEWER (1) and STANDARD (2).
    MORE is an alias for STANDARD - the API treats them identically.
    This matches the observed behavior from NotebookLM's web interface.
    """

    FEWER = 1
    STANDARD = 2
    MORE = 2  # Alias for STANDARD - API limitation
```

The backend accepted and persisted `cardQuantity = 3` — observed while probing the flashcard quantity/difficulty transposition (#2116), where a difficulty code of `3` landed in the quantity slot and the server echoed the stored artifact back as `cardQuantity=MORE(3)` rather than rejecting or clamping the value.

The documented "API limitation" is not a limitation of the API; it is a limitation of this enum.

## Why this matters

- `MORE` is a public, documented option across CLI, MCP and REST. Selecting it produces exactly the same request bytes as `STANDARD`, so the user gets a standard-sized quiz or flashcard set with no error and no warning that their choice was dropped.
- The docstring actively discourages anyone from fixing it, since it states as fact that the backend cannot express the distinction.
- Affects both quiz and flashcards, since they share `QuizQuantity`.

## Suggested fix

Set `MORE = 3` and drop the "API limitation" note from the docstring. Note that `MORE` and `STANDARD` currently share a value, so anything doing `QuizQuantity(2)` round-trips or member comparisons may be relying on the collapse — worth a grep before changing.

## Caveat

CONFIRMED-LIVE (backend accepted and persisted `cardQuantity=3`) plus code-verified.

---

Source: `docs/notes/web-rpc-vs-mobile-grpc-audit-2026-08-07.md`, §1.3

PR #2208, opened 2026-08-12
fix(ci): unbreak main — invalid allowlist JSON + ruff import order

teng-lin:
## main is currently red

`scripts/api-compat-allowlist.json` is not parseable. Every PR fails:

```
FAILED tests/_guardrails/test_public_surface_manifest.py::test_ungated_public_surface_covers_exactly_the_unpinned_modules
FAILED tests/_guardrails/test_public_surface_manifest.py::test_baseline_matches_committed_file[ungated_surface]
RuntimeError: invalid JSON in scripts/api-compat-allowlist.json:
  Expecting ',' delimiter: line 157 column 7 (char 17820)
```

Confirmed on `origin/main` itself (`bde3d46e`), not just on a branch.

## Cause — a concurrent-merge hazard, not a bad PR

#2194 (`QuizQuantity.MORE`, #2117) and #2200 (`ArtifactStatus`, #2127) each appended a `changed-enum-value` entry to the same array. **Each was valid alone and each merged green against the main it was tested against.** Merging the second textually joined two adjacent objects:

```json
      "object": "notebooklm.types.ArtifactStatus.PROCESSING",
      "reason": "Value changed 1 -> 2. ..."      <-- no closing }
      "object": "notebooklm.QuizQuantity.MORE",  <-- no opening { or "code"
```

`ArtifactStatus.PROCESSING` lost its closing `}`; `QuizQuantity.MORE` lost its opening `{` and `"code"` key. The result is one malformed object with a duplicated `"object"` key.

Git produced **no conflict marker**, because both sides were syntactically plausible line additions into the same region. Neither PR author did anything wrong, and neither PR's CI could have caught it — the breakage only exists in the merged result.

## The fix

Restores the object boundary. Nothing else changes:

- entry **content** untouched
- **33 entries**, none added or removed
- **no duplicate `object` values**
- both `QuizQuantity.MORE` import paths preserved (`notebooklm.` and `notebooklm.types.`) — these are two deliberate entries, not an accident of the merge

## Verification

| gate | result |
|---|---|
| `pytest tests/_guardrails/test_public_surface_manifest.py` | **286 passed** (was 2 failed) |
| `scripts/audit_public_api_compat.py --check-stale` | **exit 0** |
| `pytest tests/_guardrails` | **1918 passed**, 1 xfailed |
| `json.load()` | parses; 33 entries, no dupes |

## Please merge promptly

Every open PR is red until this lands, and the failure is unrelated to their contents.

## Follow-up worth considering

This file is an append-only JSON array edited by most public-surface PRs, so it will collide again. A `jq . scripts/api-compat-allowlist.json` (or `json.tool`) check in the pre-commit hook would turn a silent textual merge into an immediate local failure. Happy to open that separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the wire value for the `QuizQuantity.MORE` quiz option.
  * Added compatibility coverage to distinguish it from the `STANDARD` option, helping ensure quiz quantity selections are interpreted correctly across supported interfaces.

### DIFF
--- a/scripts/api-compat-allowlist.json
+++ b/scripts/api-compat-allowlist.json
@@ -154,6 +154,9 @@
       "code": "changed-enum-value",
       "object": "notebooklm.types.ArtifactStatus.PROCESSING",
       "reason": "Value changed 1 -> 2. #2127: this client had ArtifactStatus wire codes 1 and 2 transposed relative to Google's backend enum. Code 1 is ARTIFACT_STATUS_INITIALIZED (queued; the worker has not started) and code 2 is ARTIFACT_STATUS_PROCESSING (actively generating) \u2014 confirmed by the recovered enum dump in docs/mobile/enums.txt, by two independent live traces showing 2 -> 3 on a generating artifact, and by a live probe on the fixing branch. The old integers were wrong about the wire rather than a contract worth preserving, so this is a decode correction, not an API redesign. Member NAMES and the status-string vocabulary are unchanged: PENDING has always meant 'queued' and PROCESSING 'actively generating', so callers using .is_pending / .is_processing / .status_str / GenerationState get the correct answer with no change. Only callers comparing the raw integers must flip them. Documented in CHANGELOG.md (with the timeout-exception consequence), docs/stability.md and docs/python-api.md; pinned by tests/_guardrails/_wire_contract.py ENUM_BINDINGS against the backend dump."
+    },
+    {
+      "code": "changed-enum-value",
       "object": "notebooklm.QuizQuantity.MORE",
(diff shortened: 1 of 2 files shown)


teng-lin:
## Scope grew: main is red in **two** independent ways, both from the same merge

While verifying the JSON fix I found `Code Quality` also fails on every PR, for an unrelated reason. Both are now in this PR.

**2. ruff import order** — `tests/_guardrails/test_wire_contract.py` on `main`:

```python
from notebooklm.rpc.types import ARTIFACT_STATUS_SUGGESTED_WIRE_NAME, ArtifactStatus
import notebooklm.rpc.types as rpc_types
```

ruff's isort rule sorts plain `import x` before `from x import y`, so `pre-commit run --all-files` reports `1 error (1 fixed)` and exits 1. Because the Code Quality job runs pre-commit over **all** files, this fails every open PR regardless of its contents — exactly like the JSON break.

Swapping the two lines is the entire change.

**How I found it:** I had verified the JSON fix with the guardrail tests and the compat audit, but had not run `pre-commit` in this worktree — the same gap that CI then caught. Fixed and re-verified: `pre-commit` exit 0, `tests/_guardrails` 1918 passed.

Both defects entered with the #2200 merge and neither could have been caught by #2194's or #2200's own CI, since each only exists in the merged result.

Issue #2220, opened 2026-08-13
Idempotency probes swallow non-transport list failures and retry, which can duplicate the resource


teng-lin:
Raised by CodeRabbit on [PR #2214](https://github.com/teng-lin/notebooklm-py/pull/2214) (the #2204 `add_url` fix). Filed rather than actioned there, because it is a design question that spans all three probe paths and changing one of them in isolation would split a pattern that is currently uniform.

## The shape

All three `idempotent_create` probes treat a **non-transport** failure of their `list()` the same way — swallow it, log, and return `None`:

- `add_url._probe` (`src/notebooklm/_source/add.py`)
- `add_drive._probe` (same file, #2113 / PR #2199)
- `register_file_source._probe` (`src/notebooklm/_source/upload.py`)

`None` means "no match", so `idempotent_create` re-issues the create. All three run with `disable_internal_retries=True`, so the probe is the only thing standing between a commit-lost-response and a duplicate.

The concern: a drifted `GET_NOTEBOOK` makes the strict decoder raise `RPCError`, which is **indistinguishable from "the create did not land"** — so the wrapper retries, and if the create *had* landed, that retry duplicates the source. The probe failing is precisely when its protection is most needed.

## Why it is not obviously a bug

The alternative — propagate, don't retry — is not free either. It converts every decode blip into a hard failure of an add that, in the common case, never landed. The caller then retries by hand, which can create the duplicate anyway, only now without the warning. So:

|  | swallow + retry (today) | propagate |
|---|---|---|
| create landed, probe broken | duplicate source, logged | no duplicate, caller told |
| create did not land, probe broken | recovers correctly | hard failure, manual retry |

Neither column is strictly better; it is a question of which failure the callers of a bulk-import path would rather absorb.

## Current mitigation

#2204 raised `add_url`'s swallow from `DEBUG` to `WARNING` (matching what #2199 did for `add_drive`), because the `notebooklm` logger defaults to `WARNING` and a `DEBUG` record was being dropped before any handler saw it. So the path is now at least loud. Also worth noting the same PR found a second, quieter instance of this class: the *baseline* capture in `add_url`/`add_drive` swallowed at `DEBUG` too, and five of the repo's own VCR tests were silently exercising the disabled path. Both are now `WARNING`.

`register_file_source`'s probe swallow is still at `DEBUG` and should move to `WARNING` regardless of how the larger question is decided — that one is a straightforward parity fix.

## What to decide

[...]

Issue #2204, opened 2026-08-12
add_url idempotency probe can adopt a pre-existing same-URL source and report a phantom success


teng-lin:
Spun out of #2113 / PR #2199, which fixed the equivalent defect for `add_drive`. **Not fixed there** — filing separately so the Drive fix stays scoped.

## The shape

`add_url`'s idempotency probe (`src/notebooklm/_source/add.py`, `add_url._probe`) matches purely on equality:

```python
for source in sources:
    if source.url == url:
        return source
```

There is no baseline filter, so the probe cannot tell *"the create I just issued landed"* from *"a source with this URL was already in the notebook."*

## Why that is a bug

A notebook can legitimately hold two sources with the same URL — the backend permits it, and `SourceLister.list` dedupes by **source id**, not by URL (see the #1919 comment noting research imports re-emit a URL). So on the commit-lost-response path:

1. caller adds URL `X`, which is already present in the notebook;
2. the create fails with a retryable transport error *before committing*;
3. `idempotent_create` runs the probe, which matches the **pre-existing** source;
4. `add_url` returns that source and reports success — for a create that never happened.

The caller gets a `Source` whose `id` belongs to someone else's (or an earlier) add. With `wait=True` this is worse: the pre-existing source is already `READY`, so the wait returns immediately and the failure is fully masked.

This is exactly the failure mode #2113 hit for Drive, where a `documentId` is likewise non-unique per notebook.

## Suggested fix

Mirror what PR #2199 did for `add_drive`, which in turn mirrors `register_file_source` (`_source/upload.py`):

- capture `{source.id for source in await list_sources(notebook_id)}` before the first create attempt;
- filter probe matches to ids **not** in that baseline;
- when the baseline is unavailable (`None` sentinel) and a match exists, raise `SourceAddError` rather than guessing;
- raise on an ambiguous multi-match.

## Known cost, and why it needs a decision rather than a copy-paste

The baseline is an extra `GET_NOTEBOOK`, and the backend **writes** `lastViewedTime` when answering one (#2126) — so it bumps the notebook to the top of the user's Recent list. For `add_drive` that trade was clearly worth it. `add_url` is the highest-traffic add path and is commonly driven in bulk, so the recency churn is more noticeable and the call is worth making deliberately rather than by symmetry.

Alternatives worth weighing before implementing:
- accept the bump for correctness parity with `add_drive` / `add_file`;
[...]

Issue #2113, opened 2026-08-07
add_drive idempotency probe can never match (matches on url, which Drive sources never have) — retry after transient failure duplicates the source


teng-lin:
## Summary

`add_drive`'s idempotency probe (`_source/add.py:347-368`) matches a candidate existing source by checking `source.url` against `/d/<file_id>` — but Drive-backed sources never carry a `url` at all (confirmed live: `url=None` on all 4 real Drive sources sampled). Since `add_drive` also sets `disable_internal_retries=True`, the probe is the *only* retry-safety net — and it can never match. A 5xx or network failure after the server has already committed the create causes the probe to return `None`, and `idempotent_create` re-issues the add, producing two copies of the same Drive file.

## Evidence (live-verified)

```
Source(id='ef72c03c-…', title='Rubisco Research: Status and Future', url=None, …)
SourceRow: shape=DEEPLY_NESTED  mime=None  url=None  type_code=1
```

This client's own code already documents the root cause — `_types/sources.py:83-86` states plainly that Drive sources carry no URL, directly contradicting the probe's own docstring at `add.py:268-271`, which assumes one exists to match against.

## Suggested fix

The correct dedup key is already on the wire and unread: `metadata[0][0]` (`GoogleDocsSourceMetadata.documentId`) / `metadata[9][0]` (`GoogleDriveSourceMetadata.documentId`) — confirmed live to equal exactly the requested `file_id` (e.g. `"1FDW_u2QKNxpYBBvs-k03F4n039Ufuiv2DgsOP7TR6As"`). Expose `SourceRow.drive_document_id` and have the `add_drive` idempotency probe match on that instead of `url`.

PR #2235, opened 2026-08-13
fix(notebooks): treat a failed create baseline as unavailable, not empty (#2232)

### LINKED ISSUE #2232: notebooks.create degrades a failed baseline to an empty set(), so a pre-existing same-titled notebook can be returned as freshly created
Found while doing #2220 (PR #2225), and deliberately left out of that PR: it is the #2204-class defect for notebooks, not the probe-failure question #2220 was filed to decide.

## The shape

All four `PROBE_THEN_CREATE` paths capture a pre-create baseline so their probe can tell *"the resource I just created"* from *"one that was already there"*. Three of them treat a **failed** baseline as a sentinel:

```python
# _source/add.py (add_url, add_drive), _source/upload.py (register_file_source)
baseline_ids: set[str] | None
try:
    baseline_ids = {source.id for source in await list_sources(notebook_id)}
except Exception as exc:
    logger.warning(...)
    baseline_ids = None          # "unavailable" — the probe then refuses to guess
```

and their probes raise `SourceAddError` on **any** match while the baseline is unavailable, rather than adopting one.

`notebooks.create` degrades to an empty set instead (`src/notebooklm/_notebooks.py`):

```python
except Exception as exc:
    logger.warning("create: baseline list() failed (%s); falling back to an empty baseline…")
    baseline_ids = set()
```

The probe filter is then `nb.id not in set()` — i.e. **no filter at all**:

| pre-existing notebooks with this title | result |
|---|---|
| exactly one | returned as `PROBED`, so `create()` hands the caller **someone else's notebook** as if it had just created it — silent, wrong resource, no signal |
| two or more | hard `Cannot disambiguate` failure on a create that had nothing to do with them |

The first row is the #2204 failure mode verbatim: *"a caller handed a silent duplicate — or the wrong id — has nothing to act on and no reason to look."*

## Why it is narrow but real

It needs three things at once: a transient failure on the baseline `list()`, a commit-lost failure on the create, and a successful probe `list()`. Rare — but `notebooks.create` is retried by exactly the callers most likely to hit a flaky window, and the wrong-notebook outcome is unbounded: every subsequent `sources.add_*` / `chat.ask` in that session targets the wrong notebook.

The limitation is currently documented in-place as an *"accepted known limitation"*, predating the sentinel pattern the three source paths later adopted (#2113, #2204). It is now the only one of the four that guesses.

## Suggested fix

Adopt the same `None` sentinel: on a failed baseline, let the probe raise rather than return a match it cannot attribute. The existing ambiguity raise (`RPCError`, `Cannot disambiguate notebook with title …`) is the right type and message shape; it just needs to fire on `len(matches) >= 1` when the baseline is unavailable, not only on `> 1`.

Note the tests that pin today's behaviour will need inverting, the same way #2220 inverted the probe-swallow tests.

## Related

- #2220 / #2225 — made all four probes propagate a probe *failure*; ADR-0005 now states the rule and **explicitly names this gap as not closed**, so the ADR and the code agree today.
- #2204 (`add_url`), #2113 (`add_drive`) — the same fix already applied to the source paths.

Also noted while reviewing, lower priority and not part of this issue: three methods classified `PROBE_THEN_CREATE` in `_idempotency_policy.py` have **no probe implementation at all** (`CREATE_ARTIFACT`, `GENERATE_MIND_MAP`, `SHARE_NOTEBOOK`). The classification alone force-disables transport retries, so there is no blind-retry duplicate — but there is no recovery either. And `idempotent_create` logs its retry *trigger* at WARNING while logging both *outcomes* ("probe found existing resource" at INFO, "probe returned no match; will retry" at DEBUG) below the `notebooklm` logger's WARNING default, so in a default configuration an operator sees the probe start and never learns whether the create was recovered or blindly re-issued.

teng-lin:
Closes #2232

## The defect

All four `PROBE_THEN_CREATE` paths snapshot the resource list *before* the create so their probe can tell "the row I just made" from "one that was already there". Three of them degrade a **failed** snapshot to `None` — the "baseline unavailable" sentinel — and their probes then refuse to attribute any match.

`notebooks.create` degraded to an empty `set()` instead. An empty set reads as *"nothing existed"*, so the filter `nb.id not in baseline_ids` matched everything:

| pre-existing notebooks with this title | pre-fix result |
|---|---|
| exactly one | returned as `PROBED` — the caller gets **someone else's notebook** as if `create` had just made it. No signal, nothing to act on. |
| two or more | hard `Cannot disambiguate` failure on a create that had nothing to do with them |

The first row is the #2204 failure mode verbatim, and it is unbounded: every subsequent `sources.add_*` / `chat.ask` in that session targets the wrong notebook. It was documented in place as an *accepted known limitation*, predating the sentinel the source paths adopted in #2113 / #2204 — leaving `notebooks.create` the only one of the four that guessed.

## The change

Adopt the sentinel, mirroring `add_url`'s shape exactly (`_source/add.py:299-419`):

- `baseline_ids: set[str] | None`, `None` on failure; the WARNING now says the probe is disarmed rather than "falling back to an empty baseline".
- `baseline_error` is retained so the raise can name what broke the snapshot — the caller sees "baseline unavailable" long after that line ran.
- `_probe()` filters by title first, subtracts the baseline **only when it has one**, and otherwise raises a marked-`unconfirmed` `RPCError` naming the ambiguous rows (`_describe_notebooks`, the twin of `_describe_sources`).

The guard is on *matches*, not on the sentinel itself: a probe that lists cleanly and finds no notebook with this title has answered the question the baseline was only needed to disambiguate, so the retry still proceeds. An unavailable baseline never fails a create that would otherwise succeed.

## The tradeoff — please engage with this, not just the diff

**Silent wrong identity vs. a louder duplicate create.**

Today's failure is that a caller *silently receives a pre-existing notebook believing it is new*. That is the harder bug to ever notice: nothing is raised, nothing is logged above DEBUG, and the damage shows up later as sources and chats attached to a notebook nobody meant to touch.

Refusing to guess trades that for a different risk: when the probe cannot attribute a match, we raise instead of adopting, and the caller may end up issuing a duplicate create. A duplicate is loud — it is visible in the notebook list, it is diagnosable, and the error tells the caller exactly which rows to look at and why. The maintainer's call is that **duplicate-loudly beats silent wrong identity**, which is the same call #2113 and #2204 already made for the three source paths.

It takes three things at once to reach either outcome (a failed baseline `list()`, a commit-lost create, and a successful probe `list()`), so this is narrow — but `notebooks.create` is retried by exactly the callers most likely to hit a flaky window.

## Tests

- `test_create_baseline_failure_makes_a_match_ambiguous` (unit, parametrised over a decode `RPCError` and a transport `ServerError` baseline failure) — asserts the pre-existing id is named, `__cause__` is the baseline failure, `__context__` is the create's transport error, `unconfirmed` is set, and exactly one create attempt fired.
- `test_notebooks_create_baseline_failure_refuses_to_adopt_a_pre_existing_notebook` (wire-level, `MockTransport`) — the #2232 scenario end to end. The load-bearing assertion is negative: the call must not *return* the pre-existing notebook.
- `test_notebooks_create_baseline_failure_still_retries_when_probe_finds_nothing` — pins that the sentinel gates on matches, catching a blanket `if baseline_ids is None: raise` regression.

**Mutation-tested.** Reverting the source fix (`baseline_ids = set()` + the old unconditional filter) turns the three new tests red and leaves the other 55 in those files green. Separately, mutating `elif matches:` → `else:` turns only the "still retries" test red.

## Docs

- `CHANGELOG.md`: new `Unreleased / Fixed` entry, and the #2220 paragraph that read *"what changed there is only the log level"* is scoped to its own release — that sentence described exactly what this PR changes.

- ADR-0005 named this as the one gap the #2220 probe-failure rule left open. It now records the closed rule, and states the trade explicitly.
- `docs/python-api.md` idempotency table: the `notebooks.create` row now carries the non-unique-title caveat, matching the `add_url` row.

## Verification

```
uv run pytest -n auto --dist loadgroup --cov=src/notebooklm --cov-fail-under=90
  -> 16511 passed, 64 skipped, 1 xfailed; coverage 96.80%
uv run mypy src/notebooklm --ignore-missing-imports  -> Success, 353 files
uv run pre-commit run --all-files                    -> exit 0
```

Full 6-extra install (`browser dev markdown mcp server impersonate`).

## Live-API probe: run, with a pre-fix/post-fix contrast

This changes real idempotency behaviour on the wire, so it was probed against the real backend. Same script, same account, same injected conditions (unavailable baseline + lost create response), run on `origin/main` and on this branch:

- **pre-fix** — `create()` returned `0acef09e-…`, the *pre-existing* row, with no error raised
- **post-fix** — `RPCError: Cannot disambiguate notebook with title …` naming that row, `unconfirmed=True`, and exactly one row carrying the title afterwards (the ambiguity aborted the loop rather than firing a second create)

Also confirmed live: two `create()` calls with an identical title yield two distinct notebooks, so a title genuinely is not an identity on this backend. Full output, the script, and what the probe cannot show are in the [probe comment](https://github.com/teng-lin/notebooklm-py/pull/2235#issuecomment-5286293681).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `notebooks.create()` reliability when pre-create checks fail.
  * Prevented returning an existing same-titled notebook after an ambiguous failed creation.
  * Ambiguous matches now raise an unconfirmed error with candidate details.
  * Baseline check failures are logged as warnings and handled consistently.

* **Documentation**
  * Updated idempotency guidance and changelog entries to describe the revised behavior.


teng-lin:
@codex review. @claude review

teng-lin:
## Live-API probe — run, with a pre-fix/post-fix contrast

Auth turned out to be valid, so this was run against the real backend rather than deferred. Every `list` and `create` below is a real RPC; only the two conditions the defect *needs* are injected — an unavailable pre-create baseline, and a create whose response is lost. All notebooks are deleted afterwards and the tail re-lists to prove nothing leaked.

### Premise: notebook titles really are non-unique on the wire

```
A. two creates, same title -> '55338937-84ad-4bc2-a759-9d88aacc3ccc', '0ec06215-f1dc-4156-8250-3ee9267f9387'
A. distinct ids (premise of the fix): True
A. rows carrying that title now: 2
```

Two `create()` calls with an identical title yield two distinct notebooks that both stay in `LIST_NOTEBOOKS`. This is the `add_url`/#2204 finding repeated for notebooks: a title is not an identity, so a post-hoc title match alone can never attribute a row.

Also confirmed on the same run: with the baseline read failing but the create **succeeding**, `create()` still returns normally (`C. plain create -> '1203bb8d-…'`). An unavailable baseline disarms the probe; it never fails a create that works.

### The defect window: baseline unavailable + create lost + one pre-existing same-titled row

Same script, same account, same injected conditions, run twice.

**On `origin/main` (64ed527f), pre-fix:**

```
setup: pre-existing row -> '0acef09e-0b1a-42c2-80f9-58e0aa977c7d'
B. RETURNED '0acef09e-0b1a-42c2-80f9-58e0aa977c7d' title='probe-2232b-a265556f'
B. is the PRE-EXISTING row: True
B. rows carrying the title afterwards: 1
```

`create()` returned, with no error and nothing above DEBUG in the logs, a notebook it did not create. The caller now holds an id it believes is a fresh empty notebook; it is someone's existing one, and it is the only row in the account with that title, so a "did I get a duplicate?" check afterwards also comes up clean.

**On this branch, post-fix:**

```
setup: pre-existing row -> '1cabc4fe-07f3-4083-8672-71a5c9b7cf03'
B. raised RPCError: Cannot disambiguate notebook with title 'probe-2232b-354eb897':
   the pre-create baseline snapshot failed (RPCError), so
   1cabc4fe-07f3-4083-8672-71a5c9b7cf03 ('probe-2232b-354eb897') may either predate
   this create or be the notebook it just created. Check your notebook list before retrying.
B. unconfirmed marker: True
B. __cause__: RPCError: injected baseline failure (#2232 probe)
B. __context__: ServerError
B. rows carrying the title afterwards: 1
```

Observed and checked:

- the ambiguous row is **named** (`1cabc4fe-…`), so "check your notebook list" is actionable
- `unconfirmed=True`, so the MCP/REST/CLI surfaces keep it out of the retriable band
- `__cause__` is the baseline failure, `__context__` is the create's `ServerError` — the two-part chain documented in `docs/python-api.md`
- **one** row carrying the title afterwards: the ambiguity aborted the loop, it did not fire a second create

### Reproduce

```bash
git worktree add .claude/worktrees/fix-2232 fix/2232-notebook-create-baseline
cd .claude/worktrees/fix-2232
uv sync --frozen --extra dev --extra impersonate
uv run python f2232-live-probe2.py      # script inlined below
```

<details><summary><code>f2232-live-probe2.py</code></summary>

```python
import asyncio
import uuid

from notebooklm import NotebookLMClient, RPCError, ServerError
from notebooklm._notebooks import NotebooksAPI
from notebooklm.rpc import RPCMethod

TITLE = f"probe-2232b-{uuid.uuid4().hex[:8]}"


async def main() -> None:
    made: list[str] = []
    async with NotebookLMClient.from_storage() as client:
        try:
            # One pre-existing notebook with the title. This is the row the
            # pre-fix code adopted and handed back as "freshly created".
            pre = await client.notebooks.create(TITLE)
            made.append(pre.id)
            print(f"setup: pre-existing row -> {pre.id!r}")

            executor = client.notebooks._rpc
            real_rpc = executor.rpc_call
            real_list = NotebooksAPI.list
            list_calls = {"n": 0}

            async def flaky_list(self):
                list_calls["n"] += 1
                if list_calls["n"] == 1:
                    raise RPCError("injected baseline failure (#2232 probe)")
                return await real_list(self)

            async def lossy_create(method, *args, **kwargs):
                if method is RPCMethod.CREATE_NOTEBOOK:
                    # The create never reaches the server: this models the
                    # commit-lost 5xx where nothing was written. Whatever the
                    # probe finds afterwards therefore is NOT ours.
                    raise ServerError("injected 502 (#2232 probe)")
                return await real_rpc(method, *args, **kwargs)

            NotebooksAPI.list = flaky_list
            executor.rpc_call = lossy_create
            try:
                try:
                    got = await client.notebooks.create(TITLE)
                except Exception as exc:
                    print(f"B. raised {type(exc).__name__}: {exc}")
                    print(f"B. unconfirmed marker: {getattr(exc, 'unconfirmed', False)}")
                    print(f"B. __cause__: {type(exc.__cause__).__name__}: {exc.__cause__}")
                    print(f"B. __context__: {type(exc.__context__).__name__}")
                else:
                    print(f"B. RETURNED {got.id!r} title={got.title!r}")
                    print(f"B. is the PRE-EXISTING row: {got.id == pre.id}")
            finally:
                NotebooksAPI.list = real_list
                executor.rpc_call = real_rpc

            rows = [nb.id for nb in await client.notebooks.list() if nb.title == TITLE]
            print(f"B. rows carrying the title afterwards: {len(rows)} {rows}")
            made.extend(r for r in rows if r not in made)
        finally:
            for nb_id in made:
                try:
                    await client.notebooks.delete(nb_id)
                except Exception as exc:
                    print(f"cleanup: FAILED to delete {nb_id}: {exc}")
            leftover = [nb.id for nb in await client.notebooks.list() if nb.title.startswith(TITLE)]
            print(f"cleanup: leftover rows (must be []): {leftover}")


asyncio.run(main())
```

</details>

### What the probe cannot show

The one thing no probe can force is the *other* side of the tradeoff: a create that genuinely commits and then loses its response, where the new behaviour raises and the caller's retry produces a second notebook. Injecting a failure after a real call-through would model it, but the pre-fix code raises there too (two same-titled rows > 1 match), so it does not discriminate. The duplicate risk is argued, not measured — and it is the visible, diagnosable half of the trade.

`cleanup: leftover rows (must be []): []` on every run; nothing was left in the account.

teng-lin:
Addressed in 4497e848 — four follow-ups from a self-review pass, each checked against the code before acting.

**1. Dropped `raise ... from baseline_error`.** The original raise chained the baseline failure onto `__cause__`. `idempotent_create` documents the opposite guarantee for probe raises (`_idempotency.py`):

> The probe is awaited inside the handler for the transport failure, so that failure is always reachable through the `__context__` chain **and the traceback shows both halves**

Setting `__cause__` breaks the display half. CPython's formatter picks cause *xor* context (`elif exc.__context__ is not None and not exc.__suppress_context__`), so the create's transport failure stops printing. Verified rather than assumed:

```
--- raise ... from baseline_error ---   shows TRANSPORT: False | shows BASELINE: True
--- plain raise (sibling shape) ---     shows TRANSPORT: True  | shows BASELINE: False
--- manual __cause__ assignment ---     shows TRANSPORT: False | shows BASELINE: True
```

So it is not fixable by assigning `__cause__` by hand — the choice is which half prints. The transport failure is the more important one (it is *"the create that may have committed"*), and the baseline failure's type is already in the message. That is also exactly what `add_drive` and `register_file_source` surface; only `add_url` retains the object, through a `cause=` field `RPCError` does not have. This was the one real divergence from the siblings in the first commit.

**2. Front-loaded the action in the ambiguity message.** `server/_errors.py:61` caps messages at `_MAX_MESSAGE = 300`. Measured with a realistic title:

```
title = 'Q3 Planning — Marketing and Sales Alignment'
len(message) = 338
message[:300] ends: '... or be the notebook it just created. Che'
```

`Check your notebook list before retrying` was being cut off — the one actionable sentence. Reworded to lead with it, matching the probe-failure raise a few lines above, which already front-loads for this exact reason (`_notebooks.py`, *"Action first — the MCP/REST surfaces truncate messages at 300 characters"*). New unit assertion pins that the action survives `[:300]`.

**3. CHANGELOG.** No entry existed, and the #2220 paragraph in `Unreleased` had gone stale — it read:

> The *pre-create baseline* read deliberately still degrades rather than failing … **What changed there is only the log level**

which is precisely what this PR changes. Added the #2232 entry and scoped that sentence to its own release.

**4. Two test assertions strengthened.** The log check in `..._warns_but_proceeds` was changed to `"baseline list() failed"` — but the *pre*-#2232 message contained that substring too (`create: baseline list() failed (%s); falling back to an empty baseline…`), so it no longer discriminated the wording change at all. It now also asserts the distinctive clause `"surface as an ambiguity error"`.

**Re-verified after the change**

- `pytest -n auto --dist loadgroup --cov` → 16511 passed, 96.81%; mypy clean; `pre-commit run --all-files` exit 0
- mutation re-run: reverting the fix still reddens exactly the three new tests, 55 others green
- live probe re-run against the real backend with the new wording:

```
B. raised RPCError: Cannot disambiguate notebook with title 'probe-2232b-e2bdaf10' — check your
   notebook list before retrying: the pre-create baseline snapshot failed (RPCError), so
   9eb19e85-2a36-4215-baa4-204a710ffee9 ('probe-2232b-e2bdaf10') may either predate this create
   or be the notebook it just created.
B. unconfirmed marker: True
B. __cause__: NoneType: None      <- deliberate, see (1)
B. __context__: ServerError       <- the create's transport failure, still printed
B. rows carrying the title afterwards: 1
cleanup: leftover rows (must be []): []
```

**One finding I did not act on:** `..._still_retries_when_probe_finds_nothing` would still pass if the fix were reverted. That is by design — it guards the over-fix direction (a blanket `if baseline_ids is None: raise`), which a revert does not exercise. Its own docstring says so, and mutating `elif matches:` → `else:` does redden it.

teng-lin:
@codex review. @claude review

teng-lin:
Thanks — the sibling-asymmetry observation is correct, and I measured it rather than taking it on trust:

```
add_url baseline msg   len = 333   cut at 300 -> '... be the source it just created. Check the note'
add_drive baseline msg len = 206   (fixed length — names no rows, so it never grows)
```

So it is real for `add_url` specifically, and it gets worse per additional match because `_describe_sources` adds ~40-80 chars a row. `add_drive` and `register_file_source` are short and fixed-shape, so they are not exposed the same way.

Not fixing it in this PR on purpose: `_source/add.py` is out of scope here, and this PR is the #2232 behaviour change plus a live probe — bundling a message rewrite of a path it does not otherwise touch would make the diff harder to reason about, not easier. Filed as **#2238** with the measurements, the precedent (`add_url`'s *own* probe-failure branch already front-loads, with the rationale in a comment right there), and a note that the regression test should assert against the real `[:300]` slice rather than plain substring presence.

Worth saying plainly, since it cuts against the PR's own framing: this PR claims parity with `add_url` and then diverges from it on message ordering. The divergence is deliberate and in the better direction — `add_url` is the one that should move — but until #2238 lands, "mirrors `add_url` exactly" is true of the semantics and not of the wording.

PR #2252, opened 2026-08-17
fix(auth): normalize millisecond cookie expiry to seconds

teng-lin:
## Summary

Addresses cookie expiration values returned in milliseconds by browser extraction tools (including modern Firefox 142+ schema 16 SQLite databases and Chromium extractors via `rookie-cookies`), normalizing them to Unix epoch seconds.

This builds upon and completes the investigation initiated in #2248 by @lesterppo — special thanks to @lesterppo for discovering and raising the millisecond expiry issue!

## Root Cause & Scope

- **Firefox 142+ (Schema ≥ 16)**: `moz_cookies` stores `expiry` in milliseconds rather than seconds. When reading cookies via `rookie_cookies.firefox()` or `rookie_cookies.load()` (the unscoped `--browser-cookies firefox` or `auto` paths), `rookie-cookies` returns the raw database integer with no unit tag.
- **Chromium**: `rookie-cookies` divides raw $\mu s$ timestamps by 1,000 instead of 1,000,000, returning milliseconds.
- **Impact**: Millisecond values (~`1.77e12`) cause:
  1. Corrupted `storage_state.json` (`expires` treated by Playwright as year 57,400+ AD).
  2. `notebooklm auth check` failure (`datetime.fromtimestamp` raises `ValueError: year 57457 is out of range`).
  3. `Cookie.is_expired(now)` bypass in `_auth/psidts_recovery.py` (treating expired tokens as valid for 55,000 years).

## Fix Details

1. **Leaf-Level Normalization in `normalize_cookie_expiry()`**:
   - Rather than normalizing only at the surface adapter (`rookiepy_row_to_storage_row`), normalization is placed inside `normalize_cookie_expiry()` in `src/notebooklm/_auth/cookie_semantics.py`.
   - Uses `_MAX_PLAUSIBLE_EXPIRY_SECONDS = 32_503_680_000` (year 3000-01-01) as a boundary. Any timestamp exceeding this threshold is rescaled via integer division `value //= 1000`.
   - Automatically covers `convert_rookiepy_cookies_to_storage_state`, `CookieJar.from_rookiepy`, `CookieJar.from_storage_state`, `auth_check`, and in-memory PSIDTS recovery.
2. **Firefox `0` Sentinel**:
   - In `_firefox_containers.py`, `moz_cookies.expiry == 0` is mapped to `None` (session cookie sentinel) to avoid treating session cookies as epoch 1970-01-01.
3. **Browser Aliases**:
   - Fixed `ie` -> `internet_explorer` and `octo` -> `octo_browser` mappings in `cookie_jar.py`.
4. **Test Fixtures & Coverage**:
   - Added unit tests for ms-to-s rescaling in `test_auth_issue_2061.py` and `test_auth_cookie_types.py`.
   - Updated far-future test fixtures in `test_auth_psidts_recovery.py` from `99_999_999_999` to `4_102_444_800` (year 2100).

## Commands Run Locally

```bash
uv sync --extra dev --extra browser --extra markdown
uv run ruff format --check src tests
uv run ruff check src tests
uv run mypy src/notebooklm
uv run pytest -m "not e2e"
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Firefox cookie expiry handling for millisecond and microsecond timestamps.
  * Correctly recognizes session cookies and normalizes expiry values across supported formats.
  * Added warnings for invalid cookie expiry values instead of silently ignoring them.
  * Updated browser aliases for Internet Explorer and Octo Browser.

* **Tests**
  * Added regression coverage for expiry normalization, session cookies, and browser aliases.
  * Updated validation baselines and fixtures to reflect the corrected behavior.

### DIFF
--- a/src/notebooklm/_auth/cookie_semantics.py
+++ b/src/notebooklm/_auth/cookie_semantics.py
@@ -51,6 +51,8 @@ class NormalizedExpiry:
 # browser cookie store, so treating the ambiguity as "assume milliseconds"
 # is the safer default here.
 _MAX_PLAUSIBLE_EXPIRY_SECONDS = 32_503_680_000  # 3000-01-01T00:00:00Z
+_MAX_PLAUSIBLE_EXPIRY_MILLISECONDS = 32_503_680_000_000  # 3000-01-01 in ms
+_MAX_PLAUSIBLE_EXPIRY_MICROSECONDS = 32_503_680_000_000_000  # 3000-01-01 in µs
 
 
 def normalize_cookie_expiry(raw: Any) -> NormalizedExpiry:
@@ -60,7 +62,7 @@ def normalize_cookie_expiry(raw: Any) -> NormalizedExpiry:
     session-cookie forms.  Every other accepted numeric representation is a
     dated value, including ``-1.0`` and ``"-1"``.  Dated values use the same
     ``int(float(value))`` conversion as ``http.cookiejar``, then get rescaled
-    from milliseconds to seconds if implausibly large (see
+    from milliseconds or microseconds to seconds if implausibly large (see
     ``_MAX_PLAUSIBLE_EXPIRY_SECONDS``).
     """
     if raw is None or (type(raw) is int and raw == -1):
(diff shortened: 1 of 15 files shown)


teng-lin:
@claude review

teng-lin:
Thanks for the thorough review! Updated `normalize_cookie_expiry()` to loop with `while value > _MAX_PLAUSIBLE_EXPIRY_SECONDS: value //= 1000` and added test cases covering microsecond-scale expiries as well (`234044dc`).

Issue #1518, opened 2026-06-10
UnknownRPCMethodError.data_at_failure spliced into str()/repr() without scrubbing


teng-lin:
**Severity:** Medium (security)

**Location:** `src/notebooklm/exceptions.py:423` (set), `:436` (`__str__`), and `__repr__`; fed unscrubbed from `rpc/_safe_index.py` and `_row_adapters/chat.py`.

**Bug:** The sibling `raw_response` field is scrubbed before preview, but `data_at_failure` is spliced verbatim with `!r` into `__str__`/`__repr__`. Confirmed asymmetry: on one instance `raw_response` redacts `{'SNlM0e': '***'}` while `data_at_failure` emits the token in full. `scrub_secrets` demonstrably catches that exact value, so this is a true gap, not a no-op. The splice bypasses the logging `RedactingFilter` (it's a string-splice, not text reaching a configured handler) — exactly the threat the file already hardened `raw_response` against.

**Triggers when:** schema drift (`UnknownRPCMethodError`) on an RPC whose decoded payload at the failure position embeds a credential-shaped string, **and** the exception is rendered via `str()`/`repr()` (logged, CLI `-vv`, or traceback). `NOTEBOOKLM_DEBUG` is irrelevant — `data_at_failure` is never scrubbed in any mode.

**Fix:** Route `data_at_failure` through `scrub_secrets` at store time (in `__init__` before assignment, or inside `_safe_index._truncate` and the chat-adapter `repr()` sites), mirroring `_truncate_response_preview`. Add a test analogous to `test_unknown_rpc_method_error_scrubs_secrets_in_raw_response`.

**Cross-cutting (Theme A):** same root as the logging-redaction gap above — fix both via one canonical secret registry + parity gate.

---
*Filed from an adversarially-verified multi-agent code audit (each finding was confirmed against the source by an independent verifier). Severity is post-verification.*

Issue #1517, opened 2026-06-10
Logging redaction misses NID / LSOLH / __Host-GAPS session cookies (credential leak to logs)


teng-lin:
**Severity:** Medium (security)

**Location:** `src/notebooklm/_logging.py:169` (cookie-name alternation in the redaction pattern)

**Bug:** The redaction alternation enumerates `__Secure-*`, SAPISID, APISID, SIDCC, HSID, SSID, LSID, SID — but omits **`NID`**, **`LSOLH`**, and **`__Host-GAPS`**, all of which the project's own canonical cassette registry (`tests/cassette_patterns.py` SESSION_COOKIES / HOST_COOKIES) classifies as must-scrub session credentials. A bare `NID=g.a000-…` substring passes through `scrub_secrets` verbatim. There is no `g.a000-`/`sidts-`/`ya29.` token-shape catch-all on the logging side (it exists only in the cassette scrubber), so nothing else nets it.

**Triggers when:** a log record contains a bare (non-`Cookie:`-header) `NID=`/`LSOLH=`/`__Host-GAPS=` token. Real sink: `_auth/refresh.py` logs refresh-command `stdout`/`stderr` at DEBUG through the redacting logger — a refresh script echoing `NID=…` leaks a live credential to logs/stderr.

**Fix:** Derive the alternation from the same canonical registry the cassette scrubber uses (with a parity guardrail asserting the runtime set is a superset), and/or add `NID`/`LSOLH`/`__Host-GAPS`/`__Secure-OSID` explicitly. Port the `g.a000-`/`sidts-`/`ya29.` token-shape catch-all into `_REDACT_PATTERNS` as defense-in-depth so disclosure fails closed regardless of carrier name.

**Note:** `OSID` is *not* affected — it's incidentally caught by the `SID` alternative. The genuine leaks are NID, LSOLH, __Host-GAPS.

**Cross-cutting (Theme A):** shares a root with the `data_at_failure` scrub gap — runtime redaction is enumerated per-site instead of derived from one secret registry. Best fixed together.

---
*Filed from an adversarially-verified multi-agent code audit (each finding was confirmed against the source by an independent verifier). Severity is post-verification.*

PR #2254, opened 2026-08-20
fix(secrets): scrub the current `AQ.` Google API-key shape (#2253)

### LINKED ISSUE #2253: security: credential scrubber misses the current `AQ.`-prefixed Google API-key shape
## Problem

`AUTH_TOKEN_SHAPE_PATTERNS` (`src/notebooklm/_secrets.py`) and the cassette registry's `_AUTH_TOKEN_PATTERNS` + `_GOOGLE_API_KEY_PATTERN` (`tests/cassette_patterns.py`) recognise the Google API-key shape only as `AIza` + 35 key chars.

Google has since migrated AI Studio API keys to an `AQ.`-prefixed format (~53 chars, e.g. `AQ.Ab8…`). That shape matches **none** of the four registered patterns, so a key in the current format is not redacted — it rides through into logs, `data_at_failure` / `payload_preview`, and recorded cassettes. This is the same leak class that #1517 / #1518 closed for the legacy shape, reopened by a format change on Google's side rather than by a code regression.

## Evidence

Against a synthetic key in the current shape:

```
patterns before → matched: False
patterns after  → matched: True   (consumes the whole key, no trailing fragment)
```

Confirmed present on `main` at 3bb0c18 — `grep -c 'AQ\.'` returns 0 in both `_secrets.py` and `tests/cassette_patterns.py`.

## Severity

Not a live incident as far as I can tell, and I want to be precise about that: I have **not** verified whether the key NotebookLM embeds in `WIZ_global_data` (`JrWMbf` / `B8SWKb` / `VqImj`) has itself migrated to the new format. What I did verify is that the redaction registry cannot cover it if and when it does. The fix is two lines and purely additive, so the cost of closing this ahead of the migration is close to zero.

## Proposed fix

Register `r"AQ\.[A-Za-z0-9_\-]{20,}"` in both registries, keeping them regex-string-equal so `test_runtime_credential_shapes_equal_cassette_shapes` stays green. The scrubber and `_DETECT_AUTH_TOKEN` derive from `_AUTH_TOKEN_PATTERNS`, so nothing else needs wiring. `_GOOGLE_API_KEY_PATTERN` is left untouched — the parity guardrail pins the `AIza` string and legacy keys are still in circulation.

Open-ended `{20,}` rather than an exact count, per the no-fragment rationale already documented for the other shapes.

PR follows.

bcoppo:
## Summary

The credential scrubber recognises Google API keys only in the legacy `AIza` + 35-char shape. Google migrated AI Studio keys to an `AQ.`-prefixed ~53-character format, which matches none of the four `AUTH_TOKEN_SHAPE_PATTERNS` — so a key in the current format is not redacted and rides through into logs, `data_at_failure` / `payload_preview`, and recorded cassettes. Same leak class #1517 / #1518 closed for the old shape, reopened by an upstream format change rather than a code regression.

## Related Issue

Closes #2253

## Changes

Two lines, purely additive — the identical regex in both credential registries:

| File | Symbol |
|---|---|
| `src/notebooklm/_secrets.py` | `AUTH_TOKEN_SHAPE_PATTERNS` |
| `tests/cassette_patterns.py` | `_AUTH_TOKEN_PATTERNS` |

```python
r"AQ\.[A-Za-z0-9_\-]{20,}"
```

Keeping the two registries regex-string-equal is what keeps `test_runtime_credential_shapes_equal_cassette_shapes` green. The scrubber (`SCRUBBED` replacement) and the `_DETECT_AUTH_TOKEN` detector both derive from `_AUTH_TOKEN_PATTERNS`, so no further wiring was needed.

`_GOOGLE_API_KEY_PATTERN` is deliberately left untouched: `test_runtime_secret_registry_pins_known_leak_carriers` pins the `AIza` string, and legacy-format keys are still in circulation.

The tail quantifier is open-ended (`{20,}`) rather than an exact count, following the rationale already documented in both files for the other shapes — a fixed count would scrub only a prefix of a longer key and leave a re-matchable fragment with no name-anchored backstop.

Also adds a `[Unreleased] > Fixed` CHANGELOG entry.

## Test Plan

- [x] If this PR changes architectural shape, an ADR has been added or updated. — n/a, no architectural change

Local results on macOS 15 / Python 3.11, env from `uv sync --frozen --extra browser --extra dev --extra markdown --extra mcp --extra server --extra impersonate --extra cookies`:

```
16722 passed, 65 skipped, 1 xfailed
Required test coverage of 90% reached. Total coverage: 96.54%

check_coverage_thresholds.py        OK: 5 per-file floor(s) all met
check_cassettes_clean.py --strict   141 cassettes scanned, 0 leaks
check_cassettes_clean.py --secrets-only  66 fixtures scanned, 0 leaks
pre-commit run --all-files          ruff / ruff-format / fixture-secrets  Passed
mypy src/notebooklm                 Success: no issues found in 355 source files
tests/_guardrails/test_runtime_secret_registry_parity.py   4 passed
```

Behavioural check against a synthetic key in the current shape:

```
patterns before → matched: False
patterns after  → matched: True, and the match consumes the whole key (no trailing fragment)
```

## Notes

**Scope of the claim.** I have not verified whether the key NotebookLM embeds in `WIZ_global_data` (`JrWMbf` / `B8SWKb` / `VqImj`) has itself migrated to the new format — so this is a gap in the redaction registry, not a demonstrated live leak. I am deliberately not calling it critical. Given the fix is two additive lines behind an existing parity guardrail, closing the gap ahead of the migration seemed clearly worth it.

The `AQ.` shape was observed on a real Google AI Studio key issued in August 2026 (53 chars, `AQ.` prefix). The `{20,}` floor is well under that length, and `AQ.` followed by 20+ key chars is not plausible as legitimate fixture content — the same reasoning the existing `sidts-` / `ya29.` floors rest on.

**AI-assisted.** Written with Claude Code, reviewed and locally tested by me before submitting, per CONTRIBUTING.

## Summary by CodeRabbit

* **Security**
  * Credential redaction now recognizes current `AQ.`-prefixed Google authorization keys.
  * Keys containing 20 or more valid characters after the prefix are detected and scrubbed from logs, exception previews, and cassette fixtures.
  * Unknown-field credentials are also identified and removed before diagnostic output or leak scanning, helping prevent accidental exposure.


teng-lin (changes requested):
Thanks for catching this upstream credential-shape change and for keeping the runtime and cassette registries synchronized. I found one runtime blocker before this can merge.

`AUTH_TOKEN_SHAPE_PATTERNS` is compiled into the runtime redaction patterns, but `scrub_secrets()` returns before running that regex sweep unless the input contains an entry from `SECRET_FAST_PATH_TOKENS`. Because that tuple does not include `"aq."`, a standalone `AQ.` key under an unknown carrier—or in the WIZ fields called out here—still passes through runtime logs, `data_at_failure`, and `payload_preview` unchanged.

I reproduced this against commit `8a0d5d5`: the synthetic `AQ.` key remained verbatim after `scrub_secrets()`, while the cassette `scrub_string()` path correctly replaced it with `SCRUBBED` and `is_clean()` flagged the raw value.

Please add `"aq."` to `SECRET_FAST_PATH_TOKENS` and add a runtime regression case that exercises `scrub_secrets()` (for example via `_AUTH_TOKEN_SHAPE_CASES`; covering an exception preview would also pin the reported sink). Thank you!

PR #2262, opened 2026-08-24
feat(mcp): source_delete accepts multiple sources (bulk) (#1995)

### LINKED ISSUE #1995: MCP: source_delete has no bulk mode — accept multiple sources like source_wait
## Gap

The MCP `source_delete` tool (`src/notebooklm/mcp/tools/sources.py:339`) deletes **one** source per call:

```python
async def source_delete(ctx, notebook: str, source: str, confirm: bool = False)
```

Deleting N sources means N round-trips (each with its own confirm preview). No layer supports batch delete — `client.sources.delete()` (`_sources.py:628`) is single-id, and there's no batch source-delete RPC (unlike `DELETE_LABEL` which is batch).

## Proposal

Overload `source_delete` to accept a plural argument instead of adding a new tool (keeps us under the MCP tool-count ceiling). Mirror the shape `source_wait` already uses one function down (`sources.py:365`):

```python
source: str | None = None,
sources: list[str] | str | None = None,   # list OR comma/JSON string
```

- Resolve all ids up front; `confirm=False` returns a `needs_confirmation` preview listing **every** resolved source (title + id).
- `confirm=True` deletes each (loop over the single-id client call — no batch RPC exists) and returns a structured aggregate like `source_wait`'s: `{notebook_id, deleted, not_found, failed}` with per-bucket counts, so partial failures don't discard successes.
- Keep the single-`source` path working unchanged.

### Scope notes
- Client/CLI batch support is a separate, optional follow-up — this issue is MCP-only.
- No batch RPC on the backend, so this is a client-side fan-out (add a delay between calls per the rate-limiting pitfall in CLAUDE.md).
- Touches the ADR-0025 schema-char budget (new param + docstring) — trim docstrings to stay under, don't raise the budget.

AnnasMazhar:
## Summary

Overloads MCP `source_delete` to accept multiple sources (`sources: list[str] | str | None`), mirroring `source_wait`'s bulk interface. This allows agents to delete a batch of sources in a single tool call instead of N sequential round-trips.

## Related Issue

Closes #1995

## Changes

- **Three branches**:
  - `source=` (single ref) — **legacy wire shape preserved** (`{status, source_id, notebook_id}`), existing callers unaffected
  - `sources=` (list, comma-string, or JSON array) — preview + confirm fan-out with per-id result buckets
  - No arg — deletes all sources in the notebook (matches `source_wait`'s "all sources" path)
- **Input guards fire before any I/O**:
  - Mutual exclusion: `source` + `sources` both set → `ValidationError`
  - Empty list: explicit `sources=[]` rejected (must omit arg for "all sources")
  - Over-cap: `len(sources) > MAX_WAIT_SOURCE_IDS` rejected before resolution
- **Partial-failure correctness**: confirm path buckets every resolved id into exactly one of `deleted` / `not_found` / `failed` — a failed delete does not discard successes
- **Rate limiting**: 250 ms inter-call delay during fan-out (per CLAUDE.md rate-limiting pitfall); last id skips the delay

## Test Plan

7 new unit tests:
- `test_source_delete_bulk_preview_lists_every_resolved_source`
- `test_source_delete_bulk_comma_string_parsed_like_source_wait`
- `test_source_delete_bulk_confirm_returns_per_id_buckets`
- `test_source_delete_bulk_all_sources_when_neither_source_nor_sources`
- `test_source_delete_bulk_mutually_exclusive_source_and_sources`
- `test_source_delete_bulk_empty_sources_rejected`
- `test_source_delete_bulk_over_cap_rejected_before_resolution`

Existing single-id tests (`test_source_delete_without_confirm_previews`, `test_source_delete_with_confirm_deletes`) still pass and verify the legacy wire shape.

## Notes

- Reuses existing helpers (`coerce_list`, `resolve_sources`, `MAX_WAIT_SOURCE_IDS`, `title_for_id`, `needs_confirmation`) — no new abstractions
- Two-step confirmation mirrors `source_wait`: preview on `confirm=False`, mutate on `confirm=True`

## Summary by CodeRabbit

- **New Features**
  - Delete one source, selected sources, or all sources in a notebook.
  - Preview targeted deletions before confirming.
  - View separate results for deleted and not-found sources.
  - Prevent duplicate processing of repeated source references.
  - Confirm deletion of exactly the sources shown in the preview.

- **Bug Fixes**
  - Unknown source references no longer stop valid references from being resolved.
  - Invalid confirmation requests now return a validation error.
  - Single-source deletion retains its existing response format.


AnnasMazhar:
Friendly bump — any updates on this review? The PR is mergeable and all CI checks are green (18/18, including CodeRabbit). This implements the bulk source_delete direction you confirmed in #1995 (DeleteSources is batch-capable), keeping the single-source signature fully backwards-compatible. Happy to adjust anything.

src/notebooklm/mcp/tools/sources.py
```diff
@@ -383,30 +391,122 @@ async def source_rename(
...
             nb_id = await resolve_notebook(client, notebook)
-            src_id = await resolve_source(client, nb_id, source)
+
+            # Bulk path: ``coerced`` (explicit subset) → resolve_sources
+            # (which dedupes by resolved id); no arg → list every source.
+            # ``source=`` was short-circuited above.
+            if coerced is not None:
+                src_ids = list(dict.fromkeys(await resolve_sources(client, nb_id, coerced)))
+            else:
+                src_ids = [s.id for s in await client.sources.list(nb_id)]
```
teng-lin: **P1 — Bind confirmation to the previewed canonical IDs.** A no-argument preview captures the sources present at that moment, but the subsequent confirm call reaches this line again and lists every source currently present. I reproduced previewing `[A]`, adding `B`, and then seeing confirmation delete `[A, B]`; `B` was never shown in the confirmation preview. Require the caller to submit the previewed IDs or return an opaque confirmation token, and delete only that fixed set. The same principle applies to title/prefix refs that can resolve differently between calls.

src/notebooklm/mcp/tools/sources.py
```diff
@@ -383,30 +391,122 @@ async def source_rename(
...
+                return {"status": "deleted", "notebook_id": nb_id, "source_id": src_id}
+
             nb_id = await resolve_notebook(client, notebook)
-            src_id = await resolve_source(client, nb_id, source)
+
+            # Bulk path: ``coerced`` (explicit subset) → resolve_sources
+            # (which dedupes by resolved id); no arg → list every source.
+            # ``source=`` was short-circuited above.
+            if coerced is not None:
+                src_ids = list(dict.fromkeys(await resolve_sources(client, nb_id, coerced)))
```
teng-lin: **P1 — Derive `not_found` from one source-list snapshot before deletion.** `resolve_sources` trusts full UUIDs without checking membership, so an unknown UUID reaches `sources.delete`; that API is documented and live-verified to silently no-op, after which this tool reports the UUID as `deleted`. Conversely, a missing non-UUID ref raises here and aborts the whole batch, so valid refs are not deleted and no aggregate is returned. Resolve every ref against the snapshot, accumulate misses in `not_found`, and delete only resolved IDs. This is the required behavior from the live-RPC finding: https://github.com/teng-lin/notebooklm-py/issues/1995#issuecomment-5010890194

src/notebooklm/mcp/tools/sources.py
```diff
@@ -383,30 +391,122 @@ async def source_rename(
...
-                        "source_id": src_id,
-                        "title": title,
+                        "count": len(preview_items),
+                        "sources": preview_items,
                     }
                 )
-            await client.sources.delete(nb_id, src_id)
-            return {"status": "deleted", "notebook_id": nb_id, "source_id": src_id}
+
+            # Confirm path: there is no batch DELETE RPC on the backend, so
```
teng-lin: **P2 — This premise is outdated: `DELETE_SOURCE` supports a true multi-ID request.** The live probe in #1995 confirmed the wire shape and revised the design to add a client-level `delete_many` method and issue one RPC. The per-ID loop adds avoidable latency, throttle exposure, and partial-mutation behavior. Please implement the batch method in the typed API/backend layer and call it once here. See https://github.com/teng-lin/notebooklm-py/issues/1995#issuecomment-5010901571

src/notebooklm/mcp/tools/sources.py
```diff
@@ -383,30 +391,122 @@ async def source_rename(
...
+            deleted: list[dict[str, Any]] = []
+            not_found: list[dict[str, Any]] = []
+            failed: list[dict[str, Any]] = []
+            for i, sid in enumerate(src_ids):
+                try:
+                    await client.sources.delete(nb_id, sid)
+                    deleted.append({"source_id": sid})
+                except SourceNotFoundError as e:
+                    not_found.append({"source_id": sid, "error": str(e)})
+                except Exception as e:  # RPCError, NetworkError, etc.
```
teng-lin: **P1 — Do not convert fatal/call-wide failures into a successful tool response.** This catches network, authentication, rate-limit, and server failures, prevents `mcp_errors()` from projecting them, continues attempting every remaining ID, and then returns `status: deleted` even when `deleted_count` is zero. I reproduced a `NetworkError` producing `status=deleted`, `failed_count=2`, and two delete attempts. Fatal errors should propagate; with the required single batch RPC, an RPC failure naturally applies to the whole operation.

teng-lin (changes requested):
Requesting changes on the destructive bulk-delete path. The focused MCP suite passes both on this head and after merging onto current main (183 tests), but adversarial checks reproduce three correctness failures: confirmation can expand its target set, not_found is misclassified, and fatal transport failures are returned as status=deleted. The implementation also misses the true batch DELETE_SOURCE design confirmed in #1995.

AnnasMazhar:
Thanks — applied in 7a3c800d.

- Confirm is bound to the submitted ids (no-arg confirm is rejected so a later-added source cannot sneak in).
- `not_found` comes from one source-list snapshot; unknown UUIDs never reach delete.
- `sources.delete_many` issues one `DELETE_SOURCE` with `[[[id1],[id2],...]]`.
- Network/auth/5xx on that RPC propagate; they are not returned as `status: deleted`.

PR #2279, opened 2026-08-31
refactor(tests): split cassettes by client tier under web/

teng-lin:
## Summary

Moves the Web `batchexecute` cassette corpus from the top level of `tests/cassettes/` into `tests/cassettes/web/`, making the cassette tree symmetric with the `android/` gRPC captures. Nothing lives loose in `tests/cassettes/` any more — each client tier owns a directory.

```
tests/cassettes/
├── README.md                    (rewritten around the tier layout)
├── web/                         141 *.yaml + examples/ + gzip_coverage/
└── android/                     26 *.grpc.json (unchanged)
```

`examples/` and `gzip_coverage/` move with the corpus — they are Web fixtures. Keeping them under `web/` is what lets VCR's `cassette_library_dir` point at `tests/cassettes/web`, so **all ~365 `use_cassette("<name>.yaml")` strings in the suite stay unchanged**; the tier prefix is implied by the library dir rather than repeated at every call site.

All 144 file moves are detected by git as pure renames.

## Changes

- **`tests/vcr_config.py`** — `cassette_library_dir` → `tests/cassettes/web`.
- **Path computations** that join a bare cassette name or glob the corpus top level now resolve through `cassettes/web`: the per-test `CASSETTE_PATH` constants, shape lint, the gzip replay corpus, `check_method_coverage`, `rescrub-cassettes`.
- **Tree-wide scanners stay rooted** at `tests/cassettes/` (`check_cassettes_clean`, `test_golden_decode_coverage`, `audit_test_suite`) so they keep covering every tier, not just Web.
- **`tests/cassettes/README.md`** rewritten around the tier layout, plus cassette-path references updated across docs, guardrail failure messages, and comments.

## Three untiered constants that had to be split

The recurring hazard in this change is a single constant that means "the cassette directory" being used for two jobs at once. Each of these would have failed silently or subtly:

1. **`tests/integration/conftest.py` — `CASSETTES_DIR`.** Used both for the Web `_real_cassettes` glob *and*, since #2275, by the Android fixture as `CASSETTES_DIR / "android" / ...`. Retargeting it to `web/` sent every Android replay looking under `tests/cassettes/web/android/`. Now split into `CASSETTES_ROOT` / `WEB_CASSETTES_DIR` / `ANDROID_CASSETTES_DIR`, spelled out at each use site so the tier is never implicit. (Caught by CI — see "How this was caught" below.)

2. **`tests/scripts/check_method_coverage.py` — `CASSETTES_DIR`.** Located the Web corpus *and* excluded recordings from the test-reference grep. Narrowing it to `web/` would have let Android `*.grpc.json` captures start satisfying the reference check for RPC ids they happen to contain — a guard weakening with no failing test. Split into `CASSETTES_ROOT` (tier-wide, for the exclusion) and `CASSETTES_DIR` (the Web corpus it greps).

3. **`tests/scripts/check_cassettes_clean.py` — the default scan went vacuous.** It scanned `tests/cassettes/` non-recursively when given no arguments; with nothing left at the top level it printed `OK: no cassettes to scan` and exited 0 — a silent pass for any contributor running the guard without `--recursive`. The no-argument scan now always recurses; `--recursive` still governs explicitly-named directories. CI was never blind here (it already passes `--recursive`), and the pre-commit secrets hook matches `^tests/(cassettes|fixtures|unit/fixtures)/.*` which still covers the deeper paths.

## How this was caught

Worth recording, because the first push was green locally and red on every CI platform. The branch was cut before #2275 landed the Android gRPC cassette suite, so the `CASSETTES_DIR / "android"` use site did not exist in my working tree — the local run had nothing to break. CI tests the merge with `main` and failed 26 Android replays with paths under `tests/cassettes/web/android/`. Rebased onto `main`, split the constant, re-verified.

## Test plan

Verified after the rebase onto `main` (d737683b):

- [x] Full CI-equivalent run — `uv run pytest -n auto --dist loadgroup --cov=src/notebooklm --cov-fail-under=90`: **18192 passed, 66 skipped, 1 xfailed**, coverage **93.59%**
- [x] `uv run pre-commit run --all-files` — ruff, ruff-format, fixture-secrets all pass
- [x] `uv run mypy src/notebooklm scripts/_live_auth_scenarios --ignore-missing-imports` — clean
- [x] `check_cassettes_clean.py --strict --recursive` — 168 cassettes scanned (both tiers), 0 leaks
- [x] `check_cassettes_clean.py --strict` (no args) — now scans 168 rather than reporting "no cassettes"
- [x] `check_method_coverage.py` — 48/48 RPC methods covered, 0 grandfathered
- [x] All 7 CI Test matrix jobs green (ubuntu 3.10–3.14, macos, windows)
- [ ] Re-record one cassette under `NOTEBOOKLM_VCR_RECORD=1` to confirm it lands in `web/` (requires live auth; not run here)

## Note

`ruff format .` also wants to reformat `docs/python-api.md` and `docs/stability.md`. That drift pre-exists on `main` and is unrelated to cassettes — CI never sees it, since the pre-commit `ruff-format` hook is scoped to `types_or: [python, pyi]`. Both files are left untouched here to keep the diff scoped; worth a separate cleanup.


teng-lin:
@claude review

teng-lin:
Addressed in `a5d51c7c`: reverted the bundled `ruff format` drift in `docs/rpc-reference.md` — good catch, that was exactly the scoping inconsistency the PR body claimed to avoid. That file's diff is now 4 lines (the cassette-path fixes) instead of 665.

Audited every other changed markdown file for the same problem; all are cassette-only now:

| File | changed lines | cassette-related |
|---|---|---|
| `AGENTS.md` | 2 | 2 |
| `CONTRIBUTING.md` | 2 | 2 |
| `docs/adr/0028-*.md` | 2 | 2 |
| `docs/development.md` | 9 | 7 (+2 lines of the new tier-layout paragraph) |
| `docs/notes/pr-2102-*.md` | 2 | 2 |
| `docs/rpc-reference.md` | 8 | 8 |
| `tests/cassettes/README.md` | 73 | full rewrite (intentional) |
| `tests/integration/README.md` | 2 | 2 |
| `tests/integration/cli_vcr/README.md` | 2 | 2 |

Note this review ran against `4d53af4`, the pre-rebase commit. The branch has since been rebased onto `main` (`d737683b`) to pick up #2275's Android gRPC cassette suite, which surfaced a third untiered constant — `tests/integration/conftest.py`'s `CASSETTES_DIR`, used both for the Web `_real_cassettes` glob and by the new Android fixture as `CASSETTES_DIR / "android" / …`. Retargeting it to `web/` sent 26 Android replays looking under `tests/cassettes/web/android/`. It is now split into `CASSETTES_ROOT` / `WEB_CASSETTES_DIR` / `ANDROID_CASSETTES_DIR`. Details are in the updated PR description.

@claude review

teng-lin:
@codex review

Issue #2266, opened 2026-08-29
Python 3.14: PEP 649 deferred annotations break 10 method signatures (a method named list shadows the builtin inside its own class)


Carlos-Vera:
On Python 3.14, `uv run pytest` fails on `tests/_guardrails/test_backend_abstract_contracts.py::test_backend_implementations_preserve_abstract_method_signatures`:

```
TypeError: 'function' object is not subscriptable
src/notebooklm/_web/artifacts.py:137
```

Reproduced on `main` @ `93d325a9`, canonical extras install.

## Repro

```python
import inspect

class A:
    async def list(self) -> list[int]: ...

inspect.signature(A.list, eval_str=True)
```

```
3.13.8 -> (self) -> list[int]
3.14.0 -> TypeError: 'function' object is not subscriptable
```

## Why 3.14 flips it

PEP 649 defers annotation evaluation into a synthesized `__annotate__`, and that scope includes the class body. Before 3.14 the annotation was evaluated at `def` time, before `list` was bound as a class attribute, so it resolved to the builtin. Deferred, it resolves after the class exists. The method named `list` shadows the builtin for every annotation in the class.

## Scope

Ten methods, four classes:

| class | methods |
|---|---|
| `_notebooks.NotebooksAPI` | `get_source_ids`, `suggest_prompts`, `list` |
| `_web.artifacts.WebArtifactsAPI` | `list` |
| `_web.notebooks.WebNotebooksAPI` | `_rpc_call`, `get_source_ids`, `suggest_prompts`, `list` |
| `_web.sources.WebSourcesAPI` | `_rpc_call`, `list` |

All four modules lack `from __future__ import annotations`. The siblings that have it (`_notes.py`, `_sources.py`, `_web/notes.py`) are fine: a string annotation is eval'd against module globals, where `list` is still the builtin.

## Fix

Two shapes, both mechanical:

1. `from __future__ import annotations` in the four modules. I checked this one on `_web/artifacts.py` and it clears the failure.
2. `builtins.list[...]` at the ten sites, the spelling the rest of the client classes already use.

CI can't see it either way today: `.github/workflows/test.yml` runs 3.10 through 3.13. Adding 3.14 is what keeps it fixed.

I can send the PR with the fix and the matrix bump. Tell me which shape you want.

**Environment:** Python 3.14.0, macOS arm64, `uv sync --frozen` with `browser,dev,markdown,mcp,server,impersonate`.

teng-lin:
[...]

PR #2287, opened 2026-09-01
ci(guards): run the shadowed-builtin annotation check on every PR

Carlos-Vera:
## Summary

Follow-up to #2268 and #2270. The static checker for #2266 is in the tree and,
after #2270, precise — but nothing required runs it.
`test_class_body_annotations_do_not_name_a_shadowed_builtin` carries the
`repo_lint` marker, every compatibility cell runs `-m "not repo_lint"`, and the
only two lanes that execute `repo_lint` in bulk are the manual `repo-lint` job
(`workflow_dispatch`) and nightly. The guard written for #2266 sits in exactly
the lane that let #2266 go green.

This names it in `Run critical contract guards`, adds the matching assertion to
`test_ci_test_matrix.py`, and corrects two statements in its docstring that no
longer hold. Every claim below was checked by running it on CPython 3.14.0,
not reasoned about.

## Related Issue

No issue to close — this is the CI half of #2266, left over from #2268 (the
checker) and #2270 (its precision).

## Changes

**`.github/workflows/test.yml`** — one node id added to `Run critical contract
guards`:

| Lane | Before | After |
| --- | --- | --- |
| Canonical PR cell (3.12) | not run — `-m "not repo_lint"` deselects it | runs, named by node id |
| Compatibility cells (3.10–3.14) | not run | not run (unchanged) |
| Manual `repo-lint` + nightly | whole file | whole file (unchanged) |

That step passes node ids and no `-m`, so the marker does not deselect a test
named explicitly. Only the tree-scanning check is promoted; the regression
fixtures in the same file stay in the bounded manual lane, where their cost
belongs.

**`tests/unit/test_ci_test_matrix.py`** — the matching assertion, so the wiring
is a contract rather than a line that can be dropped unnoticed. That test
already runs in the same step, so it guards itself.

**`tests/_guardrails/test_builtin_shadowed_annotations.py`** — docstring only:

| Statement | Reality | Change |
| --- | --- | --- |
| "``builtins.list[...]`` is spelled explicitly at ~40 sites" | `git grep -o 'builtins\.list\[' -- src/notebooklm \| wc -l` → 138 before #2268, 150 when it merged, **253** today | states the convention without a number; a pinned count rots in a tree moving this fast |
| "Adding a 3.14 lane would close today's hole on one version" | reads as if the matrix had no 3.14 cell; it has had one since `068f876b` (`test.yml:202`). What is missing is a 3.14 **`repo_lint`** lane | says `repo_lint` lane |

The paragraph also now records where the check runs, since that is no longer
"nowhere required".

## Test Plan

- [x] Tests pass — 3.14.0, `-n auto --dist loadgroup --cov-fail-under=90`: 18,172 passed, 81 skipped, 1 xfailed, 92.98%. `tests/_guardrails` and `tests/unit/test_ci_test_matrix.py` included, `repo_lint` and all
- [x] Linting and formatting pass — `pre-commit run --all-files`
- [x] Type checking passes — `mypy src/notebooklm --ignore-missing-imports`, no issues in 418 source files
- [ ] If this PR changes architectural shape, an ADR has been added or updated. *(N/A — one node id, one assertion, one docstring.)*

## Notes

The wiring does what it says. Reintroduce a single bare `list[...]` in
`_web/artifacts.py` and run the two commands CI actually runs, on the same
dirty tree:

```
# the guard step with this PR applied — catches it
$ pytest -n auto --dist worksteal --timeout=180 --no-cov <the step's node ids>
E  AssertionError: class-body annotation names a builtin the class rebinds ...
E      src/notebooklm/_web/artifacts.py:119 (WebArtifactsAPI shadows 'list')
1 failed

# a compatibility cell — the behaviour before this PR
$ pytest -m "not repo_lint and not requires_playwright and not requires_chromium" \
    tests/_guardrails/test_builtin_shadowed_annotations.py
32 deselected in 0.03s
```

On a clean tree the exact guard-step command, including the new node id, is
`23 passed in 21.70s` — about 22 seconds on the canonical cell for a check that
would have caught #2266 on the PR that introduced it.

Happy to reshape any of this if you'd rather it looked different.

---

by Carlos Vera  
carlos@braveslab.com<br>
<em>“**For I can do everything through Christ, who gives me strength.**”</em>
<em>Phil 4:13 | NLT</em>

## Summary by CodeRabbit

* **Tests**
  * Added coverage to ensure class-body annotations do not shadow built-in names.
  * Included this validation in the required critical contract checks.
  * Updated CI matrix verification to confirm the new guard runs as expected.

* **Documentation**
  * Clarified annotation usage, lint protections, and how the guard is executed across test environments.


Carlos-Vera:
The `windows-latest` / 3.12 red is not from this PR.

One test fails — `tests/unit/android/test_source_upload.py::test_error_or_wait_timeout_never_dispatches_title_mutation` — asserting `SourceProcessingError` and getting `SourceTimeoutError`. It is a clock race inside `wait_until_registered`, nothing to do with the guard wiring: the base commit `453e8ca9` is green on that cell, and this diff is one node id in a step gated on `matrix.canonical`, one assertion, and docstring prose.

Cause and fix in the next comment.

Carlos-Vera:
Cause, verified.

`wait_until_registered` checks the deadline before the first `get_source`. On Windows before 3.13, `time.monotonic()` is `GetTickCount64()` and moves in 15.6 ms steps ([3.13 moved it](https://docs.python.org/3/whatsnew/3.13.html#time) to `QueryPerformanceCounter()`), so one tick between `RuntimeDeadline.start(0.01)` and that check burns the 10 ms budget before anything is polled. The timeout then carries `last_status=None` — which is why the CI message stops at `not ready after 0.0s` with no status suffix.

Reproduced without a Windows runner, with a clock that crosses one tick:

    monotonic = iter([0.0, 0.0156]).__next__

    coarse -> SourceTimeoutError: Source …0201 not ready after 0.0s | polls=0 last_status=None
    fine   -> SourceProcessingError: Source …0201 failed to process | polls=1

Fix in #2288 — `wait_timeout` 0.01 → 0.05 in that test. The deeper fix, one guaranteed look before any timeout is declared, breaks three tests that pin the current order on purpose, so I left that call to you.

teng-lin:
Thank you, @Carlos-Vera — this is exactly the kind of follow-through the repo needs. You didn't stop at "the checker exists": you traced why it never ran on the PR that introduced #2266, promoted only the tree-scanning half, pinned the wiring in `test_ci_test_matrix.py` so it can't be dropped silently, and every claim in the docstring rewrite checks out against `test.yml` and `nightly.yml`. The `-m` reasoning (explicit node ids aren't deselected by a marker the step never filters on) is right, and verified.

Your diagnosis of the Windows red on this PR was also spot on — #2288 is merged, and the refreshed run here is green on every cell, so merging now.

Two follow-ups came out of reviewing this pair, neither blocking you:
- #2293 corrects `CONTRIBUTING.md` / `docs/development.md`, which still claimed "CI runs everything by default, so `repo_lint` failures still block merge" — the exact falsehood this PR's docstring refutes.
- The Android `_wait_uploaded_source` loop gets a poll-first rewrite with an injectable clock, so the deeper fix you described in #2288's notes lands without touching the web `SourcePoller` tests you (correctly) didn't want to disturb.

One thought for later, no action needed: the whole `test_builtin_shadowed_annotations.py` file runs in ~0.45 s, so dropping its `repo_lint` marker would have run it on every matrix cell (3.10–3.14, Windows, macOS) without any node-id plumbing. Four of the seven guards already promoted in that step carry no marker. Worth considering the next time you touch it.

PR #2321, opened 2026-09-02
test: raise coverage on the lowest-covered modules from the nightly job

teng-lin:
Raises unit coverage on the lowest-covered non-generated modules from the
nightly coverage job — [run 33670880090](https://github.com/teng-lin/notebooklm-py/actions/runs/33670880090/job/100383961823),
`release/v0.8.2` @ `5c8bac5d`.

**Stacked on `release/v0.8.2`** (#2318). GitHub will retarget this to `main`
when that merges; no rebase needed unless it conflicts.

## Result

Reproducing the nightly's two-pass command locally (xdist pass + serial
Playwright `--cov-append`):

| | baseline (CI, same commit) | this branch |
|---|---|---|
| total coverage | 93.41% | **94.15%** |
| uncovered (statements + partial branches) | 3635 | **3239** (−396) |

Only two files land above their baseline uncovered count — `_auth/recovery.py`
(+4) and `_artifact/downloads.py` (+1), neither touched here; that is local-vs-CI
environment noise, not a regression.

## Per-module

| module | before | after |
|---|---|---|
| `_artifact/__init__.py` | 51% | 100% |
| `cli/_cookie_import.py` | 62% | 100% |
| `_web/transport/middleware/semaphore.py` | 71% | 100% |
| `_version_info.py` | 72% | 88% |
| `_source/__init__.py` | 76% | 100% |
| `_web/artifact/generation.py` | 76% | 99% |
| `_android/artifact_outputs.py` | 77% | 96% |
| `_android/codecs/documents.py` | 81% | 100% |
| `mcp/tools/_content_sanity.py` | 81% | 100% |
| `_android/labels.py` | 82% | 100% |
| `_android/collections.py` | 82% | 100% |
| `_android/codecs/notes.py` | 83% | 100% |
| `_app/download_specs.py` | 84% | 100% |
| `_source/markdown.py` | 84% | 100% |
| `cli/services/polling.py` | 84% | 100% |
| `_android/codecs/chat.py` | 85% | 100% |
| `_android/upload.py` | 85% | 89% |

The new cases target rejection and failure paths, not line count: range-admission
and structural-depth guards in the TailwindDoc codec, `generate_video`'s
option-compatibility rules, the organization adapters' post-write read-back and
`NOT_FOUND`-vs-other error split, `write_text_atomic`'s cancellation and
partial-file handling, the Drive metadata gate, and every registry invariant in
`_app/download_specs.py`.

## Things worth a reviewer's attention

[...]

PR #2327, opened 2026-09-02
test: cover the Android transport, bearer provider, and module entry point

teng-lin:
Second coverage tranche, continuing #2321/#2324 against the nightly job's per-file table. Branched from `main` (not a release branch, per the lesson from #2324).

## Result

| | before | after |
|---|---|---|
| total coverage | 94.16% | **94.35%** |
| uncovered (statements + partial branches) | 3235 | **3134** (−101) |

| module | before | after |
|---|---|---|
| `_android/auth.py` | 86% | **100%** |
| `_android/session.py` | 90% | **97%** |
| `_android/assets.py` | 90% | **94%** |
| `_android/codecs/__init__.py` | 35% | **100%** |
| `__main__.py` | 0% | **100%** |

No production changes — tests only.

## What these cover

The theme is credential-adjacent transport paths: the bearer provider's lifecycle, the gRPC session's epoch fencing, and the asset downloader's URL admission policy. Each is a place where a permissive or unexercised branch fails open rather than loud.

- **`_android/auth.py`** — the loop-binding assertions, `reset_after_open`'s task disposal, the profile-read failure classification (an unreadable profile must not leak its `OSError` into the auth path), `_mint_once`'s failure taxonomy, the expiry-margin arithmetic, and three lock/generation races: a mint landing after a close, a waiter cancelled while re-acquiring the lock, and a mint settling after its last waiter left.
- **`_android/session.py`** — timeout resolution (unset/infinite/NaN dispatch with no deadline at all), epoch and loop fencing, `prepare_metadata`'s frame detaching, deadlines reached while minting / inside the augmentor / mid-stream, queue-admission timeouts surfacing as DEADLINE_EXCEEDED, and stream teardown (unmapped wire errors logging only their type, a `cancel()` that itself raises, a call with no `cancel` hook, idempotent `close_resources`).
- **`_android/assets.py`** — the host allowlist's rejections (bare suffix that is not a subdomain, suffix appearing mid-host, percent-encoded and backslash dots, malformed labels), URL admission (scheme, non-default port, embedded credentials, fragments, control characters), `alr` handling, bearer scoping to the two hosts that require it, and the close sweep's failure absorption.
- **`__main__.py`** — nothing covered `python -m notebooklm` before. Renaming `notebooklm_cli.main` would have broken the documented invocation with the suite still green. Pinned both in-process (via `runpy`, so coverage sees the guarded branch) and end-to-end through a real interpreter.
- **`_android/codecs/__init__.py`** — every lazy alias, plus a subprocess import-delta proving the point of the indirection: importing the codec package pulls in no generated protobuf, and resolving an alias is what triggers the fan-out.

## Notes for review

1. **`log.py` was reported at 0% but is not a gap.** `tests/_guardrails/test_public_surface_manifest.py` covers it to 100%; the coverage job excludes `-m repo_lint`, so it reads as zero. Left alone rather than padded with a duplicate test. Worth knowing when reading that table — other entries may be the same artifact.

2. **Three branches are deliberately left uncovered, with comments.** The `(KeyboardInterrupt, SystemExit)` arms inside task loops cannot be exercised from a real `asyncio.Task`: CPython's `Task.__step` re-raises both into the event loop before any awaiter observes them, which tears down the pytest session rather than being catchable. Where the same guard was reachable by awaiting the coroutine directly (`_mint_once`), it is tested that way instead.

3. **Two tests correct my initial reading of the code, not the code.** `sanitize_escaping_exception` does not rewrite the message — it drops the traceback and chain so credential-bearing frames are unreachable, and the test asserts that. `_declared_content_length` returns `-1` for "declared but unusable", deliberately distinct from `None` for "not declared"; the test pins both.

4. **`_android/session.py` tests live in the existing `test_session.py`** rather than a sibling module, because `tests/unit/android/` has no `__init__.py` and the fakes there would otherwise have to be duplicated and could drift.

## Verification

```
  -m "not repo_lint and not requires_playwright and not requires_chromium"   # 17337 passed
                                                                             # 84 passed
```

## Summary by CodeRabbit

- **Tests**
  - Expanded automated coverage for Android asset downloads, authentication, session handling, URL validation, credential lifecycle, and cleanup behavior.
  - Added coverage for lazy codec loading and package-level alias behavior.
  - Added end-to-end checks for the `python -m notebooklm` command, including help output and invalid-command handling.
  - Improved validation of timeouts, cancellation, error handling, redirects, resource safety, and deadline calculations.

### DIFF
--- a/tests/unit/android/test_asset_url_policy.py
+++ b/tests/unit/android/test_asset_url_policy.py
@@ -1,0 +1,445 @@
+"""Host and URL admission policy for Android asset downloads.
+
+Every one of these helpers sits in front of an outbound request carrying a
+bearer token, so a permissive branch here is an SSRF/credential-leak shape, not
+a cosmetic gap. The adapter suites exercise the happy path; these pin the
+rejections exhaustively.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import os
+from pathlib import Path
+
+import pytest
+
+from notebooklm._android.assets import (
+    _MAX_APPLICATION_REDIRECT_BYTES,
+    _append_initial_alr,
+    _bearer_for,
+    _bounded_text,
+    _close_clients_and_settle_tasks,
+    _declared_content_length,
+    _default_client_factory,
+    _fsync_directory,
+    _is_android_download_host,
+    _safe_approved_host,
+    _single_location,
+    _validated_host,
+)
+from notebooklm._android.auth import BearerCredential
+
+BEARER = "ya29.asset-secret"
+APPROVED = "lh3.googleusercontent.com"
+
+
+# ---------------------------------------------------------------------------
+# _is_android_download_host
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "host",
+    [
+        pytest.param(APPROVED, id="primary-bearer-host"),
+        pytest.param("contribution.usercontent.google.com", id="contribution-host"),
+        pytest.param("r1---sn-abc.googlevideo.com", id="googlevideo-subdomain"),
+        pytest.param("usercontent.google.com", id="usercontent-subdomain-of-google"),
+    ],
+)
+def test_approved_download_hosts_are_admitted(host: str) -> None:
+    assert _is_android_download_host(host) is True
+
+
+@pytest.mark.parametrize(
+    "host",
+    [
+        pytest.param("", id="empty"),
+        pytest.param("evil.com", id="unrelated"),
+        pytest.param("googlevideo.com", id="bare-suffix-is-not-a-subdomain"),
+        pytest.param("notgooglevideo.com", id="suffix-without-the-dot"),
+        pytest.param("evil.com.googlevideo.com.attacker.net", id="suffix-in-the-middle"),
+        pytest.param("x" * 254, id="over-253-characters"),
+        pytest.param("lh3.googleusercontent.com:443", id="port-in-the-host"),
+        pytest.param("LH3.GOOGLEUSERCONTENT.COM", id="uppercase-is-not-normalised-here"),
+        pytest.param("evil%2egooglevideo.com", id="percent-encoded-dot"),
+        pytest.param("evil\\.googlevideo.com", id="backslash"),
+        pytest.param("a..googlevideo.com", id="empty-label"),
+        pytest.param("-a.googlevideo.com", id="label-starts-with-hyphen"),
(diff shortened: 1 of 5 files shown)


teng-lin:
Pushed f6a3cf75 — all four findings addressed, and all four were correct.

**Windows CI (codex P1)** — you diagnosed it exactly. `os.open(tmp_path, os.O_RDONLY)` raises on Windows, so `_fsync_directory` returned before touching the mocked `os.fsync`/`os.close` and `closed` stayed empty. The production code is right (its docstring already scopes itself to platforms exposing directory fsync); the test was platform-naive. Rather than skip on Windows I stubbed `os.open` to hand back a real descriptor, so the release-on-failure path is now exercised identically everywhere.

**Vacuous master-token assertion (codex P2)** — correct, and this was the most useful catch. `_cached` is `None` on that path regardless of whether `record = None` runs, so the test proved nothing about the cleanup it claimed to cover. It now walks the escaping exception's traceback to `_mint_once`'s frame and asserts `record` is cleared there — which is exactly what that line is for, since an escaping exception keeps its frames alive and the master token would stay reachable through them. I verified it bites by deleting `record = None` from both handlers: the cases fail, and pass again once restored.

**Tautological clock assertion (codex P2)** — correct. `deadline.timeout == 5.0` holds either way. Now drives the injected clock forward and asserts `started_at`, `elapsed()`, `remaining()` and `expired()` all track it, plus that `monotonic` is not the global one.

**Reload identity (coderabbit)** — correct, `is not None` would pass for any wrong object. Now asserts identity with `sources.decode_source`.

One thing worth flagging from my own process: my first attempt at these edits used an unbounded string replacement that silently deleted 9 tests from `test_session.py`. The suites still passed, because deleted tests simply don't run. I caught it on the commit diffstat (329 deletions for what should have been a small fix), reverted, and redid the edits with both ends bounded. The pushed diff is `+68/−21` across 4 files with the per-file test counts unchanged, and touches no `src/`.

teng-lin:
Green — 19/19 checks pass, `mergeStateStatus: CLEAN`.

The one job that failed after the review fixes was `Test (windows-latest, Python 3.11)` with `OSError: [WinError 10055] ... system lacked sufficient buffer space or because a queue was full` in `tests/unit/test_source_batch_add_service.py` — a file this PR does not touch, failing on 1 of 5 Windows jobs while the other four passed on the same commit. That is socket/handle exhaustion on the runner, not a logic failure, so I re-ran the failed job rather than "fixing" it; it passed clean.

Ready to merge whenever you are.

PR #2269, opened 2026-08-29
feat(android): close public API parity gaps

teng-lin:
## Summary

- select Android adapters for all 11 typed namespaces and close the remaining public API implementation gaps
- add exact, live-qualified protobuf contracts for artifacts, chat, notebooks, notes, organization, sharing, sources, and account settings
- preserve only two narrow Web compatibility seams for mobile routes that reject valid owned resources
- consolidate Android evidence documentation and deterministic protobuf fixtures

## Public API status

- 145 typed public callables audited
- 143 Android-selected or local-over-Android
- 2 narrow Web compatibility operations: `notebooks.remove_from_recent` and `sharing.set_view_level`
- 0 public unsupported or `_reject` paths

## Compatibility boundaries

- `notebooks.remove_from_recent`: the exact mobile route returns `INTERNAL` without removing the notebook
- `sharing.set_view_level`: the recovered mobile mutation returns `PERMISSION_DENIED` for owned notebooks

## Test Plan

- [x] Full suite: 17,783 passed, 80 skipped, 1 xfailed
- [x] Android-focused suite: 726 passed, 1 skipped
- [x] `uv run ruff check .`
- [x] `uv run mypy src/notebooklm`
- [x] `uv run python scripts/regenerate_android_protos.py --check`
- [x] `uv run pre-commit run --all-files`
- [x] `git diff --check`

## Notes

Live probes used disposable notebooks and restored or deleted scratch resources during cleanup.

PR #2332, opened 2026-09-02
test: cover status projection, ref parsing, transport, and server request policy

teng-lin:
Third coverage tranche, over the **non-`_auth`** modules from the current per-file table. The `_auth/*` cluster is excluded — it's mid-refactor.

Branched from `main`, independent of #2327 (disjoint files), so the two can merge in either order.

## Result

| | before | after |
|---|---|---|
| total coverage | 94.16% | **94.45%** |
| uncovered (statements + partial branches) | 3235 | **3099** (−136) |

| module | before | after |
|---|---|---|
| `_android/errors.py` | 76.5% | **98.8%** |
| `_source/drive.py` | 79.2% | **100%** |
| `mcp/tools/_studio_items.py` | 83.8% | **98.5%** |
| `_web/research.py` | 84.8% | **97.5%** |
| `_curl_cffi_transport.py` | 87.6% | **94.8%** |
| `_runtime/call_supervisor.py` | 89.1% | **94.6%** |
| `server/app.py` | 91.5% | **96.6%** |

## What these cover

The theme is input-admission and status-projection boundaries — the places where a permissive branch silently widens what the client accepts.

- **`_android/errors.py`** — `grpc_status` is the only thing between a dependency-controlled exception object and public error text. Every rejection is pinned: unknown/future names, non-string names, out-of-range and stringly-typed values, an empty tuple value, a `code()` that raises, a non-callable `code`, and name-beats-value precedence. Plus: the projection never carries the original message.
- **`_source/drive.py`** — admitted reference forms, and the rejections that matter: the `drive.google.com.evil.test` suffix attack, and an untrusted host carrying a well-formed query id.
- **`mcp/tools/_studio_items.py`** — the rule that makes `studio_delete` safe: a full UUID never falls through to title matching, so a note *titled* like a UUID cannot absorb a delete aimed at that id. Plus prefix ambiguity, the hex-title fallthrough, and kind scoping.
- **`_curl_cffi_transport.py`** — `delete()` was **entirely untested**, and its own docstring says it exists because its absence silently left staged files in the user's Drive under `NOTEBOOKLM_TRANSPORT=curl_cffi`. Now covered end-to-end against the local test server, along with `stream_upload` (disk streaming, caller-owned handles, progress callbacks, transport failures).
- **`_runtime/call_supervisor.py`** — the admission-generation state machine: epochs must move forward, a prepared generation can't be replaced, transitions are refused from the wrong state, and drain hooks all run even when one fails (with interpreter-exit precedence).
- **`server/app.py`** — content-type recognition by structured suffix rather than substring, `Content-Length` parsing where anything but a non-negative integer must refuse the read, body-limit table precedence, and the stale-auth startup projection.
- **`_research.py`** — cross-task provenance refusal, asserted to happen before any RPC is dispatched.

## One deletion

**`WebResearchAPI._web_wait_for_completion` (106 lines) is removed.** It has no caller in `src/`, `tests/`, or docs. `d5df15e7` ("feat(android): close public API parity gaps", #2269) hoisted the wait loop into `BaseResearchAPI._wait_for_completion` and left this copy orphaned. The Web subclass still contributes its live `_wait_observed_status` override, and the existing `test_wait_for_completion_*` cases exercise the inherited path — including the ambiguous-first-poll raise. Two wait implementations that can drift is worse than one, so it's deleted rather than tested. That accounts for most of `_web/research.py`'s jump.

## Two things I deliberately did *not* do

1. **`_sources.py` (19 uncovered, 85.0%) is skipped.** All 19 lines are `raise NotImplementedError` bodies of `@abstractmethod`s on an ABC — structurally unreachable, since Python refuses to instantiate `SourcesAPI` and every backend overrides them. Testing them would be coverage theatre. If you want the number moved, the right lever is a coverage `exclude_lines` policy for `raise NotImplementedError`, which is a config decision rather than something to slip into a test PR.

2. **`log.py`-style false positives.** As flagged on #2327, some 0%/low entries in that table are measurement artifacts of the coverage job excluding `-m repo_lint`, not real gaps. Worth keeping in mind when reading the table.

## Verification

```
  -m "not repo_lint and not requires_playwright and not requires_chromium"   # 17335 passed
                                                                             # 84 passed
```

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for request admission, media types, content lengths, and startup error reporting.
  * Added validation tests for Android gRPC status handling, Studio reference matching, call supervision, generation lifecycles, and drain hooks.
  * Added tests for transport failures, streaming uploads, research source verification, and Google Drive reference validation.

* **Maintenance**
  * Removed unused research completion polling logic.

### DIFF
--- a/src/notebooklm/_web/research.py
+++ b/src/notebooklm/_web/research.py
@@ -39,7 +39,6 @@ from ..exceptions import (
     NetworkError,
     RateLimitError,
     ResearchStartUnavailableError,
-    ResearchTimeoutError,
     RPCError,
     RPCTimeoutError,
     ServerError,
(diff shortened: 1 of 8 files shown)


teng-lin:
@claude review

teng-lin:
Pushed 3870c726 — all findings addressed, and every one was correct. They were the same class of defect: the assertion would still pass if the behaviour it named regressed, because the local test server returned 200 regardless of what the client actually sent.

**Header / cookie forwarding (codex P2 ×3, Copilot).** `do_DELETE` and `do_PUT` now echo the `Authorization`, `Content-Type`, `Cookie` and `X-Goog-Upload-Command` they received, and the tests assert on those echoes. Injection-proven: deleting `CurlOpt.HTTPHEADER` and `CurlOpt.COOKIE` from `stream_upload` fails the header and cookie cases, and passes again once restored.

**Incremental progress (codex P2).** Correct — summing the callbacks passes even if the file is buffered and reported once. Now a 512 KiB payload with `len(seen) > 1` plus every count positive.

**Non-buffering (codex P2).** Same payload change; the 4 KiB original could not distinguish streaming from buffering at all.

**Failed-`__aenter__` cleanup (codex P2).** Correct — `client.aclose()` was doing the tidying, so removing the explicit cleanup would have gone unnoticed. The curl stream context is now instrumented and the test asserts its `__aexit__` ran. Injection-proven: deleting the cleanup from `_StreamCtx.__aenter__` fails it.

**Studio kind vocabulary (Copilot).** Correct — items carry `"note"` or an artifact's *hyphenated* kind (`slide-deck`, `mind-map`), never a generic `"artifact"`. The scoping cases now use the real vocabulary via named constants.

Diff is `+68/−17` across the two test files; no `src/` changes.

teng-lin:
Pushed b76882f3 — second review round addressed, all 17 threads resolved.

**CodeQL: HTTP response splitting.** Real, and self-inflicted: the echo headers I added in the previous round reflected client-controlled request values into response headers verbatim. The test server now strips CR/LF and bounds the length before echoing. Good catch — I introduced this while fixing the *previous* round's "assertions don't actually verify anything" findings.

**Deterministic refusal (coderabbit).** Correct — port 1 is only *usually* free, and a local service binding it would turn a transport-failure test into a false pass. New `refused_url` fixture binds an ephemeral loopback socket without listening and holds it open for the test.

**Upload content integrity (coderabbit).** The server now echoes the body's leading bytes, so a corrupted or truncated read function is caught rather than only a wrong `Content-Length`.

**Task-id assertion (coderabbit).** Correct — `assert_awaited()` would pass if a different task id were sent. Now asserts `await_args.args[1] == "task-1"`.

**Comment said the opposite of the code (coderabbit).** Also correct, and worth fixing: the admission guards raise rather than returning a sentinel, so they fail *closed* and loudly. My wording claimed "fails open".

**Empty studio ref (coderabbit).** This one changed my answer rather than confirming it. Substituting `""` made the test fail — because a missing title normalises to `""`, so an empty ref genuinely *does* match an untitled item at the matcher level. That is safe only because `resolve_studio_item` runs `validate_id` first and rejects blank refs before the title path. Rather than assert a guard `_match_studio_ref` does not have, I pinned both facts: the matcher's actual behaviour as explicit characterization (with a note that direct callers must do their own rejection), and the upstream validation that makes it safe today.

The earlier codex/Copilot threads about header forwarding, incremental progress, non-buffering and `__aenter__` cleanup were fixed in 3870c726 and injection-proven; resolving those too.

PR #2335, opened 2026-09-03
test: cover the lifecycle drain path, Drive streaming, and artifact admission

teng-lin:
The last three modules from the non-`_auth` list. Branched from `main`; disjoint from #2332 and #2334, so all three can merge in any order.

## Result

| module | uncovered | before | after |
|---|---|---|---|
| `_web/sources/drive_import.py` | 35 → **13** | 90.8% | **96.6%** |
| `_android/artifacts.py` | 40 → **33** | 91.9% | **93.3%** |
| `_runtime/lifecycle.py` | 34 → **28** | 93.7% | **94.8%** |

Total on this branch: **94.41%**, uncovered 3099.

## What these cover

- **`drive_import.py`** — `_filename_from_disposition` (RFC 5987 precedence over a plain `filename=`, the optional language tag, and the misses); `_read_capped_text`'s cap slicing, asserted by *counting chunks requested* so "a body far larger than the cap must not drag in another 64 KiB" is actually verified rather than assumed; the pre-body `content-length` gate; the non-HTML 4xx hard error; and `_drain_to_temp`'s writer thread — its failure is carried back to the event loop through the one-slot exception box (without it the caller sees a silently truncated download), and the producer stops feeding a failed writer instead of blocking forever on a queue nobody drains.
- **`lifecycle.py`** — `drain`'s admission fencing (no-op when closed, negative-timeout rejection, resources left open), the close-race suppression that must not swallow an unrelated supervisor fault, and `open`'s state guards including that an admission-rollback failure does not mask the original open failure.
- **`_android/artifacts.py`** — the input admission that runs before any RPC is dispatched: audio needs a source and text instructions, an explicitly empty `source_ids` is a caller error for each supported family (`None` still means "all sources"), and the slide-deck / interactive-app downloads reject unknown output formats. Each asserts nothing reached the wire.

## One test-fake change

The lifecycle fake gains a one-shot `stop_errors` queue. `close()` also calls `stop_accepting`, so a persistent `stop_error` breaks teardown — which is why the drain-vs-close race could not be expressed before.

## Where the remaining lines are

Deliberately stopped rather than padded:

- `_android/artifacts.py`'s remaining 33 are download-time error paths that need the asset-transfer graph wired per case — real work, not a quick win.
- `lifecycle.py`'s remaining 28 are cancellation interleavings inside the close wave, several of which need two waves racing at a specific await point.
- `drive_import.py`'s remaining 13 are the queue-saturation arms of `_drain_to_temp`'s error handler, which need a writer slow enough to fill an 8-slot queue while the producer is already unwinding.

Also worth repeating from #2332: `_sources.py` (19 uncovered) is entirely `@abstractmethod` bodies and is not worth testing — if you want that number moved, the lever is a coverage `exclude_lines` policy for `raise NotImplementedError`.

## Verification

```
  -m "not repo_lint and not requires_playwright and not requires_chromium"   # 17383 passed
                                                                             # 84 passed
```

Tests only — no `src/` changes.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for input validation across generation, slide downloads, quizzes, and flashcards.
  * Added lifecycle tests for shutdown, reopening, admission control, timeouts, and failure recovery.
  * Added Drive import coverage for filename handling, response-size limits, streaming failures, back-pressure, client errors, and redirects.
  * These tests help ensure invalid requests fail safely and lifecycle or import issues are handled predictably.

### DIFF
--- a/tests/unit/android/test_artifacts.py
+++ b/tests/unit/android/test_artifacts.py
@@ -2788,3 +2788,79 @@ def test_artifact_source_uses_the_imported_exact_package_source_id() -> None:
         ),
     )
     assert request.artifact.sources[0].source_id.id == "source-1"
+
+
+# ===========================================================================
+# Input admission before any RPC is dispatched
+# ===========================================================================
+
+
+@pytest.mark.asyncio
+async def test_audio_generation_requires_at_least_one_source() -> None:
+    session, _nb, _mm, _assets, api = _graph()
+
+    with pytest.raises(ValidationError, match="at least one source id"):
+        await api.generate_audio("notebook-1", source_ids=[])
+
+    assert session.calls == []
+
+
+@pytest.mark.asyncio
+async def test_audio_instructions_must_be_text() -> None:
+    session, _nb, _mm, _assets, api = _graph()
+
+    with pytest.raises(ValidationError, match="instructions must be a string or None"):
+        await api.generate_audio("notebook-1", source_ids=["s1"], instructions=7)
+
+    assert session.calls == []
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("method", "label"),
+    [
+        pytest.param("generate_video", "Video", id="video"),
+        pytest.param("generate_report", "Report", id="report"),
+    ],
+)
+async def test_a_supported_family_requires_at_least_one_source(method: str, label: str) -> None:
+    """An explicitly empty list is a caller error; ``None`` means "all sources"."""
+    session, _nb, _mm, _assets, api = _graph()
+
+    with pytest.raises(ValidationError, match=f"{label} generation requires at least one"):
+        await getattr(api, method)("notebook-1", source_ids=[])
+
+    assert session.calls == []
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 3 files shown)


teng-lin:
@codex review
@claude review

teng-lin:
Pushed 97e5a4a2 — Windows CI fixed.

Both `_drain_to_temp` writer-failure tests asserted `IsADirectoryError`, which is the POSIX result of opening a directory; Windows raises `PermissionError` for the same call. My assertion was platform-specific for no reason — which errno the platform picks is not the contract, only that the writer thread's failure reaches the caller through the one-slot exception box.

Now asserts `OSError`. That still discriminates: `ValidationError` (the cap and empty-body guards) does not inherit from `OSError`, so a swallowed writer error would surface as the wrong type or as no error at all. Injection-proven — deleting the exception box fails both cases, restoring it passes them.

Worth flagging: this is the second POSIX-only assumption I've written into a Windows-visible test in this series (the first was assuming a directory file descriptor is obtainable, in #2327). The comment now names both platforms explicitly so the next reader doesn't re-narrow it.

PR #2338, opened 2026-09-03
test: close the remaining large coverage gaps outside _auth (15 modules, 11 to 100%)

teng-lin:
Closes the last large coverage gaps outside `_auth/*`. **Tests only — no `src/` changes.**

Produced by nine agents working in isolated worktrees (each with its own venv, to avoid the split-brain where a shared venv silently imports the wrong tree), then merged and re-verified centrally.

## Result

**11 of 15 target modules reach 100%; the rest 99%+.**

| module | before | after |
|---|---|---|
| `_android/upload.py` | 88.6% | **100%** |
| `_android/sources.py` | 89.9% | **100%** |
| `_web/sources/upload.py` | 90.4% | **100%** |
| `_android/codecs/sources.py` | 88.8% | **100%** |
| `_android/research.py` | 89.5% | **100%** |
| `_android/codecs/notebooks.py` | 89.4% | **100%** |
| `_android/drive_staging.py` | 90.4% | **100%** |
| `_artifact/polling.py` | 90.9% | **100%** |
| `_web/transport/cookie_persistence.py` | 94.6% | **100%** |
| `_web/transport/lifecycle.py` | 91.2% | **100%** |
| `cli/services/login/chromium_accounts.py` | 89.5% | **100%** |
| `_web/sources/batch.py` | 92.6% | 99.5% |
| `_android/assets.py` | 90.2% | 99.3% |
| `_research.py` | 89.7% | 99.2% |
| `cli/skill_cmd.py` | 93.9% | 99.1% |

Branch total: **95.64%**, uncovered 3235 → 2458. (That total also reflects #2327, already on `main`.)

`_android/upload.py` was the single largest gap in the codebase — 102 uncovered lines in the resumable-upload state machine — and is now fully covered.

## What these actually pin

Not line-count padding. The recurring themes are the contracts where a silent failure is expensive:

- **Three-way outcome discipline** (`confirmed` / `refused` / *unconfirmed*). An unconfirmed write must never look safely retryable — retrying a source commit duplicates it, and an unconfirmed share may already have sent an invite email.
- **Cancellation safety** — the upload pipeline's `_settle_context_exit`, the asset close sweep, and the Drive writer thread, where the failure mode is a leaked temp file, an orphaned authenticated upload, or a silently truncated download.
- **Fail-closed host and URL admission** on the download paths (SSRF boundaries).
- **Drift vs absence** — a reshaped server payload must raise, not read as "no data".

## Verification

Each agent's work was independently re-measured and audited by the integrator, not taken on report. For the security- and correctness-critical contracts I broke the source and confirmed the tests fail — e.g. neutralising both `mark_unconfirmed` call sites in `_android/sources.py` fails 4 tests; neutralising the asset host allowlist fails 7.

```
  -m "not repo_lint and not requires_playwright and not requires_chromium"   # 17980 passed
                                                                             # 84 passed
```

## One known flake, pre-existing and unrelated

`tests/unit/test_auth_mint_service.py::test_cancelled_mint_returns_before_worker_settles_and_worker_retains_only_arguments` failed intermittently during local verification. It is **untouched by this PR** and passes 10/10 in isolation on this branch. Its assertion is `assert release.wait(2.0)` — a 2-second wall-clock bound on a thread handoff — and every failure occurred while nine agents were running full test suites concurrently on the same machine (runs of 200–360s; the same suite passes at 96s under normal load). Worth tightening separately; not a blocker here.

## Notable technique

One agent solved a constraint I had written off. `KeyboardInterrupt` raised inside an `asyncio.Task` is re-raised into the event loop by `Task.__step` and tears down the pytest session, so those handlers looked untestable. It added a helper that steps the coroutine directly with `coro.send(None)` — never entering a task — with a guard that fails loudly if the coroutine actually suspends. That made the process-interrupt handlers testable across six entry points.

## Summary by CodeRabbit

- **Tests**
  - Expanded Android coverage for uploads, downloads, Drive staging, research workflows, and notebook/source decoding.
  - Added safeguards for cancellation, shutdown, malformed responses, authentication, rate limits, retries, partial files, and resource cleanup.
  - Added CLI coverage for Chromium account discovery, cookie privacy, deduplication, and packaging behavior.
  - Strengthened polling, cookie persistence, source reconciliation, batch imports, and web transport lifecycle validation.


teng-lin:
@codex review. @claude review

teng-lin:
Pushed `2c75c57a`. All three CI failures fixed, `main` merged, and one self-inflicted regression caught and reverted before it could reach you.

## The regression — worth knowing about

My commit `d45676bd` accidentally included a **one-line source mutation**:

```python
-    if wire_type == 1:        # correct
+    if not (wire_type == 1):  # inverted
```

in `_skip_value`'s fixed64 arm. Cause: I ran an adversarial reviewer that verifies findings by temporarily breaking source lines, and I pointed it at the same worktree I was staging from — so `git add -A` captured a live mutant mid-experiment.

Caught two independent ways before push: by diffing `src/` against `main` (this branch is tests-only by construction, so *any* `src/` delta is a defect), and by the reviewer itself noticing its mutant had been committed. Reverted in `2c75c57a`; `src/` is now byte-identical to `main`.

## CI fixes — none were defects in the code under test

**Python 3.10/3.11 worker crashes.** Before 3.12, `asyncio.wait_for` wraps its awaitable in a Task, so a `BaseException` is re-raised into the *event loop* rather than the awaiting frame — killing the xdist worker instead of reaching `pytest.raises`. 3.12 removed that wrapping. The production guard is version-independent; only this way of observing it is not, so those two cases are pinned to 3.12+ with the reason inline. The `CancelledError` case is unaffected and still runs everywhere.

**Python 3.10 cancellation message.** `CancelledError.args` loses its message through shielded republication on 3.10. Now follows the repo's existing `_assert_republished_cancel_message` convention, and additionally asserts the message was not *replaced* by the later cancel — the actual contract, true on every version.

**CodeQL high severity (`py/incomplete-url-substring-sanitization`).** `"https://u2.example.com" in str(error)` reads to the scanner as a URL sanitization check. Replaced with an exact assertion on the previewed URL list parsed from the message — strictly stronger, since it pins which three appear and in what order.

## Verification on the merged result

```
pytest -n auto --dist loadgroup -m "not repo_lint and not requires_playwright and not requires_chromium"
                                                          # 18116 passed
pytest tests/unit -m "(requires_playwright or requires_chromium) and not reality" -n 0
                                                          # 84 passed
pytest tests/_guardrails                                  # 2254 passed
pytest -m repo_lint                                       # 1981 passed
mypy src/notebooklm scripts/_live_auth_scenarios          # clean
ruff check / format                                       # clean
git diff origin/main...HEAD --name-only | grep ^src/      # empty — tests only
```

Branch total now **96.12%**, uncovered 2207.

## One known flake, pre-existing and unrelated

`test_auth_mint_service.py::test_cancelled_mint_returns_before_worker_settles_and_worker_retains_only_arguments` fails intermittently in local runs. It is untouched by this PR, passes 10/10 in isolation and 57/57 as a file, and its assertion is `assert release.wait(2.0)` — a 2-second wall-clock bound on a thread handoff. Every failure occurred while multiple agents were saturating the machine. Worth tightening separately.

teng-lin:
Pushed `56b606dd` — all three confirmed weak assertions from the adversarial review are fixed, each re-proven with the reviewer's own mutation.

The review ran 136 mutation experiments (8 hand-targeted, 128 automated across the 606 source lines this diff uniquely covers) for a 93% kill rate on non-`skill_cmd` lines, with every survivor explained. Three findings were attributable to this change, and all three were tests whose **name advertised a contract the assertions did not hold**:

**1. A credential-leak guard that was a no-op on Windows** (highest severity). `_assert_library_traceback_is_secret_free` selected library frames with `"/src/notebooklm/" in co_filename`. `co_filename` uses backslashes on Windows, so every frame was skipped and the secret scan checked *nothing* — silently disarming the headline assertion of three tests that exist to prove a capability URL and bearer never survive in a traceback.

Worth recording: my first fix used `PurePath(...).parts`, which **also fails** — `PurePath` only treats `\` as a separator when the test itself runs on Windows. I caught that by simulating a Windows `co_filename` rather than assuming, and switched to explicit separator normalisation. The helper now also asserts it inspected at least one frame, so it can never again pass by scanning zero, and reports frame *names* instead of interpolating `repr(f_locals)` — which in this module would print the capability URL on failure.

**2. Two queue-wait assertions that pinned nothing.** `waits[0] >= 0.0` holds for an absolute clock reading as readily as an elapsed duration, so a recorder emitting `monotonic()` instead of `monotonic() - start` passed in *both* the web and Android pipelines. Now bounded `0.0 <= wait < 1.0`.

**3. "before a single byte is read" did not observe reads.** The declared-oversize refusal asserted only the `ValidationError` and the stream exit; draining the entire body before the header check still passed. `_StubResponse` now counts chunks pulled and the test asserts zero.

Re-proven after fixing: leaking the capability URL out of `download_url` fails 3 tests; swapping either elapsed-wait computation for an absolute timestamp fails both queue-wait tests; draining the body before the declared-size check fails the oversize test. `git diff origin/main...HEAD -- src/` is empty.

**Not fixed, deliberately:** finding #4 (a test that patches `time.monotonic` process-wide, so it also swaps the event loop's clock) is brittleness, not a false pass — it fails loudly if the code under test gains a suspension point. Same for the systemic note about six places patching stdlib module objects: the reviewer found no actual cross-test poisoning, since monkeypatch restores and pytest runs serially within a worker. Both are worth a follow-up, neither belongs in a coverage PR.

**Guardrail review, separately: all green on this head.** `tests/_guardrails` 2254 passed, `-m repo_lint` 1981 passed, no monkeypatch-guardrail violations, no module shadowing, no mis-tiered tests. The only deletion in the entire 7,700-line change is three lines in `_make_pipeline` moved verbatim into an `if auth is None:` default branch — every existing caller gets a byte-identical fixture.

teng-lin:
Pushed `3941930b` — all six actionable findings fixed, each proven by mutation.

**`Task.cancelling()` on Python 3.10 (codex P1)** — correct and it would have broken the required 3.10 runner. `cancelling()` is 3.11+, and `cancelled()` is still `False` immediately after `cancel()`, so the `or` short-circuit would have hit the missing attribute whenever the cancel was requested but not yet delivered.

Replaced with `await asyncio.wait_for(bystander, timeout=5)` inside `pytest.raises(CancelledError)`. That is version-agnostic *and* strictly stronger — the old form cancelled the bystander itself on the very next line, so it passed even if `prepare_close` had cancelled nothing. Proven: making `prepare_close` skip its cancel loop now fails the test.

**Cancellation message on 3.10 (codex P1)** — already fixed before it was filed. That assertion was made version-aware earlier on this branch (`(("first",) if sys.version_info >= (3, 11) else ())`, plus an added check that the message was not *replaced* by the later cancel). The review ran against the pre-merge head.

**Event-loop handler leaks (codex P2 + coderabbit)** — both correct. One test cleared the handler to `None` unconditionally, the other never restored it; neither survived an assertion failure. Both now save the previous handler and restore it in a `finally`.

**Duplicate-warning count (coderabbit)** — correct, `emitted[-1]` would miss an earlier duplicate. Now compares every warning line against the single expected one.

**Unparseable frontmatter (coderabbit)** — correct and the sharpest of the batch: with no over-limit description in either input, "the check was skipped" and "the check ran and passed" were indistinguishable. Both inputs now carry a 1100-character description against the 1024 limit, which the existing empty-stderr assertion discriminates.

**`decode_source` spy (coderabbit)** — correct. The echo was fully decodable, so `None` alone could not tell "never decoded" from "decoded and discarded". Added a spy on the module-bound `decode_source`; proven by forcing that branch to always decode. (Public name, so outside the ADR-0007 private-attr guardrail — verified.)

**Not changed, with reason:** the systemic note about tests patching stdlib module objects process-wide. The adversarial review specifically checked for cross-test poisoning and found none — monkeypatch restores every one, and pytest runs serially within an xdist worker. It is a real follow-up, not a fix that belongs in a coverage PR.

Verification: `ruff` clean, 16568 unit tests pass, `git diff origin/main...HEAD -- src/` empty.

PR #2391, opened 2026-09-06
fix(cli): resolve ask --source before --new deletes history

### LINKED ISSUE #2381: ask --new deletes the conversation before rejecting an invalid --source
## Bug

`notebooklm ask --new` deletes the current conversation **before** it validates `--source` (and before `chat.ask`).

In `src/notebooklm/cli/chat_cmd.py` the order inside the client context is:

1. resolve notebook
2. if `--new` and a conversation exists: confirm, then `delete_conversation`
3. **then** `resolve_source_ids(...)`
4. then `chat.ask`

An invalid / ambiguous source reference therefore fails **after** the conversation is already gone. Confirmation only asked about deletion, not about the rest of the command being well-formed.

`--new` is CLI-only (MCP has no equivalent delete-then-ask path).

## Required change

Finish every preflight that can still abort the command **before** deletion:

- source id / name resolution
- other cheap validation already available without the new conversation

Then delete, then ask. If source resolution fails, the previous conversation must still exist.

Do not hold a client operation open across an interactive `click.confirm`. Confirmation can stay before the mutation; the mutation must not run until preflight passed.

## Tests

- `--new --source not-a-real-source` (or an ambiguous prefix) → nonzero exit, **conversation id unchanged**
- `--new` with valid sources still deletes once, then asks
- `--json --yes --new` with a bad source: same preservation (no stdin confirm)

teng-lin:
`ask --new` now resolves source references before confirming and deleting the previous conversation. It verifies full UUIDs against the notebook inventory as well, so stale, foreign, and ambiguous source IDs leave history intact. Confirmation runs outside a client operation.

Fixes #2381.

Validation: 109 CLI chat/resolver tests passed; Ruff check/format and mypy passed. The committed coupling baseline includes the merged history and security fixes.

### DIFF
--- a/src/notebooklm/cli/chat_cmd.py
+++ b/src/notebooklm/cli/chat_cmd.py
@@ -165,6 +165,33 @@ _format_single_qa = format_single_qa
 _format_history = format_history
 
 
+def _confirm_new_conversation_deletion(
+    conversation_id: str, *, assume_yes: bool, json_output: bool
+) -> None:
+    """Prompt for ``ask --new`` history deletion.
+
+    Sync and client-free so confirmation never sits inside a client
+    ``operation()`` scope. ``--json`` implies ``--yes`` so scripted callers
+    don't hang on stdin (which would also clobber JSON stdout purity). See
+    ``cli/artifact_cmd.py::artifact_delete`` for the same pattern.
+    """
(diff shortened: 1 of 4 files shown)


teng-lin:
@codex review

